PR #4, opened 2019-05-13
Update Travis config

cgriggs01:
This will update Travis to use newer go version and run the acceptance test during the PR check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 - docker
 language: go
 go:
-  - "1.9.x"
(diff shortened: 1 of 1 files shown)


rawmind0:
The update to .travis.yml should be included in PR #3 to avoid `make vet` error when check go libraries. 
`make testacc` are almost ok but failing when check `setting` resource due to an issue on rancher v2.2. It will be resolved on rancher v2.3. Shouldn't be included in pipeline until there.
Taking a look why `make website-test` is failing

PR #35, opened 2019-06-18
Fix: index out of range on rancher2_cluster resource rke_config_cloud_provider_vsphere

rawmind0:
Fix issue #34


rancher2/structure_cluster_rke_config_cloud_provider_vsphere.go
```diff
@@ -75,10 +75,11 @@ func flattenClusterRKEConfigCloudProviderVsphereVirtualCenter(in map[string]mana
 	}
 
 	out := make([]interface{}, len(in))
+	lenP := len(p)
```
cloudnautique, line 78: Why add a line of code, this is only used in a single place. I would just do it in place below.

rawmind0: It's just to avoid call to `len(p)` in every loop iteration.

cloudnautique (approved):

PR #44, opened 2019-06-28
Make AWS session token in EKS config sensitive

keymon:
The session token is a sensitive value and should not
be printed in the output, as the secret and key.


dstockton (approved):

dstockton:
@rawmind0 it'd be great if you have time to approve & merge this! Thanks

rawmind0 (approved):
LGTM

Issue #11, opened 2019-05-20
Add data sources for several rancher2_* resources


jnicholls:
- [x] rancher2_catalog
- [x] rancher2_cloud_credential
- [x] rancher2_cluster
- [x] rancher2_cluster_driver
- [x] rancher2_cluster_logging
- [x] rancher2_cluster_role_template_binding
- [x] rancher2_etcd_backup
- [x] rancher2_global_role_binding
- [x] rancher2_namespace
- [x] rancher2_node_driver
- [x] rancher2_node_pool
- [x] rancher2_project
- [x] rancher2_project_logging
- [x] rancher2_project_role_template_binding
- [x] rancher2_registry
- [x] rancher2_user

It seems that the suggested approach to querying existing resources is to `terraform import` them into terraform state by their id. I think there are use cases such as node pools, cloud credentials and projects where we'd want to query them by name or other attributes if applicable.

rawmind0:
Data sources merged to provider. Please, reopen issue if needed.

jnicholls:
This is amazing work, thank you @rawmind0!!

PR #49, opened 2019-07-03
Catalog data source

rawmind0:
Fix PR #11


cloudnautique:
Is this only the global catalogs? How are project catalogs handled? we should include an example for project catalogs.

rawmind0:
Just global catalog are implemented. Project and cluster catalogs are not managed yet, on the way.

cloudnautique (approved):

PR #52, opened 2019-07-08
Cluster and project logging data sources

rawmind0:
Fix issue #11


cloudnautique:
Do we need to mark the `*_configs` as sensitive on the data sources since the config might have username/passwords?

rawmind0:
I think it's not needed, due to username/passwords are already marked as sensitive on `logging*ConfigFields()` definitions.

website/docs/d/projectLogging.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+  project_id = "<project_id>"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Required) The project id to configure logging (string)
+* `name` - (Required) The name of the cluster logging config (string)
```
cloudnautique: There is on only one per project, and cluster. Why do they need the name? Type maybe...

rawmind0: You are right. Addressing it to let name as computed field.

rawmind0: Fixed `name` argument as computed.

website/docs/d/projectLogging.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+# rancher2\_project\_logging
+
+Use this data source to retrieve information about a Rancher v2 Project Logging.
+
+## Example Usage
+
+```hcl
+data "rancher2_project_logging" "foo" {
+  name = "foo"
```
cloudnautique: Name is referenced here.

website/docs/d/clusterLogging.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+# rancher2\_cluster\_logging
+
+Use this data source to retrieve information about a Rancher v2 Cluster Logging.
+
+## Example Usage
+
+```hcl
+data "rancher2_cluster_logging" "foo" {
+  name = "foo"
```
cloudnautique: is this no longer needed now?

website/docs/d/projectLogging.html.markdown
rawmind0: Addressed.

website/docs/d/clusterLogging.html.markdown
rawmind0: Addressed.

cloudnautique (approved):

Issue #38, opened 2019-06-21
You should be able to bootstrap and use create resources in the same run.


jgreat:
This pattern should work. Its pretty useless as a tool if I can't use it to immediately create resources on the rancher server I just created.

```
provider "rancher2" {
  api_url   = "https://${local.name}.${local.domain}"
  bootstrap = true
}
resource "rancher2_bootstrap" "admin" {
  depends_on = [
    "null_resource.wait_for_rancher"
  ]
  password = "${local.rancher_password}"
}
resource "rancher2_cluster" "aws_rke" {
...
}
```

Currently this returns a "401" error.
`rancher2_cluster.aws_rke: Bad response statusCode [401]. Status [401 Unauthorized]. Body: [message=must authenticate] from [https://cure53-testing-1.eng.rancher.space/v3]`

rawmind0:
@jgreat, bootstrap resource is a special resource. You couldn't mix any other provider resource with bootstrap. 

Anyway, to make your pattern work in one .tf file, take a look to https://github.com/terraform-providers/terraform-provider-rancher2/issues/13#issuecomment-496581493

jgreat:
The two provider option is ok with the limitation of TF, but the piece that's missing here is allowing for the url and token to be set by generated values.  The Helm and Kubernetes providers allow this if you populate the kubernetes block with host, username, certs, etc... This way you can use the bootstrap resource and pass the resulting values into another rancher2 provider.

Example of values generated by rke_cluster resource and then used with the helm provider.
```
provider "helm" {
  install_tiller  = true
  namespace       = "kube-system"
  service_account = "tiller"

  kubernetes {
    host                   = "${rke_cluster.rancher_server.api_server_url}"
    username               = "${rke_cluster.rancher_server.kube_admin_user}"
    client_certificate     = "${rke_cluster.rancher_server.client_cert}"
    client_key             = "${rke_cluster.rancher_server.client_key}"
    cluster_ca_certificate = "${rke_cluster.rancher_server.ca_crt}"
  }
}
```

cloudnautique:
@jgreat I don't follow, in the example block posted in @rawmind0 s comment, the second provider (non bootstrap) is being populated with the URL and API token generated from the bootstrap resource. Is there missing data that we should be exporting?

drpebcak:
When you try to do this all in one run like this:

```hcl
provider "rancher2" {
  alias     = "bootstrap"
[...]

PR #64, opened 2019-07-16
Move credential checking into client loading

SinisterMinister:
Addresses #38 by moving the credentials check to right before instantiating the client so that the error is raised during run-time instead of apply time. Here's what the error looks like in practice if the credentials aren't provided:

image


rawmind0:
Superseded by #67

Issue #72, opened 2019-07-21
Project/Namespace dependencies on the cluster being active


josmo:
This may be related to #12 

I'm in a situation where I 
1. create rancher2_cluster resource called "clusterA"
1. I then create x number of workers/etc/controllers to be part of "clusterA" with the user_data being injected from clusterA.cluster_registration_token[0].node_command
1. I have x number of "rancher2_project" with a cluster_id of "clusterA.id"

Expected:  wait till clusterA is active with the workers/etcd/controllers to create the projects

Actual: fails with an Error creating project: Cluster ID xxxxx is not active

rawmind0:
This is not related with issue #12. `rancher2_namespace` and `rancher2_project` resources need that the cluster is `active` to be created. 

Submitted PR #73, that adds `wait_for_cluster` argument on `rancher2_namespace` and `rancher2_project` resources to wait for cluster is `active`.

jnicholls:
See [my comment](https://github.com/terraform-providers/terraform-provider-rancher2/pull/73#issuecomment-514804591) on PR #73. Beyond the `rancher2_namespace` and `rancher2_project` resources, there are other kubernetes-based resources which rely on the kube-apiserver being ready. It would be great to create a new type of resource named something like `rancher2_cluster_wait_until_active` that takes the cluster_id and an optional timeout, and will wait until the cluster's kube-apiserver becomes ready before proceeding. Then, kubernetes/helm/etc. resources can explicitly be set to depend upon that resource and will thus wait for it to return.

Thoughts?

josmo:
Thanks @rawmind0!!

josmo:
@jnicholls I like the proposal, I don't think it's a huge deal for namespaces and the likes in k8s resources since you could still use the rancher resources for that, however things like setting up volume classes that doesn't have deps on namespaces, starting in the system/default namesspaces, it would be really useful

narfeta:
hi all, any update about this? it would be really useful to have this feature.

drpebcak:
I think it makes sense to have optional waits on all the rancher2 resources, but then also have something separate that can tell you if the cluster itself is ready, like `rancher2_cluster_wait_until_active`

rawmind0:
I think that @jnicholls proposal is indeed moreover to "external" rancher2 resources. For rancher2 resources, not sure it it's needed on all, but PR #73 extra argument should be enough, due to they are able to check the cluster state.

rawmind0:
[...]

PR #73, opened 2019-07-22
Added wait_for_cluster argument on rancher2_namespace and rancher2_project resources

rawmind0:
Fix issue #72


jnicholls:
So...I would really like if you instead created an entirely new resource named `rancher2_cluster_wait_until_active` or something like that. That way, it can be used as a dependency for `kubernetes` and `helm` provisioner resources so that they will wait until the cluster's api server is responding.

As a workaround, this PR's work could be used to create a dummy (or real) namespace that waits on the cluster becoming active, and in turn the rancher2_namespace resource is used as a dependency for `kubernetes` or `helm` resources.

cloudnautique (approved):

PR #80, opened 2019-07-24
Fix the delete of a cluster when we don't have full read permission

### LINKED ISSUE #50: terraform destroy eks fails with Status [403 Forbidden]
when we delete an EKS cluster, the cluster actually gets deleted in the Rancher UI, but terraform does not know that the cluster got deleted in Rancher and it still tries to destroy every 10s. Rancher responds with a "403 Forbidden" error

```
module.k8s-rancher-sz-1.rancher2_cluster.eks_cluster: Destroying... (ID: c-qphhb)
module.k8s-rancher-sz-1.rancher2_cluster.eks_cluster: Still destroying... (ID: c-qphhb, 10s elapsed)
module.k8s-rancher-sz-1.rancher2_cluster.eks_cluster: Still destroying... (ID: c-qphhb, 20s elapsed)
module.k8s-rancher-sz-1.rancher2_cluster.eks_cluster: Still destroying... (ID: c-qphhb, 30s elapsed)
Releasing state lock. This may take a few moments...

Error: Error applying plan:

1 error(s) occurred:

* module.k8s-rancher-sz-1.rancher2_cluster.eks_cluster (destroy): 1 error(s) occurred:

* rancher2_cluster.eks_cluster: [ERROR] waiting for cluster (c-qphhb) to be removed: Bad response statusCode [403]. Status [403 Forbidden]. Body: [baseType=error, code=Forbidden, message=clusters.management.cattle.io "c-qphhb" is forbidden: User "u-67xj8" cannot get resource "clusters" in API group "management.cattle.io" at the cluster scope] from [https://rancher.infra.xxxxx.xxxxxxxx/v3/clusters/c-qphhb]
```

thomas-maurice:
This pull request fixes the following behaviour happening when a user does not have the permission to list all clusters:
* tries to delete a cluster
* delete succeeds
* polls the API to get the state of the cluster
* gets a 403 because the cluster does not exist anymore but we can't list all of them (so no 404)
* fails the terraform run

Test plan:
* integration test on a running rancher instance
* unit tests
```
$ make test
==> Checking that code complies with gofmt requirements...
go test -i "./rancher2" || exit 1
echo "./rancher2" | \
                xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 ./rancher2
ok      github.com/terraform-providers/terraform-provider-rancher2/rancher2     0.058s
```

This fixes #50

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -3,10 +3,12 @@ package rancher2
 import (
 	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/rancher/norman/clientbase"
 	norman "github.com/rancher/norman/types"
 	managementClient "github.com/rancher/types/client/management/v3"
 )
@@ -254,6 +256,9 @@ func clusterStateRefreshFunc(client *managementClient.Client, clusterID string) 
 		err := client.APIBaseClient.ByID(managementClient.ClusterType, clusterID, obj)
 		if err != nil {
 			if IsNotFound(err) {
+				return obj, "removed", nil
+			} else if e, ok := err.(*clientbase.APIError); ok && e.StatusCode == http.StatusForbidden {
+				log.Printf("[INFO] Got access denied to the cluster, which means it has been removed")
 				return obj, "removed", nil
 			}
 			return nil, "", err
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -3,12 +3,10 @@ package rancher2
 import (
 	"fmt"
 	"log"
-	"net/http"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/rancher/norman/clientbase"
 	norman "github.com/rancher/norman/types"
 	managementClient "github.com/rancher/types/client/management/v3"
 )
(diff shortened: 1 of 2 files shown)


rancher2/resource_rancher2_cluster.go
```diff
@@ -255,6 +257,9 @@ func clusterStateRefreshFunc(client *managementClient.Client, clusterID string)
 		if err != nil {
 			if IsNotFound(err) {
 				return obj, "removed", nil
+			} else if e, ok := err.(*clientbase.APIError); ok && e.StatusCode == http.StatusForbidden {
+				log.Printf("[INFO] Got access denied to the cluster, which means it has been removed")
+				return obj, "removed", nil
```
rawmind0, line 263: Using `IsForbidden` function defined in `util.go`

```
-			if IsNotFound(err) {
+			if IsNotFound(err) || IsForbidden(err) {
				return obj, "removed", nil
-			} else if e, ok := err.(*clientbase.APIError); ok && e.StatusCode == http.StatusForbidden {
-				log.Printf("[INFO] Got access denied to the cluster, which means it has been removed")
-				return obj, "removed", nil
```

rancher2/resource_rancher2_cluster.go
```diff
@@ -3,10 +3,12 @@ package rancher2
 import (
 	"fmt"
 	"log"
+	"net/http"
```
rawmind0: Using `IsForbidden` function defined in `util.go`

```
-	"net/http"
```

rancher2/resource_rancher2_cluster.go
```diff
@@ -3,10 +3,12 @@ package rancher2
...
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/rancher/norman/clientbase"
```
rawmind0: Using `IsForbidden` function defined in `util.go`

```
-	"github.com/rancher/norman/clientbase"
```

rawmind0 (changes requested):
Hi @thomas-maurice , thanks for the PR. It fixes issue #50 isn't it??

What do you think if we add `IsForbidden` function on `util.go` file, instead inline code on `resource_rancher2_cluster.go`??

```
func IsForbidden(err error) bool {
	apiError, ok := err.(*clientbase.APIError)
	if !ok {
		return false
	}

	return apiError.StatusCode == http.StatusForbidden
}
```

thomas-maurice:
@rawmind0 oh I haven't realised there was already an issue opened on this one, yep indeed it does :D I agree using an IsForbidden function is going to look way cleaner, I'll do the necessary fixes !

thomas-maurice:
Actually I'll add a comment to explicit why the notfound or forbidden thing because that can be confusing at first

rawmind0 (approved):

rawmind0 (approved):
Thank @thomas-maurice 

LGTM

PR #84, opened 2019-07-30
WIP: Allow pull EKS config AWS creds from env vars and update creds before delete

keymon:
What?
----

This commit allows pull the AWS credentials for the EKS cluster from the
standard AWS cli environment variables: `AWS_ACCESS_KEY_ID`,
`AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`

By pulling the credentials from the environment we can ensure that we
always get the latest credentials, including during the delete operation,
and use a external tool to assume the role or get the STS token.

This, combined with update the rancher stored AWS creds before deletion,
solves the issue described [1] were we cannot delete a cluster without
apply first new credentials.

[1] https://github.com/terraform-providers/terraform-provider-rancher2/issues/74

Tests
-----

 - Pulling creds from environment has been unit tested
 - Update the cluster before delete, tested manually as there are no
   integration tests for EKS (AFAIK)


keymon:
I tested this in different scenarios: expired tokens or incorrect tokens, deleted ones, etc, and works as expected.

keymon:
During the destroy operation all the `schema.ResourceData` is empty except the old state attributes. I can only get the driver class and if `aws_creds_from_env` is set from the state.

I need to refactor a little bit more. 

I might also unit test this using some mocking.

keymon:
Closing due inactivity, we will come with a different implementation.

Issue #69, opened 2019-07-17
Add ability to provision applications within a project


SinisterMinister:
Provisioning apps from Rancher or Helm charts would be super useful. In my case, I'm setting up our CI/CD environment and need to deploy the Jenkins helm chart within a project.

I can try to tackle this if it's a feature that's desired. If there's a workaround instead, I'll take that as well.

rawmind0:
Submitted PR #125 to provide rancher app support.

rawmind0:
Merged PR to address this issue. Please, reopen it if needed.

PR #86, opened 2019-08-05
Added rancher2_app resource and datasource. Go, test and doc files

gabrielrcouto:
Fix #69

PR #125, opened 2019-09-04
Added rancher2_app datasource and resource

rawmind0:
Fix issue #69 

Supersedes PR #86


website/docs/r/app.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+    "ingress.annotations.nginx.ingress.kubernetes.io/force-ssl-redirect" = true
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `external_id` - (Required) The url of the app template on a catalog. If modified, app will be upgraded (string)
```
drpebcak, line 78: I think we need another solution for this. Expecting the user to craft this weird url is terrible UX. It would be better to have 3 separate args for `catalog_name`, `template_name` (which should be called something more obvious like `app_name` or `chart_name`), and `template_version` (which again should probably be called something more obvious like `chart_version` or `app_version`).

drpebcak (changes requested):

website/docs/r/app.html.markdown
cloudnautique: I agree, can we construct the id via "catalog_name", "template_name", "version" ?

SinisterMinister: If we're passing version, should we allow for it to default to latest like in the UI?

drpebcak: Yeah I think it makes sense to have `version` be optional.

rawmind0: Agreed to use `catalog_name`, `template_name` and `template_version` to construct `external_id` and facilitate UX.

For me, makes no sense to let `version` be optional. This is a generic label, not reproducible. Also, how to manage app upgrade?? what would happens when the chart is upgraded on the helm repo?? The app will be set to be updated on next terraform apply.

drpebcak: Yeah, that would be the expected behavior I think. That's how the helm provider does it as well: https://www.terraform.io/docs/providers/helm/release.html

If `version` is not specified, it installs the latest every time terraform runs (which does nothing if you are already on the latest version).

rawmind0: The problem is not when tf does nothing, the problem is when the helm char is updated and terraform is run, the app will be marked to be upgraded, without any change on tf file.

Also, on rancher API you must specify the version on the `external_id`

rawmind0: What do you think if we use a format like multi cluster app, `template_version_id = <catalog_name>-<template_name>-<template_version>`??

drpebcak: Yeah, I wouldn't want that to happen to me, but that is how the helm provider works. I believe the rancher cli also works that way. If you do not specify the version, it will install the latest.

Sort of like using the docker `latest` tag I don't think it should be recommended in production, but it is something that people do.

cloudnautique: I agree it is bad form to not specify a version, but there is enough prior art out there that suggests no version means installing the latest all the time. However, in the case of demo environments and other such non-sense that won't last long, it has some value installing the latest.

I think in this case, TF provides a reasonable method to see that it is going to do something with the plan/apply steps being discrete. At that point, adding a version == to the installed version should mitigate any future changes until desired. 

As far as x-y-z as a string.. I prefer the fields be discrete. If there is enough prior art to say this is the TF way, so be it, but no.

drpebcak: Separate fields is definitely more idiomatic.

rawmind0: ok guys, separated fields and no version allowed.

cloudnautique (approved):

drpebcak (approved):
lgtm. I'll build a binary to try this one out too.

rawmind0:
Acceptance tests passing fine locally. Almost passing at travis, but eventually failing with weird messaged, by a possible travis bug, https://travis-ci.community/t/travis-builds-fail-due-to-slowness/4571/11

Disabling acceptance tests until travis issue is solved

PR #128, opened 2019-09-05
Added rancher2_multi_cluster_app datasource and resource

rawmind0:


cloudnautique:
Can this one match up with the UX for cluster apps?

rawmind0:
Yep, i'll update it asap.

rawmind0:
Updated to match up with the UX for cluster apps.

website/docs/r/multiClusterApp.html.markdown
```diff
@@ -0,0 +1,143 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required/ForceNew) The multi cluster app name (string)
+* `roles` - (Required) The multi cluster app roles (list)
+* `targets` - (Required) The multi cluster app target projects (list)
+* `template_version_id` - (Required) The multi cluster app template version ID. Should be in the format `<catalog_name>-<template_name>-<template_version>` (string)
```
cloudnautique: This should be updated?

rawmind0: Sure, i missed one commit..... Updated

cloudnautique (approved):

PR #129, opened 2019-09-06
Add missing url path for rancher_url

ajfriesen:
When creating the azure ad authentication via the web interface in rancher the URL vill have this path:

`*/verify-auth-azure`

In terraform you need to specify that path otherwise you get into a login loop when trying to login with an azure ad account.


rawmind0 (approved):
LGTM

PR #135, opened 2019-09-18
Fix: Allow AKS monitoring to be disabled

dstockton:
`EnableMonitoring` attribute was previously not included in the marshaled JSON. Following the pattern for EKS/GKE configurations, I've switched this to a `*bool` type and updated the related code & tests.

This means `enabled_monitoring = false` is now functional for AKS clusters.


dstockton:
@rawmind0 would you be able to approve & merge this extra-small Azure AKS bug fix, please?

rawmind0 (approved):
LGTM

Issue #152, opened 2019-10-08
[Feature request] Support always_pull_images


rbq:
```tf
resource "rancher2_cluster" […] {
  rke_config {
    services {
      kube_api {
        always_pull_images = true
      }
    }
  }
}
```

Results in the following message:

> An argument named "always_pull_images" is not expected here.

rawmind0:
This argument is not suported, https://www.terraform.io/docs/providers/rancher2/r/cluster.html#kube_api-1

rbq:
@rawmind0 Thanks. Is there a workaround for RKE fields still missing from the provider that doesn't require switching from TF to RKE/YAML completely?

Would be great if [this setting](https://github.com/terraform-providers/terraform-provider-rancher2/blob/92c7c011d142ba5c8acfff52c1e7589bf1ae093d/vendor/github.com/rancher/types/client/management/v3/zz_generated_kube_apiservice.go#L16) could be used just like the others though.

rawmind0:
Submitted PR to provide support to `always_pull_images` on kube API config.

rbq:
Thanks a lot!

PR #155, opened 2019-10-10
Added always_pull_images argument to kube_api for rke clusters

rawmind0:
Fix issue #152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,13 @@
 ## 1.6.1 (Unreleased)
+
+FEATURES:
+
+ENHANCEMENTS:
+
+* Added `always_pull_images` argument on `kube_api` argument on `rke_config` argument for `rancher2_clusters` resource
+
+BUG FIXES:
+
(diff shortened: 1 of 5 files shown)


cloudnautique (approved):

rbq:
Sorry I'm so late with this, but I only now checked the change to the description in the docs (“Always pull kube API images”). That's not really what the option is for.

It actually enables the [AlwaysPullImages](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages) admission controller (see [Rancher docs](https://rancher.com/docs/rke/latest/en/config-options/services/#kubernetes-api-server-options)). This enforces the “Always” pull policy for all workloads, which makes it impossible for users to run containers based on private images of other users—circumventing the purpose of a private registry—by abusing the Docker cache on a shared node.

This was introduced with rke 0.2 and enabling it is a recommendation in the Rancher hardening guide.

rawmind0:
Thanks @rbq . Please take a look at PR #157

PR #158, opened 2019-10-11
added missing list item declaration

ebauman:
`security_groups` was missing a list item declaration (`*`) under `eks_config`, which was causing misformatting


rawmind0 (approved):
LGTM

Issue #169, opened 2019-10-21
New resource: "rancher2_token"


headcr4sh:
I am in need to automatically create Rancher 2 tokens to hook my Rancher 2 installation into some automation mechanisms.

The API that I am talking about can be found under `/v3/tokens` on any Rancher 2.x installation.

#### Example of possible API usage

```hcl
resource "random_string" "bot_password" {
  length  = 16
   special = true
}

resource "rancher2_user" "bot" {
   username    = "r-bot"
   name        = "Your friendly Rancher2 bot"
   description = "Cattle-Bot"
   password    = random_string.bot_password.result
}
 
resource "rancher2_token" "bot" {
   user_id     = rancher2_user.bot.id
   description = "Token to be used by the bot application to control the bot user"
 }
```

#### Possbile implementation

Our own in-house developed (and now deprecated) Rancher 2 provider plugin offered this functionality and it is the final blocker that keeps us from fully adopting the official Rancher 2 provider plugin.

I assume it might be rather easy to copy my code over (with just some minor tweaks) to be used here:
https://github.com/iplabs/terraform-provider-rancher2/blob/master/rancher2/resource_token.go

rawmind0:
Hey @headcr4sh, is you code working?? 

May be i'm missing something, but afaik you can't manage tokens for other users on rancher2, just your own tokens. When you create a token by the rancher2 API, no matter what you set at `user_id`, the token is created with the user id of the user that do the request.

To manage other users tokens, tf provider would need to relogin as other users to manage them. Username and password or user token would be needed to define other user token. Then, tf provider would relogin as proper user, create token and relogin back as provider user. Same for read, update and/or delete operation. Not sure that it's a good way.

BTM, i'm finishing a PR implementing Rancher2 tokens but just to manage "current" user tokens. It could be useful with terraform modules.

headcr4sh:
> May be i'm missing something, but afaik you can't manage tokens for other users on rancher2, just your own tokens. When you create a token by the rancher2 API, no matter what you set at user_id, the token is created with the user id of the user that do the request.

Mh... I am not quite sure if the code works as supposed to. I was indeed only creating tokens for the currently logged in user...

rawmind0:
Merged PR to provide rancher2_token resource

PR #178, opened 2019-10-31
New rancher2_token resource

rawmind0:
Fix issue #169

### DIFF
--- a/website/docs/r/token.html.markdown
+++ b/website/docs/r/token.html.markdown
@@ -1,0 +1,66 @@
+---
+layout: "rancher2"
+page_title: "Rancher2: rancher2_token"
+sidebar_current: "docs-rancher2-resource-token"
+description: |-
+  Provides a Rancher v2 Token resource. This can be used to create tokens for Rancher v2 provider user and retrieve their information.
+---
+
+# rancher2\_token
+
+Provides a Rancher v2 Token resource. This can be used to create Tokens for Rancher v2 provider user and retrieve their information.
+
+There are 2 kind of tokens:
+- no scoped: valid for global system.
+- scoped: valid for just a specific cluster (`cluster_id` should be provided).
+
+Tokens can't be updated once created. Any diff in token data will recreate the token. If any token expire, Rancher2 provider will generate a diff to regenerate it.
+
+## Example Usage
+
+```hcl
+# Create a new rancher2 Token
+resource "rancher2_token" "foo" {
+  description = "foo token"
+  ttl = 1200
+}
+# Create a new rancher2 Token scoped
+resource "rancher2_token" "foo" {
+  cluster_id = "<cluster-id>"
+  description = "foo token"
+  ttl = 1200
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
+* `description` - (Optional/ForceNew) Token description (string)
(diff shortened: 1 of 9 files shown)


website/docs/r/token.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
+* `description` - (Optional/ForceNew) Token description (string)
+* `enabled` - (Optional/ForceNew) Token enabled. Default `true` (bool)
```
cloudnautique: what does this do?

website/docs/r/token.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+* `expired` - (Optional/ForceNew) Token enabled. Default `false` (bool)
```
cloudnautique: what does this do? Should it be an attribute?

website/docs/r/token.html.markdown
rawmind0: `enabled` attribute is used control if the token is enabled. When you run `terraform apply` and the token is not enabled, the diff is detected and terraform will renew the token.

website/docs/r/token.html.markdown
rawmind0: `expired` attribute is used control if the token is expired. When you run `terraform apply` and the token is expired, the diff is detected and terraform will renew the token.

website/docs/r/token.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
+* `description` - (Optional/ForceNew) Token description (string)
+* `enabled` - (Optional/ForceNew) Token is enabled. This argument doesn't need to be provided, it just control and would generate a terraform diff to renew the token in case it's disabled. If i Default `true` (bool)
+* `expired` - (Optional/ForceNew) Token is expired. This argument doesn't need to be provided, it just control and would generate a terraform diff to renew the token in case it's expired. Default `false` (bool)
```
cloudnautique: With this understanding, should we change the option to `renew_expired`? I think that describes what it does better. Would we really not want to renew expired tokens? I feel like you would want to always renew if you are generating it programmatically.

rawmind0: For sure `renew_expired` described what it does better. The problem is that `enabled` and `expired` attributes are still needed to make that terraform notice the change once token is disabled or expired on Rancher side. 

Normally, you would always want to renew tf generated tokens and this is the default resource behaviour and the main reason to add `enabled` and `expired` attributes. If the token is expired or disabled, tf notice the diff and renew the token if you apply tf plan. But eventually for temporary token, may be you want to clean it up once it's expired and you get the tf diff.

cloudnautique: Looking at the Rancher API, 'enabled' should just be an attribute not an option. It is functionally useless in Rancher. A user can not disable the resource in the UI or API.

Expired should be an attributed as well.

As an option, we should allow `renew_expired` (bool) so that a user could create a single use token and not renew it after every run if they want to. So if set to false, then an expired token should not show a need to refresh on plan/apply.

rawmind0: PR updated, added `renew` argument (default `true`) and set `expired` and `enabled` as attributes…

cloudnautique (approved):

Issue #120, opened 2019-08-30
Allow creation of regional GKE clusters through terraform provider


Oats87:
Upstream work to enable regional GKE clusters can be found tagged to the Rancher issue here: https://github.com/rancher/rancher/issues/17099

This issue is to help track the requirement to be able to create regional GKE clusters with the terraform provider release that will coincide with the Rancher 2.3 release.

rawmind0:
PR to address this isssue is merged on master. It will be included on next release. Plesse, reopen issue if needed.

Issue #183, opened 2019-11-08
feature request: Azure cloud provider loadBalancerSku


ajchiarello:
In the rancher2_cluster resource, under the rke_config>cloud_provider>azure_cloud_provider, please allow for setting the loadBalancerSku setting (documented here: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/cloud-provider-config.md)

Thanks

ajchiarello:
It looks like the dependency has been met - should I look for this in the next release?

rawmind0:
Feature is added on PR #338 , that is already merged into master. It will be include in the next provider release. Please reopen issue if needed.

PR #185, opened 2019-11-14
Support for newly PRd `rancher2_cluster` features

dstockton:
What
----

Bringing in terraform provider support for a variety of new features

Feature - AKS - Auto-scaling support
Feature - EKS - Placement group support
Feature - EKS - Option to manage security groups outside of Rancher (e.g. when wrapped with Terraform)
Feature - EKS - Separate ASGs per subnet (for improved auto-scaling)
Feature - GKE - Multi-zone (aka regional) cluster support
Feature - GKE - Control of DefaultMaxPodsConstraint (to better utilize IP address space)
BugFix - EKS - Typo in clusterDriverEKS


dstockton:
See https://github.com/rancher/kontainer-engine/pull/200 for related kontainer-engine changes

dstockton:
I've cleaned up the PR to make it simpler to review.
The only thing I'd call-out is that I've renamed the rancher2_cluster resource aks_config attribute from `tag` to `tags`.
Previously this feature has not worked due to kontainer-engine expecting the `tags` field (not `tag`). I, therefore, felt it wasn't problematic to choose the better (plural) naming convention - which also is the same pattern as with GKE & EKS.

website/docs/r/cluster.html.markdown
```diff
@@ -765,6 +765,7 @@ The following arguments are supported:
 * `enabled` - (Optional) Enable the authorized cluster endpoint. Default `true` (bool)
 * `fqdn` - (Optional) FQDN for the authorized cluster endpoint (string)
 
+<<<<<<< HEAD
```
rawmind0: Could you please remove this line?? It was a mistake that it's already addressed on last release.

```
-<<<<<<< HEAD
```

rancher2/schema_cluster_aks_config.go
```diff
@@ -269,12 +276,50 @@ func clusterAKSConfigFields() map[string]*schema.Schema {
 			Default:     "10.0.0.0/16",
 			Description: "A CIDR notation IP range from which to assign Kubernetes Service cluster IPs. It must not overlap with any Subnet IP ranges",
 		},
-		"tag": &schema.Schema{
+		"tags": &schema.Schema{
```
rawmind0, line 286: I think that remove an argument suddenly is not a good idea. What do you think if we maintain both arguments for a while and we mark `tag` as deprecated recommending the use of `tags` instead?? It would be a safer option.

```
		"tag": &schema.Schema{
			Type:        schema.TypeMap,
			Optional:    true,
			Computed:    true,
-			Description: "Tags for Kubernetes cluster. For example, foo=bar",
+			Description: "Tags for Kubernetes cluster. For example, foo=bar. Deprecated, use `tags` argument instead",
		},
+		"tags": &schema.Schema{
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Computed:    true,
+			Description: "Tags for Kubernetes cluster. For example, foo=bar",
+		},
```

rancher2/structure_cluster_aks_config.go
```diff
@@ -135,8 +156,8 @@ func flattenClusterAKSConfig(in *AzureKubernetesServiceConfig) ([]interface{}, e
 		obj["subscription_id"] = in.SubscriptionID
 	}
 
-	if len(in.Tag) > 0 {
-		obj["tag"] = toMapInterface(in.Tag)
+	if len(in.Tags) > 0 {
+		obj["tags"] = toMapInterface(in.Tags)
 	}
```
rawmind0, line 163: Maintain `tag` and `tags` arguments for a while. As `tag` and `tags` are computed arguments, Setting both shouldn't generated a tf diff.

```
-	if len(in.Tag) > 0 {
+	if len(in.Tags) > 0 {
-		obj["tag"] = toMapInterface(in.Tag)
+		obj["tag"] = toMapInterface(in.Tags)
+		obj["tags"] = toMapInterface(in.Tags)
	}
```

rancher2/structure_cluster_aks_config.go
```diff
@@ -298,8 +357,8 @@ func expandClusterAKSConfig(p []interface{}, name string) (*AzureKubernetesServi
 		obj.SubscriptionID = v
 	}
 
-	if v, ok := in["tag"].(map[string]interface{}); ok && len(v) > 0 {
-		obj.Tag = toMapString(v)
+	if v, ok := in["tags"].(map[string]interface{}); ok && len(v) > 0 {
+		obj.Tags = toMapString(v)
 	}
```
rawmind0, line 368: Maintain `tag` and `tags` arguments for a while.

```
	if v, ok := in["tag"].(map[string]interface{}); ok && len(v) > 0 {
-		obj.Tag = toMapString(v)
+		obj.Tags = toMapString(v)
	}
+
+	if v, ok := in["tags"].(map[string]interface{}); ok && len(v) > 0 {
+		obj.Tags = toMapString(v)
+	}
```

website/docs/r/cluster.html.markdown
```diff
@@ -672,7 +672,7 @@ The following arguments are supported:
 * `network_policy` - (Optional) Network policy used for building Kubernetes network. Chooses from `calico` (string)
 * `pod_cidr` - (Optional) A CIDR notation IP range from which to assign Kubernetes Pod IPs when \"network plugin\" is specified in \"kubenet\". Default `172.244.0.0/16` (string)
 * `service_cidr` - (Optional) A CIDR notation IP range from which to assign Kubernetes Service cluster IPs. It must not overlap with any Subnet IP ranges. Default `10.0.0.0/16` (string)
-* `tag` - (Optional/Computed) Tags for Kubernetes cluster. For example, foo=bar (map)
+* `tags` - (Optional/Computed) Tags for Kubernetes cluster. For example, foo=bar (map)
```
rawmind0, line 683: Maintain tag and tags arguments for a while

```
-* `tag` - (Optional/Computed) Tags for Kubernetes cluster. For example, foo=bar (map)
+* `tag` - (Optional/Computed) Tags for Kubernetes cluster. For example, foo=bar. Deprecated, use `tags` argument instead (map)
+* `tags` - (Optional/Computed) Tags for Kubernetes cluster. For example, foo=bar (map)
```

rancher2/structure_cluster_aks_config_test.go
```diff
@@ -90,7 +102,7 @@ func init() {
 			"service_cidr":                           "service_cidr",
 			"subnet":                                 "subnet",
 			"subscription_id":                        "subscription_id",
-			"tag": map[string]interface{}{
+			"tags": map[string]interface{}{
 				"tag1": "value1",
 				"tag2": "value2",
 			},
```
rawmind0, line 112: Maintain `tag` and `tags` arguments for a while.

```
			"tag": map[string]interface{}{
				"tag1": "value1",
				"tag2": "value2",
			},
+			"tags": map[string]interface{}{
+				"tag1": "value1",
+				"tag2": "value2",
+			},
```

rawmind0 (changes requested):
@dstockton, thanks so much for this PR.

Afaik, `tag` argument is working fine due to json and yaml definition on golang structure. Anyway, naming was clearly a mistake, better to use plural name `tags`. Even though i think would safer to maintain both arguments, `tag` and `tags` for a while.  Please, take a look to the comments. 

Additionally, could you please add new arguments to the provider documentation??

rancher2/structure_cluster_aks_config.go
dstockton: @rawmind0 thanks for the review.
In my testing tag didn’t work because it arrives at rancher/kontainer-engine via the providers flatten method which means it’s flattened to tag and k-e expects the plural - tags.
Therefore “supporting” both would stop the terraform from throwing an error; but still wouldn’t work unless k-e is also updated to support incoming “tag” attribute.
My initial intended approach was to have the provider simply translate tag->tags on flattening; but i don’t think the current architecture will support that type of translation (unless you can maybe suggest otherwise - you will know this code better than I)

rawmind0: > In my testing tag didn’t work because it arrives at rancher/kontainer-engine via the providers flatten method which means it’s flattened to tag and k-e expects the plural - tags.

Yeah, it's flattened to `Tag` on golang struct, but the golang object is marshalled as json format before make the Rancher API call. It's defined as `tags` on json format, `json:"tags,omitempty" yaml:"tags,omitempty"`

> Therefore “supporting” both would stop the terraform from throwing an error; but still wouldn’t work unless k-e is also updated to support incoming “tag” attribute.
My initial intended approach was to have the provider simply translate tag->tags on flattening; but i don’t think the current architecture will support that type of translation (unless you can maybe suggest otherwise - you will know this code better than I)

I'm suggesting to use `Tags` on golang struct (as you proposed), but map it to `tag` and `tags` terraform arguments for a while. Flatten and expand would translate golang struct `Tags` to `tag` and `tags` tf arguments and the reverse.

dstockton:
@rawmind0 I believe I've addressed all your comments now; it'd be great if you're able to re-review. Many thanks (& have a good weekend!)

rawmind0 (approved):

rawmind0:
@dstockton , thanks for the PR. 
Fix issues #120 #183

dstockton:
Thanks @rawmind0 - much appreciated.
Just need https://github.com/rancher/kontainer-engine/pull/200 & https://github.com/rancher/rancher/pull/24158 now :)

rancher2/schema_cluster_aks_config.go
```diff
@@ -270,11 +277,55 @@ func clusterAKSConfigFields() map[string]*schema.Schema {
...
+			Optional:    true,
+			Description: "List of availability zones to use for the cluster",
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"load_balancer_sku": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Default:     "Basic",
```
rawmind0: As discussed [here](https://github.com/rancher/kontainer-engine/pull/200#discussion_r351636524), value should be "basic" instead of "Basic". Adding validation function 

```
-			Default:     "Basic",
+			Default:     "basic",
+                       ValidateFunc: validation.StringInSlice(clusterAKSLoadBalancerSku, true),
```

rancher2/schema_cluster_aks_config.go
```diff
@@ -14,6 +14,7 @@ var (
 	clusterAKSAgentStorageProfile = []string{"ManagedDisks", "StorageAccount"}
 	clusterAKSNetworkPlugin       = []string{"azure", "kubenet"}
 	clusterAKSNetworkPolicy       = []string{"calico"}
+	clusterAKSAgentPoolType       = []string{"AvailabilitySet", "VirtualMachineScaleSets"}
```
rawmind0, line 17: As just 2 values available to [`LoadBalancerSku`](https://github.com/Azure/azure-sdk-for-go/blob/master/services/containerservice/mgmt/2019-08-01/containerservice/models.go#L68) argument, make sense to add validation to it??

```
+	clusterAKSLoadBalancerSku       = []string{"basic", "standard"}
```

rancher2/schema_cluster_aks_config.go
```diff
@@ -270,11 +277,55 @@ func clusterAKSConfigFields() map[string]*schema.Schema {
...
+			Description: "List of availability zones to use for the cluster",
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"load_balancer_sku": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Default:     "basic",
+			Description: "Load balancer type (must be standard for auto-scaling)",
```
rawmind0: Adding validation to `load_balancer_sku` argument

```
+			ValidateFunc: validation.StringInSlice(clusterAKSLoadBalancerSku, true),
```

dstockton: Done :+1:

rawmind0 (approved):
Thanks @dstockton .

LGTM

rawmind0:
On hold waiting for dependencies rancher/kontainer-engine#200 & rancher/rancher#24158

Issue #20, opened 2019-05-23
[Feature Request] VSphere - Support Datastore Cluster


amioranza:
Hi,
I'm trying to create a new cluster on my VSphere environment and we have a datastore cluster but the node templates only support datastore not datastore cluster. Creating machines in this way we have to do storage migration after the machine was created to not left specific datastore with no free space.

rawmind0:
Hi @amioranza, 

this issue has dependency with https://github.com/rancher/rancher/issues/19706 . Until this issue is not addressed, nothing could be done here.

rawmind0:
Added datastore cluster on tf rpovider v1.7.1 and Rancher v2.3.3

Please, reopen issue if needed

Issue #195, opened 2019-12-03
rancher2_node_templates not recognized after upgrade to 2.3.3


mc-meta:
Hello,

we are currently using rancher2 provider version 1.7.0 and seems that after upgrading rancher from version 2.3.2 to 2.3.3 existing rancher2_node_templates in our rancher cluster are not recognized anymore and disappeared after a 'terraform refresh' from state file, while are still present as API objects.

'terraform apply' will create new node_templates afresh side by side old ones.

rawmind0:
Hi @mc-meta , what is the kind of the affected node templates?? vsphere??

mc-meta:
Hello @rawmind0 , problem seems to be present on vsphere and aws templates (that's all we have right now).

rawmind0:
vsphere node template schema has been updated on Rancher v2.3.3 but not aws. Anyway it shouldn't happens this issue on none of the node templates.

Could you please explain a little bit the steps you follow?? I'd need to reproduce the issue

mc-meta:
hello @rawmind0 

our terraform code is managing to:

1) deploy vsphere machines (vsphere provider)
2) form rke cluster from machines at 1) (rke provider)
3) deploy and bootstrap rancher on 2) (helm + kubernetes + rancher2 providers)
4) prepare needed resources on 3) to deploy new clusters on vsphere and aws (rancher2 provider)
5) create new clusters on vsphere and aws directly from (rancher2 provider)

to reproduce is probably enough to:

a) bootstrap a rancher cluster with v2.3.2
b) create node_templates for vsphere and aws and other needed resources
c) verify idempotency
d) create test aws and vsphere clusters using node_templates at point b)
e) update rancher to v2.3.3
f) verify idempotency again, is all goes as expected you'll experience that node_templates are not recognized anymore and terraform will plan to create new node_templates even if pristine objects are still there

Just for sake of completeness, we have upgraded rancher local cluster from v1.15.2-rancher1-1 to v1.15.3-rancher1-1 few days before upgrade to v2.3.3.

Let us know if you need us to perform any new tests in our lab setup

rawmind0:
Hi @mc-meta , i was able to reproduce the issue.

[...]

PR #197, opened 2019-12-03
Updated vsphere_config argument schema on rancher2_node_template resource

rawmind0:
Fix issues #20, #195


cloudnautique (approved):

website/docs/r/nodeTemplate.html.markdown
```diff
@@ -255,3 +268,15 @@ Node Template can be imported using the Rancher Node Template ID
 $ terraform import rancher2_node_template.foo <node_template_id>
 ```
 
+## Upgrading to Rancher v2.3.3
+
+Due to [this feature](https://github.com/rancher/rancher/pull/23718) included on Rancher v2.3.3, `rancher2_node_template` are now global scope objects with RBAC arounf them, instead of user scope objects as they were. This means that existing node templates are changing their `id` on upgrade, making that terraform provider will try to recreate them due to it's unable to find them by `id`.
```
cloudnautique: This means that existing node templates `id` field is changing on upgrade. Because the Terraform provider can not find the old `id`, it will try to recreate them.

rawmind0: Much better. Updated. :)

website/docs/r/nodeTemplate.html.markdown
```diff
@@ -255,3 +268,15 @@ Node Template can be imported using the Rancher Node Template ID
 $ terraform import rancher2_node_template.foo <node_template_id>
 ```
 
+## Upgrading to Rancher v2.3.3
+
+Due to [this feature](https://github.com/rancher/rancher/pull/23718) included on Rancher v2.3.3, `rancher2_node_template` are now global scope objects with RBAC arounf them, instead of user scope objects as they were. This means that existing node templates `id` field is changing on upgrade. Because the Terraform provider can not find the old `id`, it will try to recreate them.
```
apricote:
```suggestion
Due to [this feature](https://github.com/rancher/rancher/pull/23718) included on Rancher v2.3.3, `rancher2_node_template` are now global scope objects with RBAC around them, instead of user scope objects as they were. This means that existing node templates `id` field is changing on upgrade. Because the Terraform provider can not find the old `id`, it will try to recreate them.
```

rawmind0: Thanks @apricote. Typo fixed.

cloudnautique (approved):

PR #199, opened 2019-12-03
Add Not Null (API states has-value) to Alerts/Metric

Bourne-ID:
Rancher Alerts allow for metric rule of Not Null (API states this is has-value so mapped accordingly). 

Tested against Rancher 2.3.3


rawmind0 (commented):
Thanks for the PR @Bourne-ID .

Could you please add new comparison value to [clusterAlertRule doc](https://github.com/terraform-providers/terraform-provider-rancher2/blob/master/website/docs/r/clusterAlertRule.html.markdown#arguments-1)

Bourne-ID:
Updates to project and cluster alerts documentation to include has-value option.

rawmind0 (approved):
LGTM

PR #221, opened 2020-01-08
add windows_prefered_cluster option to cluster schema/structure

zparnold:
This PR adds the ability to set the `windows_prefered_cluster` field 

It also copies the spelling from the `cluster_template` files. (The word "preferred" is misspelled.) I wanted to keep it consistent however. If I should fix the spelling, please let me know.

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -2,6 +2,7 @@ package rancher2
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/customdiff"
 	"log"
 	"time"
 
@@ -27,6 +28,9 @@ func resourceRancher2Cluster() *schema.Resource {
 			Update: schema.DefaultTimeout(30 * time.Minute),
 			Delete: schema.DefaultTimeout(30 * time.Minute),
 		},
+		CustomizeDiff: customdiff.All(
+			windowsPreferredClusterCustomDiffHandler(),
+		),
 	}
 }
 
@@ -397,4 +401,13 @@ func createClusterRegistrationToken(client *managementClient.Client, clusterID s
 		return nil, fmt.Errorf("[ERROR] waiting for cluster registration token (%s) to be created: %s", newRegToken.ID, waitErr)
 	}
 	return newRegToken, nil
+}
+
+func windowsPreferredClusterCustomDiffHandler()  schema.CustomizeDiffFunc{
+	//Cluster must be recreated to implement this if there is a change to windows var if the cluster is not
+	//already a windows cluster. Currently this is unsupported by rancher, so force new resource here.
+	//https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/windows-clusters/
+	return customdiff.ForceNewIfChange("windows_prefered_cluster", func (old, new, meta interface{}) bool {
+		return new.(bool) != old.(bool)
+	})
 }
--- a/rancher2/schema_cluster.go
+++ b/rancher2/schema_cluster.go
@@ -274,6 +274,12 @@ func clusterFields() map[string]*schema.Schema {
 			Optional: true,
 			Computed: true,
 		},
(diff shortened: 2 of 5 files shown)


rancher2/structure_cluster.go
```diff
@@ -179,6 +179,11 @@ func flattenCluster(d *schema.ResourceData, in *Cluster, clusterRegToken *manage
 		return err
 	}
 
+	err = d.Set("windows_prefered_cluster", in.WindowsPreferedCluster)
+	if err != nil {
+		return err
+	}
+
```
rawmind0, line 183: Bool fields set don't need to be checked

```
-	err = d.Set("windows_prefered_cluster", in.WindowsPreferedCluster)
-	if err != nil {
-		return err
-	}
+	d.Set("windows_prefered_cluster", in.WindowsPreferedCluster)
```

rancher2/schema_cluster.go
```diff
@@ -274,6 +274,12 @@ func clusterFields() map[string]*schema.Schema {
 			Optional: true,
 			Computed: true,
 		},
+		"windows_prefered_cluster": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Windows prefered cluster",
```
rawmind0: As rancher/types field name is `WindowsPreferedCluster` to be consistent don't need to change tf field name, but description should be updated

```
-			Description: "Windows prefered cluster",
+			Description: "Windows preferred cluster",
```

rawmind0 (changes requested):
@zparnold thanks for the PR. 

Could you please take a look to changes requested?? 

Also, could you please update `rancher2_cluster` resource docs properly, adding new field??

rancher2/structure_cluster.go
zparnold: Sounds good 😄

rancher2/schema_cluster.go
zparnold: 👍

zparnold:
@rawmind0 Made changes you requested. Also did a force push so there will only be one commit when merged. 😄 Let me know if there's anything else I need to do.

rawmind0 (changes requested):
@zparnold thanks for the update. 

Last thing, could you please add new field to be updated on `resourceRancher2ClusterUpdate` function on `resource_rancher2_cluster.go` and to be tested on `structure_cluster_test.go`??

zparnold:
@rawmind0 Addressed your concerns, and welcome your feedback.

rancher2/resource_rancher2_cluster.go
```diff
@@ -398,3 +402,12 @@ func createClusterRegistrationToken(client *managementClient.Client, clusterID s
...
 }
+
+func windowsPreferredClusterCustomDiffHandler()  schema.CustomizeDiffFunc{
+	//Cluster must be recreated to implement this if there is a change to windows var if the cluster is not
+	//already a windows cluster. Currently this is unsupported by rancher, so force new resource here.
+	//https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/windows-clusters/
+	return customdiff.ForceNewIfChange("windows_prefered_cluster", func (old, new, meta interface{}) bool {
+		return new.(bool) != old.(bool)
+	})
+}
```
rawmind0: If a new object needs to be generated on this field change, i think is just enough to add `Forcenew: true` on the field definition. Not needed a cutomizediff function.

rawmind0 (changes requested):
Please, take a look on the new comments.

Also, build is failing, please execute `make fmt` and  `make test` before commit.

rancher2/resource_rancher2_cluster.go
zparnold: Gotcha, thanks!

zparnold:
@rawmind0 Hopefully this should address that concern.

rawmind0 (approved):

rawmind0:
LGTM

PR #229, opened 2020-01-19
Dns policy ingress for RKE

Bourne-ID:
Adds the dnsPolicy to the RKE Ingress section when creating a cluster. Documentation updated.

### DIFF
--- a/rancher2/schema_cluster_rke_config_ingress.go
+++ b/rancher2/schema_cluster_rke_config_ingress.go
@@ -2,6 +2,23 @@ package rancher2
 
 import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+const (
+	clusterRKEConfigIngressDNSPolicyClusterFirst            = "ClusterFirst"
+	clusterRKEConfigIngressDNSPolicyClusterFirstWithHostNet = "ClusterFirstWithHostNet"
+	clusterRKEConfigIngressDNSPolicyDefault                 = "Default"
+	clusterRKEConfigIngressDNSPolicyNone                    = "None"
+)
+
+var (
+	clusterRKEConfigIngressDNSPolicyList = []string{
+		clusterRKEConfigIngressDNSPolicyClusterFirst,
+		clusterRKEConfigIngressDNSPolicyClusterFirstWithHostNet,
+		clusterRKEConfigIngressDNSPolicyDefault,
+		clusterRKEConfigIngressDNSPolicyNone,
+	}
 )
 
 //Schemas
@@ -29,9 +46,10 @@ func clusterRKEConfigIngressFields() map[string]*schema.Schema {
 			Computed: true,
 		},
 		"dns_policy": {
-			Type:     schema.TypeString,
-			Optional: true,
-			Computed: true,
+			Type:         schema.TypeString,
+			Optional:     true,
+			Computed:     true,
+			ValidateFunc: validation.StringInSlice(clusterRKEConfigIngressDNSPolicyList, true),
 		},
 	}
 	return s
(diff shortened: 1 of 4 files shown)


website/docs/r/cluster.html.markdown
```diff
@@ -450,6 +450,7 @@ The following attributes are exported:
 
 ##### Arguments
 
+* `dns_policy` - (Optional/Computed) DNS Policy for RKE Ingress (map)
```
rawmind0: Wrong argument type. Also adding supported values.

```
- * `dns_policy` - (Optional/Computed) DNS Policy for RKE Ingress (map)
+ * `dns_policy` - (Optional/Computed) Ingress controller DNS policy. `ClusterFirstWithHostNet`, `ClusterFirst`, `Default`, and `None` are supported. [K8S dns Policy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) (string)
```

rancher2/schema_cluster_rke_config_ingress.go
```diff
@@ -28,6 +28,11 @@ func clusterRKEConfigIngressFields() map[string]*schema.Schema {
 			Optional: true,
 			Computed: true,
 		},
+		"dns_policy": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+		},
```
rawmind0, line 53: What do you think to add argument validation to limit the value of `dns_policy` just to available at [K8S dns Policy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) ??

```
import (
	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
)

+const (
+	clusterRKEConfigIngressDNSPolicyClusterFirst = "ClusterFirst"
+	clusterRKEConfigIngressDNSPolicyClusterFirstWithHostNet = "ClusterFirstWithHostNet"
+	clusterRKEConfigIngressDNSPolicyDefault = "Default"
+	clusterRKEConfigIngressDNSPolicyNone = "None"
+)
+
+var (
+	clusterRKEConfigIngressDNSPolicyList = []string{
+		clusterRKEConfigIngressDNSPolicyClusterFirst, 
+		clusterRKEConfigIngressDNSPolicyClusterFirstWithHostNet,
+		clusterRKEConfigIngressDNSPolicyDefault,
+		clusterRKEConfigIngressDNSPolicyNone
+	}
+)
...
		"dns_policy": {
			Type:     schema.TypeString,
			Optional: true,
			Computed: true,
+			ValidateFunc: validation.StringInSlice(clusterRKEConfigIngressDNSPolicyList, true),
		},
```

rawmind0 (changes requested):
@Bourne-ID , thanks for the PR. 

Could you please take a look to the changes requested??

rancher2/schema_cluster_rke_config_ingress.go
Bourne-ID: I thought about this but hesitated as tests on the Rancher API shows it accepts any string (even if this means the cluster fails as the dns policy is not recognised) - is it correct to add restrictions on the Terraform provider when there are no restrictions on Rancher? I'm for this change, but at the same time don't want to cause a limitation if there has been a decision to make this an any-entry string on Rancher...

website/docs/r/cluster.html.markdown
Bourne-ID: Great catch - fixed and agreed with supported values + link.

rancher2/schema_cluster_rke_config_ingress.go
Oats87: The Rancher API accepts anything but RKE itself should not: https://github.com/rancher/rke/blob/master/cluster/validation.go#L171

Bourne-ID: Changes added

rawmind0 (approved):
LGTM

PR #251, opened 2020-02-09
Make kube_config sensitive

### LINKED ISSUE #250: Auth token exposure: rancher2_cluster kube_config attribute should be sensitive
The `kube_config` attribute of the `rancher2_cluster` resource contains administrator credentials, and it is printed to console in plaintext in the Terraform plan. Terraform has a [Sensitive flag](https://www.terraform.io/docs/extend/best-practices/sensitive-state.html#using-the-sensitive-flag) that should be used for this attribute.

armsnyder:
Closes #250

The kube_config attribute of the rancher2_cluster resource contains administrator credentials, and it is printed to console in plaintext in the Terraform plan. By making it sensitive, it will not be printed.

### DIFF
--- a/rancher2/schema_cluster.go
+++ b/rancher2/schema_cluster.go
@@ -113,8 +113,8 @@ func clusterFields() map[string]*schema.Schema {
 			ValidateFunc: validation.StringInSlice(clusterDrivers, true),
 		},
 		"kube_config": &schema.Schema{
-			Type:     schema.TypeString,
-			Computed: true,
+			Type:      schema.TypeString,
+			Computed:  true,
 			Sensitive: true,
 		},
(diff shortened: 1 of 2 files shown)


rawmind0 (changes requested):
@armsnyder , thanks for the PR.

Could you please update cluster doc [attributes-reference](https://github.com/terraform-providers/terraform-provider-rancher2/blob/master/website/docs/r/cluster.html.markdown#attributes-reference) accordingly??

armsnyder:
@rawmind0 Sure, updated.

rawmind0 (approved):

rawmind0:
@armsnyder , thanks for the update.

LGTM

Issue #256, opened 2020-02-18
API server does not start because of wrong audit log policy


czunker:
Rancher Version: 2.3.5
terraform versions:
Terraform v0.12.20
+ provider.aws v2.48.0
+ provider.rancher2 v1.7.2
k8s version: 1.17.2

I tried to setup a cluster with audit logs enabled, but the api server didn't start.
I used the example provided with [docs](https://www.terraform.io/docs/providers/rancher2/r/cluster.html#example-usage):

    services {
      kube_api {
        audit_log {
          enabled = true
          configuration {
            max_age = 5
            max_backup = 5
            max_size = 100
            path = "-"
            format = "json"
            policy = jsonencode({"rules":[{"level": "Metadata"}]})
          }
        }
      }
    }

This results in the following error:

> I0212 12:31:11.888844    1 server.go:149] Version: v1.16.6
> 
> Error: loading audit policy file: unknown group version field audit.k8s.io/__internal, Kind=Policy in policy: from file /etc/kubernetes/audit-policy.yaml

The audit-policy.yaml looks like this:

    metadata:
      creationTimestamp: null
    rules:
    - level: Metadata

The missing pieces are:

    apiVersion: audit.k8s.io/v1
    kind: Policy

Adding this manually, the api server starts.
I also tried to add it in terraform:

    services {
          kube_api {
                audit_log {
                    enabled = true
                    configuration {
                        max_age = 5
                        max_backup = 5
                        max_size = 100
                        path = "-"
                        format = "json"
                        policy = jsonencode({"rules":[{"level": "Metadata"}], "apiVersion": "audit.k8s.io/v1", "kind": "Policy"})
                    }
                }
          }
    }

But it didn't get passed into the file.

Removing the line

    policy = jsonencode(....

The api server starts and the audit-policy.yaml is generated correctly:

    apiVersion: audit.k8s.io/v1
    kind: Policy
    metadata:
      creationTimestamp: null
    rules:
    - level: Metadata

rawmind0:
Reproduced issue. Submitted PR #257 to fix it.

rawmind0:
@czunker , finally `policy` has been implemented in yaml format. Take a look to example, https://github.com/terraform-providers/terraform-provider-rancher2/pull/257/files#diff-e06e84c47d0fb7fa4ee40b06919f98ecR28

rawmind0:
PR merged and will be included on next release. Please, reopen issue if needed.

PR #257, opened 2020-02-18
Fix audit_log.configuration.policy argument to `rke_config.services.kube_api` argument

rawmind0:
Fix issue #256


website/docs/r/cluster.html.markdown
```diff
@@ -43,7 +43,7 @@ resource "rancher2_cluster" "foo-custom" {
             max_size = 100
             path = "-"
             format = "json"
-            policy = jsonencode({"rules":[{"level": "Metadata"}]})
+            policy = jsonencode({"apiVersion":"audit.k8s.io/v1","kind":"Policy","rules":[{"level":"RequestResponse","resources":[{"group":"","resources":["namespaces"]}]}]})
```
cloudnautique: Can we do this another way, that feels kind of ugly. Should we recommend the user create a file and read in the yaml snippet?

cloudnautique (changes requested):

website/docs/r/cluster.html.markdown
rawmind0: Agreed. Better now??

cloudnautique: Why are we reading JSON and not YAML? Writing JSON is not fun for a human.

rawmind0: Agreed, but policy is expected with this format. Tell me if more changes are needed

rawmind0: Ok, you win. Updated to yaml format. :)

cloudnautique (approved):

Issue #295, opened 2020-04-01
Upgrade strategy failed


nickvth:
Error: [ERROR] waiting for cluster (c-556rs) to be updated: unexpected state 'upgrading', wanted target 'active, provisioning, pending'. last error: %!s(<nil>)

rawmind0:
PR included on release v1.8.2. Please reopen issue if needed.

PR #296, opened 2020-04-01
Fix upgrading state on resourceRancher2ClusterUpdate(), rke supported k8s version and catalog version

rawmind0:
Fix issues #295, #298 and #299


rancher2/config.go
```diff
@@ -2,10 +2,15 @@ package rancher2
 
 import (
 	"fmt"
+<<<<<<< HEAD
```
luthermonson: conflicts yo

rawmind0: Oh, i let last commit without push. Updated

cloudnautique (approved):

PR #301, opened 2020-04-02
Fix typo with node template ID fix function.

patrickdappollonio:
Hello!

While looking at the fix for node template IDs I noticed the function name has a typo.

This MR changes all occurrences from `Tempate` to `Template`.


rawmind0 (approved):
Good catch @patrickdappollonio , thanks for your PR.

LGTM

Issue #305, opened 2020-04-03
Error: Timeout, Rancher is not ready: <nil>


ahmedfourti:
Hello,

It seems the new version `1.8.2` doesn't work as expected.
Here is my code for provider

provider "rancher2" {
  alias = "bootstrap"

  api_url   = "https://${var.rancher_hostname}"
  bootstrap = true
}

The variable `rancher_hostname` is built from an outputs of infrastructure.

However, it hangs for some time and then it throws error `Error: Timeout, Rancher is not ready: <nil>`
I tried the retries option as explained here  #292  but still the same issue.

With version `1.8.1` it works fine.

mcmcghee:
Same issue here.

rawmind0:
Got it. Investigating the issue.

rawmind0:
Hi all, could you please provide the tf file your are using??

ahmedfourti:
Here is mine

```
provider "rancher2" {
  alias = "bootstrap"

  api_url   = "https://${var.rancher_hostname}"
  bootstrap = true
}

resource "rancher2_bootstrap" "admin" {
  provider = rancher2.bootstrap

  current_password = var.rancher_current_password
  password         = var.rancher_password
}
```

rawmind0:
@ahmedfourti , is your terraform apply stuck before plan or it's failing when applying?? If it's failing on applying plan, what is the `retries` value you have tried??

ahmedfourti:
@rawmind0  it fails when it tries to refresh, so no matter plan or apply, it fails.
I tried the default retries to 5, then 50 then 100.

rawmind0:
I'm asking you because matters if plan or apply. If stuck before plan, retries doesn't matter, but if fails applying it's a question of retries. It's a problem how terraform init provdiers.

Anyway, what is the initial value of `${var.rancher_hostname}`?? Is it empty and would be generated during same apply??

ahmedfourti:
It is stuck during refresh, so before both plan and apply.
Yes the variable is empty and generated later during the apply

rawmind0:
PR #306 will address this issue. In the meanwhile we test and release, please continue using provider v1.8.1

rawmind0:
PR merged. Fix will be included on v1.8.3. Please reopen issue if needed

Issue #303, opened 2020-04-02
Cluster / Project Monitoring Version Upgrade


toddexp:
It seems there is currently no way to upgrade the cluster or project monitoring from the Terraform provider. I don't see any version specified in the Terraform code. Reissuing the cluster or project monitoring code again doesn't upgrade it either.

It seems that the monitoring version is set by the helm template version that is deployed for the cluster / project monitoring.

Is there a way to upgrade the monitoring version via this Terraform code?

rawmind0:
PR #306 will add this feature

toddexp:
Thank you for addressing this so quickly. I took a quick look at the PR. It appears this will allow versioning for the cluster monitoring. Is it possible to also allow versioning for project monitoring as well? It's possible I missed this ability for the project monitoring when looking at the PR though.

rawmind0:
Take a look now.

toddexp:
Looks great. Thank you again for fixing this so quickly. Much appreciated.

rawmind0:
PR merged. Fix will be included on v1.8.3. Please reopen issue if needed

PR #306, opened 2020-04-03
Fixed rancher timeout on bootstrapping and added monitoring version on rancher2_cluster

rawmind0:
Fix issues #307 #305 and #303

### DIFF
--- a/website/docs/r/cluster.html.markdown
+++ b/website/docs/r/cluster.html.markdown
@@ -96,6 +96,7 @@ resource "rancher2_cluster" "foo-custom" {
       "prometheus.resources.core.requests.memory" = "750Mi"
       "prometheus.retention" = "12h"
     }
+    version = "0.1.0"
   }
 }
 ```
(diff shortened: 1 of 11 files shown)


website/docs/r/cluster.html.markdown
```diff
@@ -96,6 +96,7 @@ resource "rancher2_cluster" "foo-custom" {
       "prometheus.resources.core.requests.memory" = "750Mi"
       "prometheus.retention" = "12h"
     }
+    version = "0.1.0"
```
cloudnautique, line 99: can we use a more modern version, if people copy and paste this seems to old.

rawmind0: rancher-monitoring `v0.1.0` is the latest versions by the moment, https://github.com/rancher/system-charts/tree/dev-v2.5/charts/rancher-monitoring

cloudnautique: Oops, you are right, we went from 0.0.8 or something :)

cloudnautique (approved):

PR #326, opened 2020-05-05
Fix typo in documentation of GKE enable_network_policy_config field

bashofmann:
There is a copy&paste error in the docs for the GKE `enable_network_policy_config` field.


bashofmann:
The flag is actually not working at the moment. See https://github.com/rancher/rancher/issues/26917.

rawmind0 (approved):
LGTM

Issue #404, opened 2020-07-31
event_rate_limit issue on cluster config


zaggash:
Trying to import a cluster with some existing event_rate_limit config in the cluster.yaml.
It failed with the following error:

```
# terraform import rancher2_cluster.custom c-n6ds2                                                                                                                                                                rancher2_cluster.custom: Importing from ID "c-n6ds2"...

Error: rke_config.0.services.0.kube_api.0.event_rate_limit.0.configuration.limits: '' expected type 'string', got unconvertible type '[]interface {}'
```

If we try to update/create a cluster with event_rate_limit, we cannot set the `limits` values in the `configuration`
There are no mentions of how to set this up in the documentation.
Trying to validate the file, we have the following output:
```
"Incorrect attribute value type", specified as "Inappropriate value for attribute "configuration": element "limits": string
```

```
event_rate_limit {
          configuration = {
            api_version = "eventratelimit.admission.k8s.io/v1alpha1"
            kind = "Configuration"
            limits = {
              burst = 30000
              qps = 6000
              type = "Server"
            }
          }
          enabled = false
        }
```

I tried to do the following, but the admission.yaml get overwritten by the default values.
```
  provisioner "file" {
    source      = "admission.yaml"
    destination = "/etc/kubernetes/admission.yaml"
  }

  rke_config {
    services {
      kube_api {
        event_rate_limit {
          enabled = true
        }
      }
    }
  }
```

gz#11331

rawmind0:
Issue has been reproduced. Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/405 to address the issue. `event_rate_limit.configuration` should be passed in yaml format.

event_rate_config.yaml
```
apiVersion: eventratelimit.admission.k8s.io/v1alpha1
kind: Configuration
limits:
- type: Server
  burst: 35000
  qps: 6000
```

```
...
        event_rate_limit {
          configuration = file("event_rate_config.yaml")
          enabled = true
        }
...
```

rawmind0:
@zaggash , PR is merged to master. branch Please, reopen issue if needed.

zaggash:
Hello @rawmind0 
Thanks for fixing it.

I think the same issue apply for the kubeapi `secrets_encryption_config` configuration item.

rawmind0:
@zaggash, PR #423 should fix `secrets_encryption_config` custom config issue

rawmind0:
@zaggash , PR #423 merged into master

PR #405, opened 2020-08-04
Fix rke_config.event_rate_limit.configuration argument to work properly

rawmind0:
Fix issue https://github.com/rancher/terraform-provider-rancher2/issues/404


paraglade (approved):
I approve

Issue #389, opened 2020-07-18
Cluster Members Cannot Create Projects With Monitoring Enabled


armsnyder:
This applies when using the Rancher2 provider with an API token generated from a user with the Cluster Member role. They are able to create new `rancher2_project` resources when only the `cluster_id` and `name` are specified, but they encounter an error `Error: action [editMonitoring] not available` if they specify `enable_project_monitoring = true`.

This is inconsistent with the behavior of using the UI, where Cluster Members can turn on cluster monitoring.

Terraform shows the diff successfully, but after confirming the apply, it errors.

Strangely, the project is successfully created with monitoring enabled. However, Terraform returns an error, and the resource is not added to Terraform state.

Another oddity is that there is a workaround -- the user can apply a `rancher2_project` resource with `enable_project_monitoring = false` first. Then update it by setting `enable_project_monitoring = true`, and the apply will succeed.

Example config:

```hcl
provider "rancher2" {
  token_key = "token-redacted"
  api_url = "https://redacted"
}

resource "rancher2_project" "foo" {
  cluster_id = "c-redacted"
  name = "foo"
  enable_project_monitoring = true
  project_monitoring_input {
    answers = {
      "operator.resources.limits.memory"          = "500Mi"
      "exporter-kubelets.https"                   = true
      "exporter-node.enabled"                     = true
      "exporter-node.ports.metrics.port"          = 9796
      "exporter-node.resources.limits.cpu"        = "200m"
      "exporter-node.resources.limits.memory"     = "200Mi"
      "grafana.persistence.enabled"               = true
      "grafana.persistence.size"                  = "10Gi"
      "grafana.persistence.storageClass"          = "default"
      "prometheus.persistence.enabled"            = true
      "prometheus.persistence.size"               = "10Gi"
      "prometheus.persistence.storageClass"       = "default"
      "prometheus.persistent.useReleaseName"      = "true"
      "prometheus.resources.core.limits.cpu"      = "1000m",
      "prometheus.resources.core.limits.memory"   = "1500Mi"
      "prometheus.resources.core.requests.cpu"    = "750m"
      "prometheus.resources.core.requests.memory" = "750Mi"
      "prometheus.retention"                      = "168h"
    }
  }
}
```

Full log after confirming the apply:

```
Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

[...]

Issue #407, opened 2020-08-05
Project Monitoring doesnt upgrade


Arola1982:
When changing versions of the monitoring terraform shows it modifying the project resource, but in actual fact the resource in rancher stays on the current version.

Shouldnt this trigger a redeploy of the later version?

armsnyder:
Possible duplicate of #376?

rawmind0:
PR #409 is merged to master and should address this issue. Please reopen it if needed.

Issue #398, opened 2020-07-27
rancher2_token does not apply labels


Fresa:
The documentation indicates that labels can be defined as arguments.
https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/token

When defining labels, they are identified in the plan but they never get applied onto the token resource in kubernetes,.

```
resource "rancher2_bootstrap" "admin" {
  provider = rancher2.bootstrap
  password = "admin"
}

resource "rancher2_token" "test" {
  provider = rancher2.admin
  cluster_id = "local"
  labels = {
    "foo" = "bar"
  }
}
```

![image](https://user-images.githubusercontent.com/1190300/88539666-f617d580-d011-11ea-83c4-5a7593e4a6eb.png)

`$ kubectl get token/token-g6gkp -o yaml:`
![image](https://user-images.githubusercontent.com/1190300/88539421-80ac0500-d011-11ea-8cac-d2723b2c2c4a.png)

I have tried setting labels on the rancher2_namespace resource, and that works as expected.

Versions
======
```
$ terraform version
Terraform v0.12.29
+ provider.rancher2 v1.9.0
```
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:52:00Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

```
```
$ docker version
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b7f0
 Built:             Wed Mar 11 01:25:46 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
[...]

Issue #376, opened 2020-06-25
Monitoring version upgrade


spencergilbert:
Related to #306

We just added the version key, and set it to "0.1.1" while our clusters are running "0.1.0" expecting this to upgrade the monitoring app.

However it did not, is this because because the existing terraform resource had no version set?

rawmind0:
PR #409 is merged to master and should address this issue. Please reopen it if needed.

PR #409, opened 2020-08-05
Fix rancher2_cluster and rancher2_project resources to be used by not global admin user

rawmind0:
Fix issues #389, #333, #376, #398, #407 and #414


paraglade (approved):
approved

PR #423, opened 2020-08-20
Fix rke_config.secrets_encryption_config.custom_config argument to work properly

rawmind0:
Related to #404


paraglade (approved):
approved

PR #456, opened 2020-09-14
Fix Error: string is required upgrading rancher2 provider from v1.10.0 or lower

### LINKED ISSUE #435: Error: string is required when updating from 1.10.0 to 1.10.1
`Error: string is required` with `terraform plan` when we update from `1.10.0` to `1.10.1`

We are still using Terraform v0.12.x

Currently, trying to find which string is required.

rawmind0:
Closes #435


paraglade (approved):
approved

PR #468, opened 2020-09-28
Added new computed ca_cert argument at rancher2_cluster. Fixed `rke_config.monitoring.replicas` argument

### LINKED ISSUE #339: terraform-rancher2(RFE):Expose caCert in the rancher2_cluster data source
Details:
The rancher2 terraform provider does not expose the caCert in the attributes of either the rancher2_cluster resource, or more importantly, the rancher2_cluster data source. 

It may not make sense to add it to the resource, since it can take some time (1-2 minutes) for the value to appear. But it should exist in the data source.

The caCert value is available in the Rancher API when querying a cluster. That's the value that is needed. (This feature request would only need to apply to the terraform provider.)
FYI:
[https://www.terraform.io/docs/providers/rancher2/d/cluster.html](url)

### LINKED ISSUE #438: can no longer delete namespaces after upgrading from v1.8.3 to v1.9
After upgrading from v1.8.3 (where namespaces are successfully destroyed) to v1.9 of the rancher2 TF provider, destroying namespaces is now timing out.

There were no other changes made to the TF script besides changing the version number of the provider. These actions are performed using the admin credentials created by the rancher2 bootstrap. In the Rancher UI, these namespaces become stuck on the "Removing" status.

screenshot attached of debug output from v1.9.0 timeouts when attempting to destroy namespaces

image

gz#11485

### LINKED ISSUE #464: Suggestion: rancher2_app should be deleted if the installation fails
When a rancher2_app resource is applied for the first time, but the installation fails within the default 10 minute timeout, the rancher2 provider throws an error like this:

`Error: [ERROR] waiting for app (p-7sxs7:someapp) to be created: timeout while waiting for state to become 'active' (last state: 'installing', timeout: 10m0s)`

Now after fixing the actual cause (E.g. a faulty docker image tag in the values.yaml) and when re-running the deployment, the subsequent run will fail with the following error:

`Error: Bad response statusCode [409]. Status [409 Conflict]. Body: [message=apps.project.cattle.io "someapp" already exists, baseType=error, code=AlreadyExists] from [https://rancher-url.foo/v3/project/c-5spdc:p-7sxs7/apps]`

It would be awesome to have an argument for the resource like _delete_on_failure_ = true|false or just taint the terraform resource so that it will be deleted before the next run.

rawmind0:
This PR provides:
* Enhancement: Added new computed `ca_cert` argument at `rancher2_cluster` resource and datasource. Closes #339 
* Fix: Fixed `rke_config.monitoring.replicas` argument to set default value to 1 if monitoring enabled. Closes #438 
* Enhancement: Delete `rancher2_app` if created and got timeout to be active. Closes #464


paraglade (approved):
lgtm

PR #480, opened 2020-10-20
Expose app timeout

GaryWL:
Expose the application timeout, as the default value of 300 seconds, 5
minutes is sometimes insufficient.

Signed-off-by: Gary Lockyer <gary@catalyst.net.nz>

### DIFF
--- a/rancher2/structure_app.go
+++ b/rancher2/structure_app.go
@@ -158,5 +158,7 @@ func expandApp(in *schema.ResourceData) (*projectClient.App, error) {
 		obj.Labels = toMapString(v)
 	}
 
+	obj.Timeout = int64(in.Timeout(schema.TimeoutCreate).Seconds())
+
 	return obj, nil
 }
--- a/rancher2/structure_app_test.go
+++ b/rancher2/structure_app_test.go
@@ -40,6 +40,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 1200,
 	}
 	testAppInterfaceGlobal = map[string]interface{}{
 		"catalog_name": "test",
@@ -85,6 +86,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 1200,
 	}
 	testAppInterfaceClusterLocal = map[string]interface{}{
 		"catalog_name":     "local:test",
@@ -129,6 +131,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 1200,
 	}
 	testAppInterfaceCluster = map[string]interface{}{
 		"catalog_name":     "c-XXXXX:test",
@@ -173,6 +176,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 1200,
 	}
 	testAppInterfaceProject = map[string]interface{}{
 		"catalog_name":     "p-XXXXX:test",


rawmind0 (commented):
I think would be better option to use [terraform resource operation timeout](https://www.terraform.io/docs/configuration/resources.html#operation-timeouts), instead of adding a new argument. What do you think??

simonsCatalyst:
Hi @rawmind0 ,

The terraform resource operation timeout does something different. That's the amount of time the terraform operation is taking to happen, what Gary's patch is doing is allowing the setting of the Helm timeout ( https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback ). This is the amount of time Helm - which is being executed by rancher - is waiting for an operation to happen.

The problem we are trying to solve here is we have Helm Chart hooks ( https://helm.sh/docs/topics/charts_hooks/ ) that are taking longer that 5 minutes to complete - these are happening asynchronously from the app creation in Rancher. A successful app creation in rancher returns to Terraform an OK before the chart hooks have completed.

So Terraform completes successfully, but once the default 5 minute timeout for hooks has been reached Rancher decides the app installation has failed if the chart hooks have not completed. This timeout is configurable in the Rancher UI, Gary's patch would allow it to be configured via Terraform.

Setting `wait` ( https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/app#wait ) has no effect here.

I hope this makes sense

Simon

rawmind0:
Hi @simonsCatalyst , yes, i know the difference between terraform resource operation timeout and the helm chart deployment timeout, but helm chart timeout could be less or the same that terraform resource operation timeout. The PR makes all the sense, i'm just proposing to use `terraform resource operation timeout` as value to pass to helm chart deployment timeout, instead create a new argument.

GaryWL:
@rawmind0  what needs to be done to do this? as I'm still finding my way around the code. Do I need to add a resource?

rancher2/structure_app.go
```diff
@@ -158,5 +161,9 @@ func expandApp(in *schema.ResourceData) (*projectClient.App, error) {
 		obj.Labels = toMapString(v)
 	}
 
+	if v, ok := in.Get("timeout").(int); ok && v > 0 {
+		obj.Timeout = int64(v)
+	}
```
rawmind0: To use terraform operation timeout, obj.Timeout should get the value from resource

```
-	if v, ok := in.Get("timeout").(int); ok && v > 0 {
-		obj.Timeout = int64(v)
-	}
+ obj.Timeout = int64(in.Timeout(schema.TimeoutCreate).Seconds())
```

rancher2/schema_app.go
```diff
@@ -86,6 +86,12 @@ func appFields() map[string]*schema.Schema {
 			Default:     true,
 			Description: "Wait until app is deployed and active",
 		},
+		"timeout": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     300,
+			Description: "Application deployment step timeout",
+		},
```
rawmind0: This change is not needed to use terraform operation timeout

rancher2/structure_app.go
```diff
@@ -84,6 +84,9 @@ func flattenApp(d *schema.ResourceData, in *projectClient.App) error {
 		return err
 	}
 
+	if in.Timeout > 0 {
+		d.Set("timeout", in.Timeout)
+	}
```
rawmind0: This change is not needed to use terraform operation timeout

rancher2/structure_app_test.go
```diff
@@ -40,6 +40,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 1800,
```
rawmind0: To use terraform operation timeout, 
```
-		Timeout: 1800,
+		Timeout: 1200,
```

rancher2/structure_app_test.go
```diff
@@ -85,6 +87,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		Timeout: 300,
```
rawmind0: To use terraform operation timeout, 
```
-		Timeout: 1800,
+		Timeout: 1200,
```

rancher2/structure_app_test.go
```diff
@@ -129,6 +132,7 @@ func init() {
...
```
rawmind0: To use terraform operation timeout, 
```
-		Timeout: 1800,
+		Timeout: 1200,
```

rancher2/structure_app_test.go
```diff
@@ -173,6 +177,7 @@ func init() {
...
```
rawmind0: To use terraform operation timeout, 
```
-		Timeout: 1800,
+		Timeout: 1200,
```

rancher2/structure_app_test.go
```diff
@@ -64,6 +65,7 @@ func init() {
 			"option1": "value1",
 			"option2": "value2",
 		},
+		"timeout": 1800,
```
rawmind0: This change is not needed to use terraform operation timeout

rawmind0 (changes requested):
@GaryWL , take a look to requested changes to use terraform operation timeout

GaryWL:
Thanks @rawmind0 that makes sense.

rawmind0 (approved):
LGTM

rawmind0:
Acceptance tests are failing due to dns issues. Passing fine locally.

PR #481, opened 2020-10-23
Added testAndApply function to openldap, freeipa and activedirectory auth providers

### LINKED ISSUE #429: Manual LDAP service account credential entry is still required when using rancher2_auth_config_activedirectory resource
When using the rancher2_auth_config_activedirectory resource in the Rancher2 terraform provider, after the initial deployment it is still required to log in with the local admin and re-input the LDAP service account credentials before it works.  It would be great if this step was no longer needed.

Screenshot below is the rancher2_auth_config_activedirectory configuration:

image

gz#11487

rawmind0:
Closes #429


paraglade (approved):
lgtm

Issue #496, opened 2020-10-31
Linode node_template Label issue 


lajos88:
Hey,

I recently run into a problem with the Linode node_template.
The "label" property is a string property and it has a `default: false` value here:
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_node_template_linode.go#L69

This means if I don't set anything to the "label" property in my terraform file it will be translated to "0" during the plan apply. This means that the resource can't be created (Linode error - Label must be between 3 and 255 chars).
This is a problem if I don't include the property at all or if I include the property with "" value too.

I found a temporary solution - I can add "label" = " " to my terraform file, and it is working, but I think the best would be to remove the `default: false` from the provider's code.

What do you think?

Thanks in advance,
Lali

lajos88:
PR: https://github.com/rancher/terraform-provider-rancher2/pull/497

rawmind0:
Hi @lajos88 , you are right, this default value was wrong. Merged your PR. 

Please reopen issue if needed

PR #497, opened 2020-10-31
Removed unnecessary default: true from the label property

lajos88:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/496


rawmind0 (approved):
LGTM

PR #509, opened 2020-11-09
fix rancher url property doc for auth resources

mattmichal:
Updates documentation for various auth resources to correct a typo in the `rancher_api_host` property. Per [RFC 1738](https://tools.ietf.org/html/rfc1738#section-2.1) the "https" portion of a URL is the `scheme`, not the `schema` as described in the docs.


rawmind0 (approved):
LGTM

PR #558, opened 2021-01-19
Avoid misleading diffs for password questions/answers

mouellet:
The Rancher API filters out attributs of type `password` (e.g. `norman:"type=password"`), which is a good thing.

However, when deploying a cluster using the `rancher2_cluster` resource using `cluster_template_answers` to override `cluster_template_questions` of a RKE Template, those `password` type attribute are always shown as diff in the plan even if they haven't changed.

This PR fixes this by preserving the values of `cluster_template_answers` (available in the state) for `cluster_template_questions` of type `password`. It also add the `password` type for `cluster_template_questions` as it is valid according to the API schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.12.0 (Unreleased)
+
+ENHANCEMENTS:
+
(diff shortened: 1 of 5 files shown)


rawmind0 (approved):
Thanks for the PR @mouellet 

LGTM

PR #585, opened 2021-02-17
Added rancher2_node_template.node_taints argument to resource and datasource

### LINKED ISSUE #584: Support node_taints on rancher2_node_template
When I create a node template through the UI, I can add node taints to the template, which will then be applied to any node started from that template. A template made that way will show this in the API:

```json
nodeTaints": [
    {
        "effect": "PreferNoSchedule",
        "key": "myorg.net/specialnode",
        "type": "/v3/schemas/taint",
        "value": "true"
    }
],
```

With the Terraform provider, it seems there is no way to add this, if I'm reading the docs correctly. Would it be possible to add support for that? Right now I need to add node labels in the Template and node taints in the Node Pool. I'd like to add them both through the template.

Thank you!

rawmind0:
Closes https://github.com/rancher/terraform-provider-rancher2/issues/584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 FEATURES:
 
+* **New Argument:** `rancher2_node_template.node_taints` - (Optional) Node taints. For Rancher v2.3.3 or above (List)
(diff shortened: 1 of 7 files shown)


paraglade (approved):
approved

PR #589, opened 2021-02-19
Update rancher2_cluster docs, mentioning kubeVersion requirement on monitoring V1 version 0.2.0 or above

rawmind0:
Issue https://github.com/rancher/rancher/issues/31364

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -60,6 +60,8 @@ resource "rancher2_cluster" "foo-custom" {
 
 ### Creating Rancher v2 RKE cluster enabling and customizing monitoring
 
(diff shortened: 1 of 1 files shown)


deniseschannon (approved):

RiccardoBoettcher:
The added comment doesn't help for a terraform user. What does "until cluster is fully deployed" mean. What is the condition to state "a cluster is fully deployed"?
I'm trying to import a EKS cluster which is already fully deployed in AWS and while trying to import the cluster into rancher it fails with errors related to "kubeversion".
Unfortunatly the hint doesn't give any guidedance how to fix this problem at the users side, or the hint that this problems needs a fix in
* rancher
* terraforms rancher2 provider
* terraforms helm provider
* ???
Please bring more clarification here.

PR #592, opened 2021-02-22
Add `nodes` attribute reference to `rancher2_cluster_sync` resource

mouellet:
This adds the cluster node list from `/v3/nodes` as attribute reference to the `rancher2_cluster_sync`.

I used the `rancher2_cluster_sync` resource as it is said to _compute attributes with useful cluster related data_ and the nodes are expected to be up & running when the resource is created.

Use case: using `terraform output -json` as input for post-deployment smoke test.


rawmind0 (approved):
@mouellet , thanks for this PR
LGTM

frouzbeh:
@mouellet 
 Hi, does this pr mean cluster_sync makes sure all nodes are ready and active? I’m trying to make sure the nodes are active before executing a script on them but I haven’t been able to.

Issue #591, opened 2021-02-22
rancher2_app_v2 resource doesn't respect system default registry when updating app


dbason:
On a Rancher cluster configured with a private default registry the rancher/shell:v0.1.5 image is trying to pull from public docker hub when the app is upgraded using terraform.  When upgraded using the UI the following values are passed to the API:
```
"values": {
    "global": {
        "systemDefaultRegistry": "registry.address:registry_port",
        "cattle": {
            "clusterId": "cluster_id",
            "clusterName": "cluster_name",
            "systemDefaultRegistry": "registry.address:registry_port"
        }
    }
}
```
The systemDefaultRegistry entries should be included in the global values by the provider so the default registry is respected.

gz#15233

rawmind0:
Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/605 to fix the issue

rawmind0:
Fix included at new Rancher2 tfp release v1.13.0

PR #605, opened 2021-03-12
Fix rancher2_app_v2 to respect Rancher system default registry

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## 1.13.0 (Unreleased)
+
+FEATURES:
+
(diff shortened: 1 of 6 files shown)


bashofmann (approved):
Works great for me:

* I deployed a Rancher with a private default registry
* I deployed an app (cis benchmarks) with rancher2_app_v2
* With the terraform 1.12.0 provider version, the default system registry is not added to the images
* With the change in this PR, the default system registry is added to the images

Issue #601, opened 2021-03-08
Creation of EKS cluster with rancher 2.5.6 does not work.


wojtasool:
After rancher server upgrade from 2.5.5 to 2.5.6 creation of EKS clusters does not work anymore.
Tried plugin rancher2 with version 1.12.0 but no sucess.
It complaints about launchTemplate beeing nil and also about ResourceTags being nil for node group.
Example:
"Field [resourceTags] cannot be nil for nodegroup [harbor_node_group] in non-nil cluster [c-8pl69]"

Such fields can not be specified with there versions of terraform module.
I'm using eks_config_v2 configuration parameter.

gz#15934

wojtasool:
| Field [launchTemplate.ID] cannot be nil for nodegroup [test-node_group] in non-nil cluster [c-5mj79]
-- | --

carlcauchi:
**[ BLOCKING ISSUE ]**

I have managed to replicate the same issue using Rancher v2.5.7 using Rancher2 provider v1.12.0:
When creating an EKS Cluster, its errors out on:

**`Field [launchTemplate.ID] cannot be nil for nodegroup [xxxxNodeGroupNamexxxx] in non-nil cluster [c-b69l4]`**

@rawmind0

rawmind0:
Fix included at new Rancher2 tfp release v1.13.0

PR #608, opened 2021-03-18
Fix rancher2_cluster.eks_config_v2 to deploy properly EKS clusters

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/601


carlcauchi (approved):

carlcauchi:
I have managed to test out the PR manually as its not yet merged and noted two issues which one of them might not be related to this change per se:
1. When a cluster is destroyed via terraform, the Node Templates created automatically (by each node group), they won't be deleted automatically and recreating the same cluster will fail as it errors out on the NodeTemplate saying it already exist
2. More generic issue - when you run a terraform destroy which would include an EKS Cluster, it will just delete the EKS Cluster from Rancher but it will NOT update/delete the EKS cluster from AWS side - so it remains Active. What it does delete though, is the NodeGroups but not the Cluster itself - so any empty cluster is left Active

- The mentioned issues are related to Rancher rather then the terraform provider. This change can be approved.

paraglade (approved):
IGTM

Issue #590, opened 2021-02-19
Method for confirming clusterrepos are created and downloaded in Rancher v2.5 clusters


axeal:
Attempting to deploy a chart from the `rancher-charts` or `rancher-partner-charts` repositories in a Rancher v2.5 provisioned cluster, via the `rancher2_app_v2` resource, immediately after cluster provisioning will fail with a message of the following format, if the clusterrepos has not yet been created.

```
Error: Bad response statusCode [404]. Status [404 Not Found]. Body: [code=NotFound, message=clusterrepos.catalog.cattle.io "rancher-charts" not found] from [https://178.128.253.220/k8s/clusters/c-wmhlg/v1/catalog.cattle.io.clusterrepos/rancher-charts] 
```

User's request is to have a method of verifying that the clusterrepo has been created and downloaded, so that the creation of the `rancher2_app_v2` is only attempted when this is true. One workaround is to use the `state_confirm` argument of the `rancher2_cluster_sync` resource, but this is arbitrary and cannot guarantee the repository is available yet.

gz#15112

rawmind0:
Added retry logic to get V2 catalogs and apps at PR https://github.com/rancher/terraform-provider-rancher2/pull/569 . This should address the issue.

rawmind0:
Fix included on new provider release v1.12.0

Please, reopen issue if needed

julienym:
This is still a problem with 1.12.0, but in our case it's not happening all the time.

Error: Bad response statusCode [404]. Status [404 Not Found]. Body: [code=NotFound, message=clusterrepos.catalog.cattle.io "rancher-charts" not found] from [https://XXXXXXXXXXXXXXX/k8s/clusters/c-57c5v/v1/catalog.cattle.io.clusterrepos/rancher-charts]

Why not use a data source and verify a refreshed attribute?

julienym:
@rawmind0 could you reopen this please?

rawmind0:
@julienym there is no `refreshed` attribute for clusterrepos, just a [`Downloaded` condition](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_catalog_v2.go#L147). This condition is checked when you add a new clusterrepo, waiting until the clusterrepo is downloaded https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_catalog_v2.go#L45

[...]

PR #609, opened 2021-03-19
rancher2_catalog_v2, wait until donwloaded status

rawmind0:
This PR constains
- Fix rancher2_catalog_v2 to wait until donwloaded status
- Updated `rancher2_cluster_sync` to allow wait until all catalogs are downloaded and active

Issue https://github.com/rancher/terraform-provider-rancher2/issues/590


bashofmann (approved):
Looks good and tested that it also works: When I create a catalog and deploy a new app from that catalog, terraform only continues after the download worked and the installation runs through successfully.

Issue #603, opened 2021-03-09
Import rke2 clusters


ron1:
Provider should deliver the same level of support to rke2 clusters as it does to k3s clusters. At a minimum, the importation of rke2 clusters should be supported.

rawmind0:
PR https://github.com/rancher/terraform-provider-rancher2/pull/626 has been tested and merged. The feaure will be availbale on next tfp release.

Please, reopen issue if needed

ron1:
@rawmind0 Since this is a significant, customer-facing capability, would it be possible to cut a new tfp release soon that includes this feature?

PR #626, opened 2021-04-13
Added rancher2_cluster support to rke2 config

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/603


jmatsushita:
It seems that this PR only supports upgrade strategy and version fields for rke2. Are there plans to support all the rke2 [configuration options](https://docs.rke2.io/install/install_options/install_options/#configuration-file)?

superseb:
@jmatsushita This is because the current support for managing RKE2 clusters in Rancher is limited to upgrade, the configuration of RKE2 clusters still needs to be done in RKE2. As soon as this support lands, we can consider adding it to the terraform provider.

superseb (approved):

Issue #627, opened 2021-04-13
rancher2_app_v2 fails waiting on system-default-registry setting propagation


mouellet:
I'm using v1.13.0 to deploy an rke cluster with `rancher2_cluster`, making sure the cluster is active with the `rancher2_cluster_sync` resource and install gatekeeper using the `rancher2_app_v2` resource but I'm getting this error:

> Error: Bad response statusCode [404]. Status [404 Not Found]. Body: [code=NotFound, message=settings.management.cattle.io "system-default-registry" not found] from [https://<host_name>/k8s/clusters/<cluster_id>/v1/management.cattle.io.settings/system-default-registry]

It looks like the settings sync made by the `cattle-cluster-agent` workload can take up to 2 minutes. Is this expected?

Should the `rancher2_cluster_sync` resource have an option to wait on settings?

rawmind0:
Issue reproduced. Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/632 with a fix

rawmind0:
Attached tfp custom binaries with patch applied to be tested, https://github.com/rancher/terraform-provider-rancher2/issues/648#issuecomment-832019144

rawmind0:
The fix is available at tfp v1.14.0

armsnyder:
@rawmind0 Hi Raul. Thanks for all the updates to this provider. ~Is using v1.14.0 supposed to resolve this, or are there some required attributes we need to set on the `rancher2_cluster_sync` resource?~

Also, the related upstream bug: https://github.com/rancher/rancher/issues/32642

~By the way my team is using imported clusters and still seeing this issue using v1.14.0.~

EDIT: Sorry, in our case it appears the original 404 did actually go away when we upgraded to v1.14.0. We specified the wrong catalog name and it was a different legitimate 404. Please disregard. 😄  The issue #662 is still valid however.

rawmind0:
Hi @armsnyder , 
* about issue https://github.com/rancher/rancher/issues/32642 , global settings were synced on all k8s clusters API, but not anymore, so global settings as `system-default-registry`, has to be got from `local` cluster.
* about the issue #662 , retries seems to work fine, but not enough with default conf, agreed that it should be configurable. 

Working on a fix for both issues.

PR #632, opened 2021-04-20
rancher2 cluster fixes

rawmind0:
This PR contains:
* Fix `rancher2_cluster` resource, fixing provider crash if `cluster_monitoring_input` argument is deleted
* Improved `rancher2_cluster_sync` with new cluster state check method and new option to wait until alerting is enabled

Issues https://github.com/rancher/terraform-provider-rancher2/issues/627 and https://github.com/rancher/terraform-provider-rancher2/issues/633


rancher2/config.go
```diff
@@ -1111,16 +1139,57 @@ func (c *Config) UpdateClusterByID(cluster *managementClient.Cluster, update map
...
+				return true, obj, nil
+			}
+			return false, obj, nil
+		}
+	}
+
+	return false, obj, nil
+}
+
+func (c *Config) isClusterMonitoringEnabled(id string) (bool, *managementClient.Cluster, error) {
```
superseb: Not really sure about this naming, this check if cluster monitoring is Active right?

rancher2/config.go
```diff
@@ -1111,16 +1139,57 @@ func (c *Config) UpdateClusterByID(cluster *managementClient.Cluster, update map
...
+				return true, obj, nil
+			}
+			return false, obj, nil
+		}
+	}
+
+	return false, obj, nil
+}
+
+func (c *Config) isClusterAlertingEnabled(id string) (bool, *managementClient.Cluster, error) {
```
superseb: Same as with `isClusterMonitoringEnabled`

rawmind0: This function is checking if cluster condition, `clusterAlertingEnabledCondition   = "AlertingEnabled"` is true

rancher2/config.go
rawmind0: yeah, this is a question how cluster state is checked. This function is checking if cluster condition, `clusterMonitoringEnabledCondition = "MonitoringEnabled"` is true

superseb: Can we change the function name to include condition so its clear it is checking the condition being enabled and not the config cluster monitoring? Or does that not make sense?

rawmind0: Updated function name to `isClusterMonitoringEnabledCondition`

rancher2/config.go
rawmind0: Updated function name to `isClusterAlertingEnabledCondition`

superseb (approved):

Issue #637, opened 2021-04-22
rancher2_catalog_v2 shouldn't set git_branch when adding an http(s) repo


MG2R:
Right now when you add a rancher2_catalog_v2 with an http address like below, the provider will also set the git_branch. This somehow causes the cluster-agent to continuously start refreshing the resource, making it consume an average of about 700m cpu instead of <100 during normal operation. I don't see this weirdness when adding an http catalog through the Rancher UI (Apps and Marketplace).

```Terraform
resource "rancher2_catalog_v2" "dynamic" {
  name       = nexus
  url        = "https://<fqdn>/nexus/repository/helm-hosted/"
  cluster_id = data.rancher2_cluster.cluster.id
}                                                                          
```

As you can see `git_branch` gets set to master:
```
  # module.rancher_cluster_configuration.rancher2_catalog_v2.dynamic will be created
  + resource "rancher2_catalog_v2" "dynamic" {
      + annotations      = (known after apply)
      + cluster_id       = <cluster_id<>
      + enabled          = true
      + git_branch       = "master"
      + id               = (known after apply)
      + insecure         = false
      + labels           = (known after apply)
      + name             = "nexus"
      + resource_version = (known after apply)
      + url              = "https://<fqdn>/nexus/repository/helm-hosted/"
    }
```

This creates the following object in Rancher:
```YAML
actions: null
apiVersion: catalog.cattle.io/v1
kind: ClusterRepo
links: null
metadata:
  creationTimestamp: "2021-04-22T14:52:03Z"
  generation: 1
  managedFields:
  - apiVersion: catalog.cattle.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:enabled: {}
        f:gitBranch: {}
        f:url: {}
      f:status:
        .: {}
        f:conditions: {}
        f:downloadTime: {}
        f:indexConfigMapName: {}
        f:indexConfigMapNamespace: {}
        f:indexConfigMapResourceVersion: {}
        f:observedGeneration: {}
        f:url: {}
    manager: agent
    operation: Update
    time: "2021-04-22T14:52:09Z"
  name: nexus
  resourceVersion: "4077212"
  selfLink: /apis/catalog.cattle.io/v1/clusterrepos/nexus
  uid: <UUID>
spec:
  enabled: true
  gitBranch: master
  url: https://<fqdn>/nexus/repository/helm-hosted/
status:
  conditions:
[...]

PR #638, opened 2021-04-22
Fix rancher2_catalog_v2 resource, just setting default git_branch value if git_repo is specified

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/637


rancher2/structure_catalog_v2.go
```diff
@@ -74,11 +74,12 @@ func expandCatalogV2(in *schema.ResourceData) *ClusterRepo {
...
 	}
-	if v, ok := in.Get("git_branch").(string); ok {
-		obj.Spec.GitBranch = v
-	}
-	if v, ok := in.Get("git_repo").(string); ok {
+	if v, ok := in.Get("git_repo").(string); ok && len(v) > 0 {
 		obj.Spec.GitRepo = v
+		obj.Spec.GitBranch = catalogV2DefaultGitBranch
+	}
+	if v, ok := in.Get("git_branch").(string); ok && len(v) > 0 {
```
superseb, line 81: Should this move up, nested in `git_repo` check or is there a need to be able to specify `git_branch` outside of using `git_repo`? I know this is probably caught with the added conflicts but code-wise I can't see the logic why it is not nested.

rawmind0: The problem here was with the default value assigned by tfp. As `git_branch` had a default value, it was adding it to all catalogs v2, `git_repo` and `url` types. The logic here is that `git_branch` got a default value, just if `git_repo` has been specified. If `git_branch` has been also specified in the tf file, then the defined value is overwriting the default value.

bashofmann (approved):
Functionality looks good:
* A repo with a http url does not contain a gitBranch field
* A repo with only a gitRepo sets gitBranch to master as default
* You can change the default value of it with git_branch

superseb (approved):

PR #640, opened 2021-04-23
Fix eks_config_v2 issues when import eks cluster

rawmind0:
Issue https://github.com/rancher/rancher/issues/30480

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ BUG FIXES:
 
 * Fix `rancher2_node_pool` resource, adding `forcenew` property to not updatable arguments
 * Fix `rancher2_catalog_v2` resource, just setting default `git_branch` value if `git_repo` is specified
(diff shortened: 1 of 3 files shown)


superseb (approved):

bashofmann:
What happens when you create a new eks cluster from Rancher? Would this be a private or public cluster without having a default value? Or would you always need to explicitly set one of the flags?
If yes, this should be added to the example in the docs (https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#creating-eks-cluster-from-rancher-v2-using-eks_config_v2-for-rancher-v25x-or-above).

rawmind0:
> What happens when you create a new eks cluster from Rancher? Would this be a private or public cluster without having a default value? Or would you always need to explicitly set one of the flags?
> If yes, this should be added to the example in the docs

As both arguments are computed, they will get the eks-operator default values. Anyway, good point @bashofmann, it's a good practice to add them. Added both to the docs examples.

PR #653, opened 2021-05-06
Add cidr opts for pods and services to oke_config

jrosinsk:
This PR adds the ability to set the kubernetes pod and service cidr when creating an oke cluster.

```
resource "rancher2_cluster" "dev" {
     oke_config {
          ...
          pod_cidr = "10.208.0.0/16"
          service_cidr =  "10.74.0.0/16" 
          ...
```    
  
Here are the 2 corresponding merges to enable this:  
https://github.com/rancher-plugins/kontainer-engine-driver-oke/pull/37  
https://github.com/rancher-plugins/kontainer-engine-driver-oke/pull/36


rawmind0:
Is `rancher-plugins/kontainer-engine-driver-oke` already released with a version including these fixes??

jrosinsk:
It's on master, but hasn't been released yet.  That should happen today if all goes according to plan.

rawmind0:
@jrosinsk , please confirm us when `rancher-plugins/kontainer-engine-driver-oke` including these fixes, is released.

jrosinsk:
@rawmind0 , rancher-plugins/kontainer-engine-driver-oke with these fixes has been released.
https://github.com/rancher-plugins/kontainer-engine-driver-oke/releases/tag/v1.6.0

rawmind0:
Thanks for confirming this @jrosinsk

rancher2/schema_cluster_oke_config.go
```diff
@@ -201,6 +203,18 @@ func clusterOKEConfigFields() map[string]*schema.Schema {
...
+			Optional:    true,
+			Default:     "10.242.0.0/16",
+			Description: "Optional specify the pod CIDR",
+		},
+		"service_cidr": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Default:     "10.92.0.0/16",
+			Description: "Optional specify the service CIDR",
+		},
```
rawmind0, line 215: Both arguments are set as `optional` but setting a default value, both values are required?? I mean, can't be these parameters let empty `""` at some oke deployment scenario??  Just checking if would be better to set both arguments also as computed, instead of setting default values

jrosinsk: This is definitely optional.  The value is not required at all, and if you don't provide the flag, those values listed as default are what you will get.  
The terraform docs suggest a slightly different use case for Computed

```
Computed is often used to represent values that are not user configurable or can not be known at time of terraform plan or apply, such as date of creation or a service specific UUID.
...
```

rawmind0: TF fields can have both properties, `Optional` (user configurable) and `Computed` (not user configurable but getting data from provider source). 

Anyway, the question is not around my preference to be Computed fields or not, moreover it would be howto facilitate use case. If the fields are optional, i assume that the fields can be empty. So if default value has been defined, for a use case where the user wants to let field empty it should defined `pod_cidr=""`. Other option to set the fields as Computed, may be not set default value or set it as `Default: ""` . I'm just asking as i don't have any experience using oke_config. If this config is fine for you, LGTM.. :)

jrosinsk: What this PR is trying to achieve is to add the ability to define those variables for the user who might want alternate CIDR's for their cluster pods and services.  If the user isn't worried about it, they would not even define pod_cidr or service_cidr in the oke_config which would result in them getting the OKE defaults for cluster creation.  Here is a snippet of the cluster being created with the option in question.  We want the user to not even define those if they don't need to and have the cluster create work, which is why it's Optional.

Screen Shot 2021-05-12 at 11 50 27 AM

There is a similar example for a similar variable in the 'schema_cluster_aks_config'.

jrosinsk: @rawmind0, I  get your point now, I removed the Default settings for these parameters.

rawmind0 (approved):
LGTM

Issue #655, opened 2021-05-10
rke_config in rancher2_cluster resource for custom provider seems useless


rule88:
It is hard to figure out that apparently rancher2_cluster does not provision hosts when using the "custom" provider, despite configurations imply it will.

Documentation is also not clear about this. Especially arguments like `bastion_host` and `ssh_agent_auth` do suggest provisioning nodes with RKE (as what RKE is built for) is possible.

rule88:
To elaborate, we would like to provision new `child` clusters on top of `custom` machines. We assumed it would be possible to use the `rke_config` attribute in the `rancher2_cluster` resource for this as the `parent` cluster is also provisioned with rke (using tf module rancher/rke).

Is there any automation in this case possible other than fetching a token from the `rancher2_cluster` resource and fiddling around with executors?

rawmind0:
Custom `rancher2_cluster` is not deploying any node nor imply it, same behaviour as got from Rancher UI. [Rancher doc custom nodes](https://rancher.com/docs/rancher/v2.0-v2.4/en/cluster-provisioning/rke-clusters/custom-nodes/) is also explaining it. Custom cluster is assuming nodes are already deployed, so once the cluster is created, [`node_command`](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#node_command) is generated and it's needed to be executed at nodes, with proper k8s roles.

If you want to deploy a cluster (including the infrastructure deployment), you should use `rancher2_node_template` and `rancher2_node_pools`. You can find an example in the [Rancher2 tf provider docs](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#creating-rancher-v2-rke-cluster-assigning-a-node-pool-overlapped-planes)

rule88:
It does imply deploying k8s using RKE. The confusion is RKE itself, as with RKE you normally deploy RKE on existing machines (machines are ready and only awaiting the initial docker command).

Using the RKE TF module (also provided by rancher), the same attributes are accepted and this module **does** deploy k8s to the provided list of hosts.

rawmind0:
Rancher2 and RKE works different at this point, so the tf providers are working different. RKE is connecting by ssh to nodes to run the needed components, Rancher2 instead is just providing the command and you need to execute it in yous nodes and they will connect to Rancher.

rule88:
[...]

PR #659, opened 2021-05-11
add note for rke_config 

rule88:
Puts down a small note to prevent confusion on the expected behavior of `rke_config` parameter.

should prevent issues similar to: https://github.com/rancher/terraform-provider-rancher2/issues/655

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -560,6 +560,8 @@ The following attributes are exported:
 
 ### `rke_config`
 
+**Note** `rke_config` works the same as within Rancher GUI; it will not _provision_ hosts when using no `cloud_provider` or when `cloud_provider` is set to custom. It is expected that nodes are registered by having the `node_command` run on each node. Running the `node_command` is outside the scope of this provider. 
+
 #### Arguments
 
 * `addon_job_timeout` - (Optional/Computed) Duration in seconds of addon job (int)


docs/resources/cluster.md
```diff
@@ -560,6 +560,8 @@ The following attributes are exported:
 
 ### `rke_config`
 
+**Note** `rke_config` works the same as within Rancher GUI, thus it will not _provision_ hosts when using `cloud_provider` custom. Provisioning clusters could be done by using the [Terraform provider RKE](https://github.com/rancher/terraform-provider-rke).
+
```
rawmind0, line 564: `...thus it will not _provision_ hosts when using cloud_provider custom...` may be is more clear to say that it's not deploying nodes if no cloud provider is configured?? 

`Provisioning clusters could be done by using the [Terraform provider RKE](https://github.com/rancher/terraform-provider-rke).` I thinks this is confusing as cluster can be provisioned from this provider, and you are not getting same result using RKE provider. I mean, using RKE and importing the cluster to Rancher would be an option, but not doing same task, you are not generating a custom cluster. I think a clarification saying that on custom clusters nodes are not deployed/configured automatically and it's expected to be already up and ready to execute the node_command, would be more proper.

rawmind0 (commented):
Thanks for the PR @rule88. Please, take a look to comments.

docs/resources/cluster.md
rule88: @rawmind0 - do you think the updated note is better? I focused more on what is done and what is expected to be done from the provider point of view.

rawmind0: @rule88 - much better to me, thanks!

rawmind0 (approved):
LGTM

Issue #674, opened 2021-05-27
Resource "rancher2_auth_config_keycloak" not supporting optional "Entity ID Field"


rule88:
An alternative `Client ID` in Keycloak than the default described in the [Rancher documentation](https://rancher.com/docs/rancher/v2.x/en/admin-settings/authentication/keycloak/) is not possible as the `rancher2_auth_config_keycloak` resource does not provide a way to override the default.

rule88:
PR: https://github.com/rancher/terraform-provider-rancher2/pull/675

rule88:
I have to admit that this PR is not tested yet.

rawmind0:
The PR #675 doesn't pass CI pipeline, https://github.com/rancher/terraform-provider-rancher2/pull/675#issuecomment-851399232

hnc-yusop:
Is there any progress for this issue???

rule88:
@hnc-yusop 

code is updated and merged in master, now it is waiting for next release which will contain this feature.

rawmind0:
PR #675 id already merged. Feature will be available at next tf provider release. 

Please, reopen issue if needed.

PR #675, opened 2021-05-27
add optional entity_id argument

rule88:
In favor of https://github.com/rancher/terraform-provider-rancher2/issues/674

Adds optional argument `entity_id`.

### DIFF
--- a/docs/resources/auth_config_keycloak.md
+++ b/docs/resources/auth_config_keycloak.md
@@ -41,6 +41,7 @@ The following arguments are supported:
 * `enabled` - (Optional) Enable auth config provider. Default `true` (bool)
 * `annotations` - (Optional/Computed) Annotations of the resource (map)
 * `labels` - (Optional/Computed) Labels of the resource (map)
+* `entity_id` - (Optional/Computed) KeyCloak Client ID field (string)
 
 ## Attributes Reference
 
--- a/rancher2/schema_auth_config_keycloak.go
+++ b/rancher2/schema_auth_config_keycloak.go
@@ -48,6 +48,11 @@ func authConfigKeyCloakFields() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Required: true,
 		},
+		"entity_id": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+		},
 	}
 
 	for k, v := range authConfigFields() {
--- a/rancher2/structure_auth_config_keycloak.go
+++ b/rancher2/structure_auth_config_keycloak.go
@@ -38,6 +38,7 @@ func flattenAuthConfigKeyCloak(d *schema.ResourceData, in *managementClient.KeyC
 	d.Set("sp_cert", in.SpCert)
 	d.Set("uid_field", in.UIDField)
 	d.Set("user_name_field", in.UserNameField)
+	d.Set("entity_id", in.EntityID)
 
 	return nil
 }
@@ -107,6 +108,10 @@ func expandAuthConfigKeyCloak(in *schema.ResourceData) (*managementClient.KeyClo
 
 	if v, ok := in.Get("user_name_field").(string); ok && len(v) > 0 {
 		obj.UserNameField = v
+	}
+
+	if v, ok := in.Get("entity_id").(string); ok && len(v) > 0 {
+		obj.EntityID = v
 	}
 
 	return obj, nil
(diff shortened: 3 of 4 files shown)


rawmind0:
Error at CI pipeline
```
...
== Running testing...
--
325 | go test "./rancher2" \|\| exit 1
326 | --- FAIL: TestProvider (0.03s)
327 | 00_provider_test.go:42: err: 1 error occurred:
328 | * resource rancher2_auth_config_keycloak: entity_id: One of optional, required, or computed must be set
...
```

@rule88 at least `make test`, preferably `make testacc`, should be passed before submit PR.

rule88:
Allright, locally test now succeeds, but not sure if the approach is ok. As the argument is not required, I removed the test (as this is true as well for most non-required arguments).

rawmind0:
Where is the new `entity_id` argument defined at `schema_auth_config_keycloak.go` ?? The argument should be defined at the resource schema and it should be added to the structure tests (optional fields are already tested)

rule88:
when having `entity_id` defined in `schema_auth_config_keycloak.go`:

```
        "entity_id": {
            Type:     schema.TypeString,
            Required: false,
        },
```

I receive this message:

```
--- FAIL: TestProvider (0.02s)
    00_provider_test.go:42: err: 1 error occurred:
        	* resource rancher2_auth_config_keycloak: entity_id: One of optional, required, or computed must be set
        
FAIL
FAIL	github.com/terraform-providers/terraform-provider-rancher2/rancher2	5.172s
FAIL
```

Should I explicitly somehow set it to optional ?

rule88:
I think I get it now, (see latest commit) :-).

Sorry for the hazzle, not (yet) a real GO programmer here

docs/resources/auth_config_keycloak.md
```diff
@@ -41,6 +41,7 @@ The following arguments are supported:
 * `enabled` - (Optional) Enable auth config provider. Default `true` (bool)
 * `annotations` - (Optional/Computed) Annotations of the resource (map)
 * `labels` - (Optional/Computed) Labels of the resource (map)
+* `entity_id` - (Optional) KeyCloak Client ID field, defaults to `https://yourRancherHostURL/v1-saml/keycloak/saml/metadata` (string)
```
rawmind0: Where is coming the default value for `entity_id` value??

rancher2/schema_auth_config_keycloak.go
```diff
@@ -48,6 +48,10 @@ func authConfigKeyCloakFields() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Required: true,
 		},
+		"entity_id": {
+			Type:     schema.TypeString,
+			Optional: true,
```
rawmind0, line 53: If `entity_id` has a default value, one of 2 options should be implemented to avoid false diff:
* Add default value on the argument definition. `Default: "<DEFAULT_VALUE>",`
* Define the argument also as computed. `Computed: true,`

rawmind0 (changes requested):
No problem @rule88 , thanks for the PR update.

Please, take a look to changes requested

docs/resources/auth_config_keycloak.md
rule88: De "default" value is from Rancher itself. Now I am reconsidering, does that mean the value actually is `Optional/Computed` (removing the 'defaults to `https://yourRancherHostURL/v1-saml/keycloak/saml/metadata`' part of doc)?

rancher2/schema_auth_config_keycloak.go
rule88: If not provided, the Rancher API will default to what is [documented](https://rancher.com/docs/rancher/v2.x/en/admin-settings/authentication/keycloak/), that implies to the user it is `Computed`, but it `Computed` by `Rancher`, not `Terraform` (this is probably where I am a bit confused).

Will the following be correct?

```	
Type:     schema.TypeString,			
Default: "https://yourRancherHostURL/v1-saml/keycloak/saml/metadata"
Optional: true,
Computed: true,
```

docs/resources/auth_config_keycloak.md
rawmind0: This may be one option. Define the argument as `Optional/Computed` without default value

rancher2/schema_auth_config_keycloak.go
rule88: (answering my own question) I guess not, the "Default" is generated by `Rancher` with it's known `yourRancherHostURL`, not sure now what to provide as default in Terraform case.

rawmind0: Indeed, a `computed` tf argument use to be computed externally, in that case at Rancher side. If the field is not computed, the default value should be specified at definition to avoid false diff on resource comparisson

rule88:
TIL: 

Terraform `computed` != `computed` by Terraform

Thanks @rawmind0 ;-) !

rawmind0 (approved):
LGTM

Issue #690, opened 2021-06-21
Support drainBeforeDelete field for nodepools


rodcloutier:
The Rancher 2.5 `drainBeforeDelete` is not currently supported by the provider. 
We would like to have the support for this field as it is required to be set to `true` in an autoscaler user case. 

```
    "annotations": {
        "lifecycle.cattle.io/create.nodepool-provisioner": "true",
        "nodepool.cattle.io/reconcile": ""
    },
    "baseType": "nodePool",
    "clusterId": "c-n2p22",
    "controlPlane": false,
    "created": "2021-06-17T16:31:08Z",
    "createdTS": 1623947468000,
    "creatorId": "u-j7gfdsy6zo",
    "deleteNotReadyAfterSecs": 0,
    "displayName": "",
    "drainBeforeDelete": false,    # <------------------ Need to be able to change this
    "driver": "ubicloud",
    "etcd": false,
    "hostnamePrefix": "c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
    "id": "c-n2p22:c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
    "labels": {
        "cattle.io/creator": "norman"
    },
    "links": {
        "nodes": "…/v3/nodes?nodePoolId=c-n2p22%3Ac-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
        "remove": "…/v3/nodePools/c-n2p22:c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
        "self": "…/v3/nodePools/c-n2p22:c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
        "update": "…/v3/nodePools/c-n2p22:c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp"
    },
    "name": "c-398ji7x5fnmks48-np-398ji7x5fnmks48-o1vi74crjvsvp",
    "namespaceId": null,
    "nodeAnnotations": null,
    "nodeLabels": null,
    "nodeTemplateId": "cattle-global-nt:nt-t4w8d",
    "quantity": 3,
    "state": "active",
    "status": {
        "conditions": [
            {
                "lastUpdateTime": "2021-06-17T16:31:28Z",
                "status": "True",
                "type": "Updated"
            }
        ],
[...]

Issue #620, opened 2021-04-07
Feature request: Add support for node pool node labels and annotations


rodcloutier:
The Rancher api allows to set the labels and annotations on nodes through the node pool resource. 
The `nodeTaints` is already supported but `nodeAnnotations` and `nodeLabels` are not. 

```
{
    "annotations": {
        "lifecycle.cattle.io/create.nodepool-provisioner": "true",
    },
    "baseType": "nodePool",
    "clusterId": "c-7lrvt",
    "controlPlane": true,
    "created": "2021-04-07T17:03:16Z",
    "createdTS": 1617814996000,
    "creatorId": "u-r4zs2fm23u",
    "deleteNotReadyAfterSecs": 0,
    "displayName": "",
    "driver": "internal",
    "etcd": false,
    "hostnamePrefix": "cluster-master",
    "id": "c-7lrvt:cluster-master",
    "labels": {
        "cattle.io/creator": "norman",
    },
    "links": {
        "nodes": "…/v3/nodes?nodePoolId=c-7lrvt%3Acluster-master",
        "remove": "…/v3/nodePools/c-7lrvt:cluster-master",
        "self": "…/v3/nodePools/c-7lrvt:cluster-master",
        "update": "…/v3/nodePools/c-7lrvt:cluster-master"
    },
    "name": "cluster-master",
    "namespaceId": null,
    "nodeAnnotations": null,
    "nodeLabels": null,
    "nodeTaints": [
    ],
    "nodeTemplateId": "cattle-global-nt:nt-4pr48",
    "quantity": 1,
    "state": "active",
    "status": {
        "conditions": [
            {
                "lastUpdateTime": "2021-04-07T17:03:20Z",
                "status": "True",
                "type": "Updated"
            }
        ],
        "type": "/v3/schemas/nodePoolStatus"
    },
    "transitioning": "no",
    "transitioningMessage": "",
    "type": "nodePool",
    "uuid": "2a344a9b-ec01-4799-a056-22a974c23b48",
    "worker": false
}
```

mitchellmaler:
This can be done using the NodeTemplate resources labels and annotations fields.

Unfortunately to add labels and annotations to NodePools (which are applied to the downstream nodes) would require a Rancher Server change allowing the capability first before it can be added to tf provider.

rodcloutier:
Thanks for the hint. I will use that instead. 

Just a question then, what are those fields purpose?

mitchellmaler:
[...]

PR #695, opened 2021-06-25
New rancher2_node_pool.drain_before_delete argument

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/690
Issue https://github.com/rancher/terraform-provider-rancher2/issues/620

### DIFF
--- a/docs/resources/node_pool.md
+++ b/docs/resources/node_pool.md
@@ -65,6 +65,7 @@ The following arguments are supported:
 * `hostname_prefix` - (Required) The prefix for created nodes of the Node Pool (string)
 * `node_template_id` - (Required) The Node Template ID to use for node creation (string)
 * `delete_not_ready_after_secs` - (Optional) Delete not ready node after secs. For Rancher v2.3.3 or above. Default `0` (int)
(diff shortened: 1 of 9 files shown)


docs/resources/node_pool.md
```diff
@@ -65,6 +65,7 @@ The following arguments are supported:
 * `hostname_prefix` - (Required) The prefix for created nodes of the Node Pool (string)
 * `node_template_id` - (Required) The Node Template ID to use for node creation (string)
 * `delete_not_ready_after_secs` - (Optional) Delete not ready node after secs. For Rancher v2.3.3 or above. Default `0` (int)
+* `drain_before_delete` - (Optional) Drain nodes before delete (bool)
```
paynejacob: what is the default behavior?

paynejacob (changes requested):

docs/resources/node_pool.md
rawmind0: All tf bool without explicit default value are treat as `false`. Do you prefer to be explicit here?

rawmind0: @paynejacob Updated docs being explicit with `drain_before_delete` default value

paynejacob (approved):

Issue #382, opened 2020-07-09
Optional rancher2_notifier config elements set outside Terraform config not removed on terraform apply


axeal:
## Type ##

Bug

## Steps to reproduce ##

1. Provision a Rancher v2.4.5 instance with single all role node custom cluster (I used https://github.com/superseb/tf-do-rancher2/)
2. Create a simple Terraform config `main.tf` per the below example:

```
provider "rancher2" {
  api_url    = "https://164.90.187.250"
  access_key = "token-l6tz2"
  secret_key = "fvvkz89c9dp6grrpn8xrjtgjw78gl42jgtkbxjch8854b5q6fttvtt"
  insecure   = true
}

resource "rancher2_notifier" "foo" {
  name          = "foo"
  cluster_id    = "c-94x2h"
  description   = "Terraform notifier acceptance test"
  send_resolved = "true"
  smtp_config {
    default_recipient = "alex@example.org"
    host              = "smtp.example.org"
    port              = 25
    sender            = "rancher@example.org"
    tls               = false
  }
}

resource "rancher2_notifier" "bar" {
  name          = "bar"
  cluster_id    = "c-94x2h"
  description   = "Terraform notifier acceptance test"
  send_resolved = "true"
  webhook_config {
    url = "example.org"
  }
}
```
3. Run `terraform init` and `terraform apply`
4. Via the UI add a username to the `foo` smtp configuration and a proxy to the bar webhook config
5. Run `terraform apply` and observe output that terraform will remove the username and proxy configs:

```
Terraform will perform the following actions:

  # rancher2_notifier.bar will be updated in-place
  ~ resource "rancher2_notifier" "bar" {
        annotations   = {}
        cluster_id    = "c-94x2h"
        description   = "Terraform notifier acceptance test"
        id            = "c-94x2h:n-kfdvp"
        labels        = {
            "cattle.io/creator" = "norman"
        }
        name          = "bar"
        send_resolved = true

      ~ webhook_config {
          - proxy_url = "http://proxy" -> null
            url       = "example.org"
        }
    }

  # rancher2_notifier.foo will be updated in-place
  ~ resource "rancher2_notifier" "foo" {
        annotations   = {}
        cluster_id    = "c-94x2h"
        description   = "Terraform notifier acceptance test"
        id            = "c-94x2h:n-qqc78"
        labels        = {
            "cattle.io/creator" = "norman"
        }
        name          = "foo"
        send_resolved = true

      ~ smtp_config {
[...]

PR #706, opened 2021-07-22
Updated rancher2_notifier resource to be replaced on update

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/382

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## 1.16.1 (Unreleased)
+
+FEATURES:
+
+
+
+ENHANCEMENTS:
+
+
+
+BUG FIXES:
+
+* Updated rancher2_notifier resource to be replaced on update
+
 ## 1.16.0 (July 15, 2021)
 
 FEATURES:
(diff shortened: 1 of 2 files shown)


nickgerace (approved):
~~The logic LGTM, but what if users want these features?~~
EDIT: this was not supposed to be an approval... thanks GitHub + TAB + ENTER

nickgerace (commented):
The logic LGTM, but what if users want these features?

rawmind0:
> The logic LGTM, but what if users want these features?

@nickgerace , what user features do you mean?? 

The root cause of the issue is that the rancher2_notifier is merged instead of being replaced at Rancher API. The merge is not working fine on some scenarios due to null or empty fields are omitted at json marshal/unmarshal, so once any of this field is set, it can't be nullified nor emptied using the API. The PR is just forcing to replace the resource instead of update it.

nickgerace:
> @nickgerace , what user features do you mean??

The providers themselves, but reading your message, it makes sense now.

> The root cause of the issue is that the rancher2_notifier is merged instead of being replaced at Rancher API. The merge is not working fine on some scenarios due to null or empty fields are omitted at json marshal/unmarshal, so once any of this field is set, it can't be nullified nor emptied using the API. The PR is just forcing to replace the resource instead of update it.

Ah gotcha. Thanks!

superseb (approved):

Issue #710, opened 2021-07-29
unclear documentation around `rancher2_etcd_backup`


wc-s:
It's not clear whether `rancher2_etcd_backup` just takes a one-off snapshot or whether it takes periodic snapshots.

It seems to correspond to the `v3/etcdbackups` API, which should mean that it only takes one-off snapshots. However, the terraform resource / API also allows one to specify `interval_hours`, which seems to suggest that it can take periodic snapshots.

I've tried creating a `rancher2_etcd_backup` object via terraform, and it seems the `interval_hours` is simply overwritten with the default.

If I am right in thinking that it only takes one-off snapshots, then my next question is - what's the point of the option to set the `manual` flag in the terraform resource. I'd have thought that, be definition, any snapshot I take via creating a `rancher2_etcd_backup` terraform resource should be counted as a `manual` snapshot, as opposed to the non-manual, automatic, periodic snapshots that rancher takes.

tl-weichung-shaw:
Just to clarify, when I mentioned
> I've tried creating a backup object via the API, and it seems the interval_hours is simply overwritten with the default.

I meant that I tried to create a `rancher2_etcd_backup` object via terraform

rawmind0:
Updated docs at PR https://github.com/rancher/terraform-provider-rancher2/pull/717, already merged. The new docs will be available at next tfp release, v1.17.0

Please, reopen issue if required

PR #717, opened 2021-08-10
Clarify rancher2_etcd_backup resource docs

axeal:
The rancher2_etcd_backup resource is used to trigger one-time snapshots, in accordance with the cluster backup configuration, i.e. if no S3 backup config is present in cluster backup config a local snapshot is created if S3 backup config is present in cluster backup config an S3 snapshot is created. The backup_config on the rancher2_etcd_backup cannot be set to create a different type of backup and should probably just be computed.


rawmind0 (approved):
Issue https://github.com/rancher/terraform-provider-rancher2/issues/710

LGTM

PR #731, opened 2021-08-23
add "own" to allowed list of verbs for policy rules

maxaudron:
Rancher uses this for example in the "Project Owner" role, and the API
supports setting it without any problem. We wanted to duplicate this
role with our own changes and couldn't do so properly because we could
not set the verb to "own".

### DIFF
--- a/docs/resources/global_role.md
+++ b/docs/resources/global_role.md
@@ -51,7 +51,7 @@ The following attributes are exported:
 * `non_resource_urls` - (Optional) Policy rule non resource urls (list)
 * `resource_names` - (Optional) Policy rule resource names (list)
 * `resources` - (Optional) Policy rule resources (list)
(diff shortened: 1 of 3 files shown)


rawmind0 (dismissed):
LGTM

rawmind0 (changes requested):
@maxaudron , thanks for the PR.

Could you please add the new allowed verb to the docs?? [global_role](https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/global_role.md#rules) and [role_template](https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/role_template.md#rules)

maxaudron:
@rawmind0 done, felt like I forgot something :slightly_smiling_face:

rawmind0 (approved):
Thanks @maxaudron , i had same feeling.. :)

LGTM

Issue #741, opened 2021-09-10
rancher2_project_role_template_binding recreation loop when using user_principal_id


gillarda:
## Versions

* Terraform 1.0.6
* Rancher provider 1.17.2
* Rancher server v2.4

## Description

I am trying to set permissions on projects for users originating for an OpenLDAP authentication source.
The inial creation of the `rancher2_project_role_template_binding` works fine 👍 

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # module.rancher-terraform-module.module.project["tfepsilon"].rancher2_project_role_template_binding.prtb_test will be created
  + resource "rancher2_project_role_template_binding" "prtb_test" {
      + annotations        = (known after apply)
      + group_principal_id = (known after apply)
      + id                 = (known after apply)
      + labels             = (known after apply)
      + name               = "prtb-test-people"
      + project_id         = "c-jrmjd:"p-qhacq"
      + role_template_id   = "project-member"
      + user_principal_id  = "openldap_user://cn=usertest,ou=org,ou=People,dc=domain,dc=fr"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

However,  upon subsequent `terraform plan` the resource is planned for replacement as the `rancher2_project_role_template_binding` references the `user_id` as well as the `user_principal_id` :
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # module.rancher-terraform-module.module.project["tfepsilon"].rancher2_project_role_template_binding.prtb_test must be replaced
-/+ resource "rancher2_project_role_template_binding" "prtb_test" {
      ~ annotations        = {
          - "auth.cattle.io/principal-display-name"                = "usertest"
          - "lifecycle.cattle.io/create.cluster-prtb-sync_c-jrmjd" = "true"
          - "lifecycle.cattle.io/create.mgmt-auth-prtb-controller" = "true"
        } -> (known after apply)
      + group_principal_id = (known after apply)
      ~ id                 = "p-qhacq:prtb-test-people" -> (known after apply)
      ~ labels             = {
          - "cattle.io/creator" = "norman"
        } -> (known after apply)
        name               = "prtb-test-people"
      - user_id            = "u-chpb4czuic" -> null # forces replacement
        # (3 unchanged attributes hidden)
    }

[...]

PR #743, opened 2021-09-15
Updated user_id and group_id arguments as computed at cluster and project role template binding

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/741


a-blender (approved):
Looks good! You could also make your git commits imperative by saying 'Update user_id and group_id, etc'

PR #750, opened 2021-09-22
Wait for permissions to be applied when creating namespace

evhan:
Hi there! We're hitting a problem where this provider reliably fails to create `rancher2_namespace` resources as a project-level user, seemingly due to a timing issue in the way roletemplate permissions are applied in the target cluster. I'll do my best to explain...

In our permissions model, we have users who can create projects, and then create namespaces within them, but cannot `list` or `get` other namespaces. (Pretty typical I guess.) When one of these users creates a namespace, Rancher sees that and creates/updates the user's cluster/rolebindings for the new namespace according to their project-level permissions. But, this isn't instantaneous, so there's a brief period during which the namespace exists but the user cannot (yet) `get` it, and instead gets a 403 when they try. This causes problems for this provider, because it doesn't distinguish between `IsForbidden` and `IsNotFound` when creating a namespace, so when it hits that short window in `WaitForState` it throws an error (the error is `"waiting for namespace (foo) to be created: unexpected state 'removed', wanted target 'active'"`). To fix this, we've allowed the `StateChangeConf` to wait out the 403s when creating a namespace.

I'm not sure this is the best fix, but was hoping to get some feedback. First, does that explanation make sense, and if so is there a better fix for this problem? I was thinking that instead, perhaps the server should always update the user's `namespaces-edit` clusterrole with the new namespace _before_ creating it, but I haven't had a chance to look into that yet, maybe it does already and I'm missing something else. On the other hand, if this all sounds right, I _think_ namespaced resource types may have a similar issue, for example you might see the same with a `rancher2_secret_v2` resource if the rolebinding that grants `secrets-view` is created after the one that allows accessing the target namespace, right?

### DIFF
--- a/rancher2/resource_rancher2_namespace.go
+++ b/rancher2/resource_rancher2_namespace.go
@@ -67,7 +67,7 @@ func resourceRancher2NamespaceCreate(d *schema.ResourceData, meta interface{}) e
 	d.SetId(newNs.ID)
 
 	stateConf := &resource.StateChangeConf{
-		Pending:    []string{"activating"},
+		Pending:    []string{"activating", "forbidden"},
 		Target:     []string{"active"},
 		Refresh:    namespaceStateRefreshFunc(client, newNs.ID),
 		Timeout:    d.Timeout(schema.TimeoutCreate),
@@ -216,7 +216,7 @@ func resourceRancher2NamespaceDelete(d *schema.ResourceData, meta interface{}) e
 
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"removing"},
-		Target:     []string{"removed"},
+		Target:     []string{"removed", "forbidden"},
 		Refresh:    namespaceStateRefreshFunc(client, id),
 		Timeout:    d.Timeout(schema.TimeoutDelete),
 		Delay:      1 * time.Second,
@@ -238,8 +238,11 @@ func namespaceStateRefreshFunc(client *clusterClient.Client, nsID string) resour
 	return func() (interface{}, string, error) {
 		obj, err := client.Namespace.ByID(nsID)
 		if err != nil {
-			if IsNotFound(err) || IsForbidden(err) {
+			if IsNotFound(err) {
 				return obj, "removed", nil
+			}
+			if IsForbidden(err) {
+				return obj, "forbidden", nil
 			}
 			return nil, "", err
 		}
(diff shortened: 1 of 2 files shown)


rawmind0 (changes requested):
Hi @evhan , thanks for your PR.

Your explanation makes all the sense. Moreover using the provider with non-admin user. For this use case you are right, a race condition may occur on that point. As you mentioned, same should be done for namespaced resources as `rancher2_secret_v2` to aovid .

Your fix is fine ,it's a good way to fix it, but it's incomplete. `namespaceStateRefreshFunc` function is also used at  [resource_rancher2_namespace_test.go#L157](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_namespace_test.go#L157) . Could you please also update the resource tests??

rawmind0:
> Your explanation makes all the sense. Moreover using the provider with non-admin user. For this use case you are right, a race condition may occur on that point. As you mentioned, same should be done for namespaced resources as rancher2_secret_v2 to aovid .

Taking a deeper look into it, i've to amend my last comment. `v2` objects shouldn't be affected due to not checking pending state, [create](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_secret_v2.go#L50), [delete](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_secret_v2.go#L123), just waiting for final state.

evhan:
> Your fix is fine ,it's a good way to fix it, but it's incomplete. `namespaceStateRefreshFunc` function is also used at [resource_rancher2_namespace_test.go#L157](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_namespace_test.go#L157) . Could you please also update the resource tests??

Sure, just updated now. Thanks for having a look, @rawmind0.

> Taking a deeper look into it, i've to amend my last comment. `v2` objects shouldn't be affected due to not checking pending state, [create](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_secret_v2.go#L50), [delete](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_secret_v2.go#L123), just waiting for final state.

I see. In the case of `create` that makes sense, but for `delete` don't we need to consider "forbidden" as an end state, as this PR does for namespace [here](https://github.com/evhan/terraform-provider-rancher2/blob/0f0b581d59b379975a34541ef0e0d98cee9f26cb/rancher2/resource_rancher2_namespace.go#L219)?

rawmind0:
Thanks for the update @evhan.

> I see. In the case of create that makes sense, but for delete don't we need to consider "forbidden" as an end state, as this PR does for namespace here?

On `v2` objects, it's the same on create and delete, as it's just waiting for the final state, `active` or `removed`. It's not taking into account intermediate states, so i think should be fine.

rawmind0 (approved):
Thanks for the update @evhan 

LGTM

PR #754, opened 2021-09-27
fix: support updating logging enable_json_parsing field

vbehar:
right now, the `enable_json_parsing` is only used when reading or creating a new resource, not when updating one.
This fix ensures the `enable_json_parsing` field is also set when updating an existing resource.


rawmind0 (approved):
LGTM

PR #768, opened 2021-10-09
Add `rancher2_principal` data source

NaurisSadovskis:
This PR adds `rancher2_principal` data source. This functionality simplifies email -> principal lookups, making easier to manage IAM binding management in resources like `rancher2_global_role_binding`, `rancher2_project_role_template_binding`, etc. 

Example:
```hcl
# User lookup
data "rancher2_principal" "user" {
  name = "user@google.com"
  type   = "user"
}

output "user_principal" {
  value = data.rancher2_principal.user
}

# Group lookup
data "rancher2_principal" "group" {
  name = "group@google.com"
  type   = "group"
}

output "group_principal" {
  value = data.rancher2_principal.group
}
```

Running `terraform plan` results in: 
```
Changes to Outputs:
  + group_principal = {
      + name         = "group@google.com"
      + id            = "googleoauth_group://..."
      + type          = "group"
    }
  + user_principal  = {
      + name          = "user@google.com"
      + id            = "googleoauth_user://..."
      + type          = "user"
    }
```

**Note**: I have only tested this with the Google provider, but looking at the API, the behaviour should be the same across all other providers.

### DIFF
--- a/docs/data-sources/principal.md
+++ b/docs/data-sources/principal.md
@@ -1,0 +1,30 @@
+---
+page_title: "rancher2_principal Data Source"
+---
+
+# rancher2\_principal Data Source
+
+Use this data source to retrieve information about a Rancher v2 Principal resource.
+
+## Example Usage
+
+```hcl
+data "rancher2_principal" "foo" {
+  email = "group@example.com"
+  type  = "group"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `email` - (Required) The email address of the identity (string)
+* `type` - (Required) The type of the identity (string). Only `group` and `user` values are supported (string)
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - (Computed) The ID of the resource (string)
--- a/docs/data-sources/principal.md
+++ b/docs/data-sources/principal.md
@@ -10,8 +10,7 @@ Use this data source to retrieve information about a Rancher v2 Principal resour
 
 ```hcl
 data "rancher2_principal" "foo" {
-  email = "group@example.com"
-  type  = "group"
+  name = "user@example.com"
 }
 ```
 
@@ -19,8 +18,8 @@ data "rancher2_principal" "foo" {
 
 The following arguments are supported:
 
-* `email` - (Required) The email address of the identity (string)
-* `type` - (Required) The type of the identity (string). Only `group` and `user` values are supported (string)
+* `name` - (Required) The full name of the principal (string)
+* `type` - (Optional) The type of the identity (string). Defaults to `user`. Only `user` and `group` values are supported (string)
 
 
 ## Attributes Reference
--- a/rancher2/data_source_rancher2_principal.go
+++ b/rancher2/data_source_rancher2_principal.go
@@ -72,6 +72,7 @@ func flattenDataSourcePrincipal(d *schema.ResourceData, in *managementClient.Pri
 	}
 
 	d.Set("id", in.ID)
+	d.Set("name", in.Name)
 	d.Set("type", in.PrincipalType)
 
 	return nil
--- a/rancher2/data_source_rancher2_principal.go
+++ b/rancher2/data_source_rancher2_principal.go
@@ -20,10 +20,6 @@ func dataSourceRancher2Principal() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"provider_name": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 		},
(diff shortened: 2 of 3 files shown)


rancher2/data_source_rancher2_principal.go
```diff
@@ -0,0 +1,67 @@
...
+	return flattenDataSourcePrincipal(d, &principals.Data[0])
+}
+
+func flattenDataSourcePrincipal(d *schema.ResourceData, in *managementClient.Principal) error {
+	if in == nil {
+		return nil
+	}
+
+	d.SetId(in.ID)
+	d.Set("id", in.ID)
```
rawmind0: It seems that `in.ID` is assigned twice here. May be a typo trying to assign `email` argument??

```
	d.SetId(in.ID)
-	d.Set("id", in.ID)
+	d.Set("email", in.Name)
```

docs/data-sources/principal.md
```diff
@@ -0,0 +1,30 @@
...
+  email = "group@example.com"
+  type  = "group"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `email` - (Required) The email address of the identity (string)
```
rawmind0: Why are you calling new argument as `email`?? The name of the user is not always the email, it depends of the auth provider. I'd propose you to update it to `name` instead

```
-* `email` - (Required) The email address of the identity (string)
+* `name` - (Optional) The full name of the principal (string)
```

rancher2/data_source_rancher2_principal.go
```diff
@@ -0,0 +1,67 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
rawmind0, line 34: If the `type` argument is required and have limited values, what do you think to add validation and set it as optional with default value `user`??

```
import (
	"fmt"

	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
)

+const (
+	principalTypeGroup = "group"
+	principalTypeUser   = "user"
+)
+
+var (
+	principalTypes                = []string{principalTypeGroup, principalTypeUser}
+)

...

			"type": {
				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Default : principalTypeUser,
+				ValidateFunc: validation.StringInSlice(principalTypes, true),
			},
```

rawmind0 (changes requested):
@NaurisSadovskis , thanks for the PR.

Please, take a look to the changes requested.

rancher2/data_source_rancher2_principal.go
NaurisSadovskis: Thanks, my bad - fixed!

docs/data-sources/principal.md
NaurisSadovskis: Thanks for pointing this out, it makes sense. I've updated the docs and the example.

rancher2/data_source_rancher2_principal.go
NaurisSadovskis: Awesome, thanks for pointing this out - much more elegant :) Fixed!

rancher2/data_source_rancher2_principal.go
rawmind0: Not yet fixed. 

```
	d.SetId(in.ID)
-	d.Set("id", in.ID)
+	d.Set("name", in.Name)
	d.Set("type", in.PrincipalType)
```

rawmind0 (changes requested):

rancher2/data_source_rancher2_principal.go
NaurisSadovskis: Done 🚀

rawmind0: Please, @NaurisSadovskis , use the proper function to set the id

```
-	d.Set("id", in.ID)
+	d.SetId(in.ID)
```

rawmind0 (changes requested):

rancher2/data_source_rancher2_principal.go
NaurisSadovskis: Apologies, done.

rawmind0 (approved):
Thanks for the update @NaurisSadovskis 

LGTM

NaurisSadovskis:
Great, feel free to merge it then! Thanks again for your help!

Issue #765, opened 2021-10-06
Error 403 Forbidden when creating `rancher2_cluster_v2` resource with a "Standard User"'s token


simu:
## Context

I'm provisioning an RKE2 cluster in Rancher 2.6.0 with an API token belonging to an LDAP user with role "Standard User". I can create new RKE2 clusters in the Web interface with the same user without any issues. I can also use the same API token to create a `rancher2_cluster` resource with Terraform without any problems.

Terraform version: v1.0.7
Provider Rancher2 version: v1.20.1
Rancher version: 2.6.0, upgraded from 2.5.9 using Helm chart
Rancher Helm values:

```
auditLog:
  destination: hostPath
  hostPath: /var/log/auditlog/
  level: 2
  maxAge: 180
  maxBackup: 40
  maxSize: 250
hostname: test.rancher.<redacted>
ingress:
  tls:
    source: letsEncrypt
letsEncrypt:
  email: tech@<redacted>

```

Looking at the "Standard User" role definition from that Rancher instance I can see the following interesting snippets (from the YAML view):

```
rules:
[ ... snipped other rules ... ]
- apiGroups:
  - management.cattle.io
  resources:
  - clusters
  verbs:
  - create
[ .. snipped other rules .. ]
- apiGroups:
  - provisioning.cattle.io
  resources:
  - clusters
  verbs:
  - create
[ .. snipped other rules .. ]
```

It looks like standard users don't have permissions to `get` `clusters.provisioning.cattle.io` resources on our Rancher 2.6.0 instance. I can't determine whether that's the default "Standard User" role or if this has been customized in the past on our test instance.

## Actual behavior

Terraform fails with a permission error `Getting cluster V2 ID (fleet-default/rke2-terraform-test): Bad response statusCode [403]. Status [403 Forbidden].` Despite the Terraform error, the RKE2 cluster is created in Rancher and can be inspected in the Rancher web interface.

Find the full `terraform apply` output for the following Terraform below:

```
terraform {
  required_version = ">= 1.0"
  required_providers {
    rancher2 = {
      source  = "rancher/rancher2"
      version = "1.20.1"
    }
  }
}

resource "rancher2_cluster_v2" "cluster" {
  name = "rke2-terraform-test"

  kubernetes_version = "v1.21.4-rke2r2"
}
```

```
simon@phoenix:~/work/k8s/rke2-terraform-mve $ terraform apply

[...]

PR #770, opened 2021-10-11
Conflict and retry fixes to v2 objects

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/765

This PR contains:
* Added `IsConflict` function to retry on update v2 resources
* Added `RestartClients` function to restart Rancher2 clients
* Updated `CatalogV2Client` function to also retry if err `IsNotFound` or `IsForbidden`
* Refactored `activateNodeDriver` and `activateKontainerDriver` functions to retry if got conflict on node driver action
* Refactored v2 resources CRUD functions to be defined at resources files. Added retry if got conflict on update
* Updated `rancher2_cluster_v2` to support creation using Rancher2 standard user token
* Refactored `config.NormalizeURL` function to return error

### DIFF
--- a/rancher2/provider.go
+++ b/rancher2/provider.go
@@ -254,11 +254,10 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 }
 
 func providerValidateConfig(config *Config) (*Config, error) {
-	if config.URL == providerDefaultEmptyString {
-		return &Config{}, fmt.Errorf("[ERROR] No api_url provided")
+	err := config.NormalizeURL()
+	if err != nil {
+		return &Config{}, fmt.Errorf("[ERROR] %v", err)
 	}
-
-	config.URL = NormalizeURL(config.URL)
 	if config.Bootstrap {
 		// If bootstrap tokenkey accesskey nor secretkey can be provided
 		if config.TokenKey != providerDefaultEmptyString {
--- a/rancher2/resource_rancher2_app_v2.go
+++ b/rancher2/resource_rancher2_app_v2.go
@@ -1,12 +1,16 @@
 package rancher2
 
 import (
+	"context"
 	"fmt"
 	"log"
+	"net/url"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/rancher/norman/types"
+	types2 "github.com/rancher/rancher/pkg/api/steve/catalog/types"
 )
 
 func resourceRancher2AppV2() *schema.Resource {
@@ -51,7 +55,7 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
 	}
 	d.Set("system_default_registry", systemDefaultRegistry.Value)
 
-	repo, chartInfo, err := meta.(*Config).InfoAppV2(clusterID, repoName, chartName, chartVersion)
+	repo, chartInfo, err := infoAppV2(meta.(*Config), clusterID, repoName, chartName, chartVersion)
 	if err != nil {
 		return err
 	}
@@ -61,7 +65,7 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
-	chartOperation, err := meta.(*Config).InstallAppV2(clusterID, repo, chartInstallAction)
+	chartOperation, err := createAppV2(meta.(*Config), clusterID, repo, chartInstallAction)
 	if err != nil {
 		return err
 	}
@@ -94,7 +98,7 @@ func resourceRancher2AppV2Read(d *schema.ResourceData, meta interface{}) error {
 		d.Set("system_default_registry", systemDefaultRegistry.Value)
 	}
 	_, rancherID := splitID(d.Id())
-	app, err := meta.(*Config).GetAppV2ByID(clusterID, rancherID)
+	app, err := getAppV2ByID(meta.(*Config), clusterID, rancherID)
 	if err != nil {
 		if IsNotFound(err) || IsForbidden(err) {
 			log.Printf("[INFO] App V2 %s not found at %s", name, clusterID)
@@ -114,7 +118,7 @@ func resourceRancher2AppV2Update(d *schema.ResourceData, meta interface{}) error
 	chartVersion := d.Get("chart_version").(string)
 	log.Printf("[INFO] Updating App V2 %s at %s", name, clusterID)
 
-	repo, chartInfo, err := meta.(*Config).InfoAppV2(clusterID, repoName, chartName, chartVersion)
+	repo, chartInfo, err := infoAppV2(meta.(*Config), clusterID, repoName, chartName, chartVersion)
 	if err != nil {
 		return err
 	}
@@ -123,7 +127,7 @@ func resourceRancher2AppV2Update(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
-	chartOperation, err := meta.(*Config).UpgradeAppV2(clusterID, repo, chartUpgradeAction)
+	chartOperation, err := upgradeAppV2(meta.(*Config), clusterID, repo, chartUpgradeAction)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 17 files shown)


rancher2/provider.go
```diff
@@ -254,11 +254,10 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 }
 
 func providerValidateConfig(config *Config) (*Config, error) {
-	if config.URL == providerDefaultEmptyString {
-		return &Config{}, fmt.Errorf("[ERROR] No api_url provided")
+	err := config.NormalizeURL()
+	if err != nil {
+		return &Config{}, fmt.Errorf("[ERROR] %v", err)
```
nickgerace, line 259: Why don't we just return the same config back that was given to us or return `nil`?

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+		return err
+	}
+	resource := &types.Resource{
+		ID:      app.ID,
+		Type:    app.Type,
+		Links:   app.Links,
+		Actions: app.Actions,
+	}
+	var resp interface{}
+	return client.Action(appV2APIType, "uninstall", resource, map[string]interface{}{}, resp)
```
nickgerace, line 332: For `map[string]interface{}{}` should we not use `nil` and have the calling function `make` its own map as needed?

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+		return nil, nil, err
+	}
+	resource := types.Resource{
+		ID:      repo.ID,
+		Type:    repo.Type,
+		Links:   repo.Links,
+		Actions: repo.Actions,
+	}
+	link := "info"
+	if resource.Links == nil && len(resource.Links[link]) == 0 {
```
nickgerace: What happens if one condition is satisfied and not the other?

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+	if len(clusterID) == 0 || len(id) == 0 {
+		return nil, fmt.Errorf("Getting app V2: Cluster ID and/or App V2 ID is nil")
+	}
+	resp := &AppV2{}
+	err := c.getObjectV2ByID(clusterID, id, appV2APIType, resp)
+	if err != nil {
+		if !IsServerError(err) && !IsNotFound(err) && !IsForbidden(err) {
+			return nil, fmt.Errorf("Getting App V2: %s", err)
+		}
+		return nil, err
```
nickgerace, line 402: This looks functionally the same, but with a wrapper. Is that intentional?

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+	if len(clusterID) == 0 || len(id) == 0 {
+		return nil, fmt.Errorf("Getting app V2 operation: Cluster ID and/or App V2 ID is nil")
+	}
+	resp := map[string]interface{}{}
+	err := c.getObjectV2ByID(clusterID, id, appV2OperationAPIType, &resp)
+	if err != nil {
+		if !IsServerError(err) && !IsNotFound(err) && !IsForbidden(err) {
+			return nil, fmt.Errorf("Getting App V2 logs: %s", err)
+		}
+		return nil, err
```
nickgerace, line 420: Same question as above

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+		return "", fmt.Errorf("Getting app V2 operation logs: Provider config is nil")
+	}
+	if len(clusterID) == 0 {
+		return "", fmt.Errorf("Getting app V2 operation logs: Cluster ID is nil")
+	}
+	if op["id"].(string) == "" {
+		return "", fmt.Errorf("Getting app V2 operation logs: App V2 operation id is nil")
+	}
+
+	links := toMapString(op["links"].(map[string]interface{}))
```
nickgerace: Should we check if this conversion will fail, or do we _know_ its value always?

rancher2/resource_rancher2_secret_v2.go
```diff
@@ -148,3 +150,115 @@ func secretV2StateRefreshFunc(meta interface{}, clusterID, secretID string) reso
...
+	}
+	if secret == nil {
+		return nil, fmt.Errorf("Creating secret V2: object is nil")
+	}
+	// Converting secret V2 object to map[string]interface{} as type fields is duplicated
+	secret2, err := interfaceToMap(secret)
+	if err != nil {
+		return nil, err
+	}
+	secret2["type"] = secret2["_type"]
```
nickgerace, line 170: What is going on here? I'm a bit confused 😅

rancher2/resource_rancher2_secret_v2.go
```diff
@@ -148,3 +150,115 @@ func secretV2StateRefreshFunc(meta interface{}, clusterID, secretID string) reso
...
+	if len(clusterID) == 0 || len(id) == 0 {
+		return nil, fmt.Errorf("Getting secret V2: Cluster ID and/or Secret V2 ID is nil")
+	}
+	resp := &SecretV2{}
+	err := c.getObjectV2ByID(clusterID, id, secretV2APIType, resp)
+	if err != nil {
+		if !IsServerError(err) && !IsNotFound(err) && !IsForbidden(err) {
+			return nil, fmt.Errorf("Getting secret V2: %s", err)
+		}
+		return nil, err
```
nickgerace, line 211: Same question as further above

rancher2/resource_rancher2_app_v2.go
rawmind0: Yes, this is intentional. Returning err as is if `IsServerError` or `IsNotFound` or `IsForbidden`, due to can be evaluated at other functions.

rancher2/resource_rancher2_app_v2.go
rawmind0: Same answer as above

rancher2/resource_rancher2_app_v2.go
rawmind0: Good catch.. Should be or , updating

rancher2/resource_rancher2_app_v2.go
```diff
@@ -243,3 +247,202 @@ func appV2OperationWait(meta interface{}, clusterID, opID string) error {
...
+	link := "logs"
+	if links == nil && len(links[link]) == 0 {
```
rawmind0: Here should else be a or

rancher2/resource_rancher2_app_v2.go
rawmind0: Adding checking

rancher2/resource_rancher2_secret_v2.go
rawmind0: Same answer as above

rancher2/provider.go
rawmind0: Yes, this should work but in fact is generating a dump, so returning empty object

rancher2/resource_rancher2_app_v2.go
rawmind0: The `client.Action` call was failing (at least in the past, not sure if it's addressed) if setting nil instead of `map[string]interface{}{}`

rancher2/resource_rancher2_app_v2.go
rawmind0: Updated

rancher2/resource_rancher2_app_v2.go
rawmind0: Updated

rancher2/resource_rancher2_secret_v2.go
rawmind0: This is due to the secret2 struct, https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_secret_v2.go#L18 . `norman.Resource` and  `v1.Secret` has a duplicated `type` json definition. That's why the object is first converted to map[string]interface{}

nickgerace: Ah gotcha. Thanks

nickgerace (approved):

PR #772, opened 2021-10-12
fix importing of AKS clusters

### LINKED ISSUE #771: Errors when attempting to import an existing AKS cluster
When attempting import an existing AKS cluster by creating a new AKS cluster with "imported" set to true, many unnecessary fields are populated and sent to Rancher, causing the Rancher server to error out as it does not allow changes to those properties during an import.

polivbr:
When importing an existing AKS cluster most fields should be left
unpopulated as Rancher is not expecting them to be set and will
error out on any discrepancy.

Closes rancher/terraform-provider-rancher2#771


rawmind0:
Superseded by https://github.com/rancher/terraform-provider-rancher2/pull/775

Issue #792, opened 2021-11-04
Add option enable_cri_dockerd in rke_config


nickvth:
Please add this in the rancher2_cluster Resource under rke_config: https://rancher.com/docs/rke/latest/en/config-options/#cri-dockerd

So we can use cri_dockerd in the future.

rawmind0:
Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/794 adding the new feature

rawmind0:
PR https://github.com/rancher/terraform-provider-rancher2/pull/794 is already merged. The feature will be available on next release.

nickvth:
Great job @rawmind0

Issue #791, opened 2021-11-03
cluster_registration_token is randomly missing node_command which causes nodes fail to join


riuvshyn:
rancher version: 2.6.2
provider 1.21.0
terraform 1.0.8

`cluster_registration_token` object randomly missing `node_command` field after initial apply which causes all nodes fail to join.

example of terraform state
```
                "cluster_registration_token": [
                  {
                    "annotations": {},
                    "cluster_id": "c-5r4gg",
                    "command": "",
                    "id": "c-5r4gg:default-token",
                    "insecure_command": "",
                    "insecure_node_command": "",
                    "insecure_windows_node_command": "",
                    "labels": {},
                    "manifest_url": "",
                    "name": "default-token",
                    "node_command": "",
                    "token": "<EDITED>",
                    "windows_node_command": ""
                  }
                ],
```

if I re run `terraform apply` it actually updates `cluster_registration_token` attributes with values

example of terraform state after re-apply
```
                "cluster_registration_token": [
                  {
                    "annotations": {},
                    "cluster_id": "c-5r4gg",
                    "command": "kubectl apply -f https://<EDITED>/v3/import/<EDITED>-5r4gg.yaml",
                    "id": "c-5r4gg:default-token",
                    "insecure_command": "curl --insecure -sfL https://<EDITED>/v3/import/<EDITED>-5r4gg.yaml | kubectl apply -f -",
                    "insecure_node_command": "",
                    "insecure_windows_node_command": "",
                    "labels": {},
                    "manifest_url": "https://<EDITED>/v3/import/<EDITED>-5r4gg.yaml",
                    "name": "default-token",
                    "node_command": "sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run  <EDITED>/rancher/rancher-agent:v2.6.2 --server https://<EDITED> --token <EDITED> --ca-checksum <EDITED>",
                    "token": "<EDITED>",
                    "windows_node_command": "PowerShell -NoLogo -NonInteractive -Command \"& {docker run -v c:\\:c:\\host -e AGENT_IMAGE=<EDITED>/rancher/rancher-agent:v2.6.2 <EDITED>/rancher/rancher-agent:v2.6.2 bootstrap --server https://<EDITED> --token <EDITED> --ca-checksum <EDITED> | iex}\""
                  }
                ],
```

[...]

Issue #788, opened 2021-11-02
Cannot delete v2 provisioning cluster that was successfully provisioned and active at one point, but is no longer available


Oats87:
If you create a v2 provisioning cluster with Rancher v2.6.2+ using Terraform, i.e.

https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster_v2

If you delete the underlying machines or if the cluster becomes unavailable for any reason i.e. cluster-agent dies or something, you cannot delete it via Terraform because Terraform gets stuck trying to refresh the state for the cluster

```
chriskim@Endeavor tf % terraform destroy -auto-approve
rancher2_cluster_v2.cluster: Refreshing state... [id=fleet-default/mycluster]
╷
│ Error: Setting cluster V2 legacy data: Getting cluster Kubeconfig: Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [message=Post "https://10.43.0.1:443/apis/cluster.cattle.io/v3/namespaces/cattle-system/clusterauthtokens": net/http: TLS handshake timeout, baseType=error, code=ServerError] from [https://<my-rancher>/v3/clusters/c-m-gtxldgsk?action=generateKubeconfig]
│
│   with rancher2_cluster_v2.cluster,
│   on main.tf line 14, in resource "rancher2_cluster_v2" "cluster":
│   14: resource "rancher2_cluster_v2" "cluster" {
│
╵
```

The cluster was created with the snippet:
```
terraform {
	required_providers {
	  rancher2 = {
		source = "rancher/rancher2"
		}
	}
}

provider "rancher2" {
  api_url    = "<my-rancher>"
  token_key = "<goaway>"
}

resource "rancher2_cluster_v2" "cluster" {
    name = "mycluster"
    kubernetes_version = "v1.21.5+rke2r2"

    rke_config {
        etcd {
            # Broken in rancher, see rancher/rancher#35163
            # s3_config {}
        }
        local_auth_endpoint {
            enabled = true
        }
    }
}
```

I then logged into Rancher and used the custom cluster registration command to register the cluster and have it come up to be active

I deleted the underlying infrastructure machines which made the cluster go unavailable, then tried to `destroy` it only to be blocked on destruction.

/cc @braunsonm

rawmind0:
Issue reproduced. Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/794 with a patch for it

rawmind0:
[...]

Issue #784, opened 2021-10-26
Support labels and annotations for rancher2_cluster_v2


kgtw:
The `rancher2_cluster` resource supports annotations & labels (https://github.com/rancher/terraform-provider-rancher2/issues/582), this functionality appears to be missing for the `rancher2_cluster_v2` resource.

rawmind0:
Labels and annotations are already supported by `rancher2_cluster_v2` https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_v2.go#L91

kgtw:
Ok, looks like docs need updating as there is no mention of them being available for the resource.

rawmind0:
You are right @kgtw . We'll update docs properly.

rawmind0:
Fixed with PR https://github.com/rancher/terraform-provider-rancher2/pull/794

rawmind0:
PR https://github.com/rancher/terraform-provider-rancher2/pull/794 is already merged. The fix will be available on next release

PR #794, opened 2021-11-08
Cluster fixes and enhancements

rawmind0:
This PR contains:
* Added `rancher2_cluster.rke_config.enable_cri_dockerd` argument. https://github.com/rancher/terraform-provider-rancher2/issues/792
* Updated `rancher2_cluster_v2` docs adding labels and annotations arguments. https://github.com/rancher/terraform-provider-rancher2/issues/784
* Updated `findClusterRegistrationToken` function checking for correct Cluster Registration Token https://github.com/rancher/terraform-provider-rancher2/issues/791
* Updated `getClusterKubeconfig` function to properly delete a cluster if cluster not available https://github.com/rancher/terraform-provider-rancher2/issues/788


rancher2/resource_rancher2_cluster.go
```diff
@@ -531,14 +531,19 @@ func findFlattenClusterRegistrationToken(client *managementClient.Client, cluste
...
+		for retry, retries := 1, 10; retry <= retries; retry++ {
+			regToken, err := client.ClusterRegistrationToken.ByID(regTokenID)
+			if err != nil {
+				if !IsNotFound(err) {
+					return nil, err
+				}
+				break
+			}
+			if (len(regToken.Command) > 0 && len(regToken.NodeCommand) > 0) || retry == retries {
+				return regToken, nil
```
superseb, line 543: Isn't reaching maximum retries and not having a regToken considered an error? Or is this expected and handled by when its called? If it's empty and retries has been reached, regToken is not complete and can't be used so it should be an error or no?

rawmind0: It's indeed.  If the regToken is found and retries has been reached, a partial regToken is returned instead of error. The regToken is just a computed field for the cluster resource.

superseb (approved):

riuvshyn:
@rawmind0 thanks for addressing these issues, really looking forward for next release, is there any change to have 1.21.1 with these changes 🙏🏽 🙇🏽

Issue #693, opened 2021-06-24
eks-config-operator fails during eks node group provisioning and node template is empty


wojtasool:
`During` creation of new node_group with launch template, rancher does not see launch template in the UI and the eks-operator fails.
When I edit cluster and manually point new node group to use specific launch_template then it works.
Otherwise it is empty.

I define environment using terraform provider 1.15.1

Plan also shows that launch_template will be created and used within eks node_group.

This is the code i use:

`resource "rancher2_cluster" "EKS-custom" {
  depends_on  = [var.rancher_dns, var.rancher_route_record_depends, var.rancher2_token, aws_security_group.eks_sg_allowall, aws_launch_template.eks_launch_template]
  name        = var.cluster-name
  description = "EKS Kubernetes cluster"

  eks_config_v2 {
    cloud_credential_id = rancher2_cloud_credential.eks_credential.id
    kubernetes_version  = var.kubernetes_version
    region              = var.region

    dynamic node_groups {
      for_each = var.node_groups
      content {
        name                   = node_groups.value.name
        instance_type          = try(node_groups.value.spot, false) ? "" : node_groups.value.instance_type
        desired_size           = node_groups.value.desired_size
        max_size               = node_groups.value.max_size
        min_size               = node_groups.value.min_size
        ec2_ssh_key            = aws_key_pair.eks_key_pair.key_name
        request_spot_instances = try(node_groups.value.spot, "false")
        spot_instance_types    = [try(node_groups.value.spot, false) ? node_groups.value.instance_type : ""]
        gpu                    = try(node_groups.value.gpu, "false")
        disk_size              = (try(node_groups.value.disk_size, 0) > 0) ? node_groups.value.disk_size : "60"
        dynamic launch_template {
          for_each = (try(node_groups.value.longhorn_disk_size, 0) > 0) ? [1] : []
          content {
            id      = aws_launch_template.eks_launch_template[node_groups.value.name].id
            version = aws_launch_template.eks_launch_template[node_groups.value.name].latest_version
            name    = aws_launch_template.eks_launch_template[node_groups.value.name].name
          }
        }
        resource_tags = merge(
          {
            Creator = "Rancher-Terraform"
          },
          var.tags,
        )
        tags = merge(
          {
            Name    = "${var.cluster-name}-node"
            Creator = "Terraform"
          },
          var.tags,
        )
      }
    }
    public_access   = false
    private_access  = true
[...]

PR #805, opened 2021-11-18
Flatten EKS launch template config to map

thedadams:
The EKS launch template config was being left as a slice. However,
Rancher expects this config as a map. Therefore the flatten/expand
functions for EKS launch template configs are updated to output/input
maps.

Issue:
https://github.com/rancher/terraform-provider-rancher2/issues/693

### DIFF
--- a/rancher2/structure_cluster_eks_config_v2.go
+++ b/rancher2/structure_cluster_eks_config_v2.go
@@ -6,7 +6,7 @@ import (
 
 // Flatteners
 
-func flattenClusterEKSConfigV2NodeGroupsLaunchTemplate(in *managementClient.LaunchTemplate, p []interface{}) []interface{} {
+func flattenClusterEKSConfigV2NodeGroupsLaunchTemplate(in *managementClient.LaunchTemplate, p []interface{}) map[string]interface{} {
 	if in == nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


a-blender (approved):
Not a terraform SME yet but if this has been smoke tested to make sure the new func produces a correctly configured launch template, LGTM! Build is also failing

thedadams:
@annablender Thanks! `gofmt` strikes again.

thedadams:
I'm not sure how to interpret the current test failures. They seem unrelated to the changes here.

rawmind0 (changes requested):
@thedadams , i don't think this change is required. 

The [`launch_template`](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_cluster_eks_config_v2.go#L89) argument is defined as [terraform list](https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist) which has an `[]interface{}` data structure. The flatten and expand functions are "converting" from rancher data structure field [LaunchTemplate](https://github.com/rancher/rancher/blob/release/v2.6/pkg/client/generated/management/v3/zz_generated_node_group.go#L33) to tf data structure argument `[]interface{}`.

> I'm not sure how to interpret the current test failures. They seem unrelated to the changes here.

The test is failing due to expander is not getting expected value `structure_cluster_test.go:706: Unexpected output from expander` when testing EKS cluster

thedadams:
@rawmind0 What I am trying to say with this PR is the launch template data should not be a list. It should just be a map.

The issue at hand here is when the launch template data gets passed to Rancher as a list, the ID and version gets dropped because the Go unmarshaler does not expect a slice.

rawmind0:
@thedadams , the launch_template data is not passed as a slice to Rancher, it's passed as [`*LaunchTemplate`](https://github.com/rancher/rancher/blob/release/v2.6/pkg/client/generated/management/v3/zz_generated_node_group.go#L33) type that it's expected by Rancher API cli . `launch_template` argument is just set as slice at terraform state but not at Rancher.

PR #810, opened 2021-11-22
Fix node_template.vsphere_config.cloud_config description

spawnia:
The documentation indicates that the configuration option should be passed the path to a cloud-config file,
however it actually requires the contents of the file to be passed, e.g. using the HCL function `file()`.


rawmind0 (approved):
LGTM

Issue #806, opened 2021-11-19
rancher2_cluster_v2 ACE enabled does not work on RKE2


braunsonm:
Given:

```
resource "rancher2_cluster_v2" "example" {
  ...
  rke_config {
    local_auth_endpoint {
      enabled = true
    }
  }
}
```

The endpoint will not be enabled and if you download your kubeconfig there will be no ACE endpoints added.

rawmind0:
Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/815 to fix this issue

rawmind0:
PR #815 is already merged. The fix will be available at next tfp release. Please, reopen issue if required.

PR #815, opened 2021-12-07
Bump Rancher dependency to fix EKS launch template issue

thedadams:
There is an issue where EKS launch templates did not get sent to Rancher as
expected. This was because the ID field was missing in the Rancher client. This
has been updated in Rancher and this change brings those changes here.

In addition, there were other changes in the Rancher client. This change brings
in those changes as well (mostly related to ACE config on RKE2 clusters).


rancher2/schema_cluster_v2_rke_config.go
```diff
@@ -16,15 +16,6 @@ func clusterV2RKEConfigFields() map[string]*schema.Schema {
...
 		},
-		"local_auth_endpoint": {
-			Type:        schema.TypeList,
-			MaxItems:    1,
-			Optional:    true,
-			Description: "Cluster V2 local auth endpoint",
-			Elem: &schema.Resource{
-				Schema: clusterV2RKEConfigLocalAuthEndpointFields(),
-			},
-		},
```
rawmind0, line 27: To follow provider attribute removal [best practices](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal), Deprecate the field should be better option instead of removing it. 

```
		"local_auth_endpoint": {
			Type:        schema.TypeList,
			MaxItems:    1,
			Optional:    true,
+                       Deprecated: "Use rancher2_cluster_v2.local_auth_endpoint instead",
			Description: "Cluster V2 local auth endpoint",
			Elem: &schema.Resource{
-				Schema: clusterV2RKEConfigLocalAuthEndpointFields(),
+				Schema: clusterV2LocalAuthEndpointFields(),
			},
		},

```

rawmind0 (changes requested):
@thedadams thanks for this PR.

Please take a look to the comments at the code. 

Also, to complete the PR could you please update the docs adding new argument description on [this section](https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/cluster_v2.md?plain=1#L213) and updating [deprecated argument description](https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/cluster_v2.md?plain=1#L250)

rawmind0:
This PR should fix issues https://github.com/rancher/terraform-provider-rancher2/issues/693 and https://github.com/rancher/terraform-provider-rancher2/issues/806

bashofmann:
Is this maybe also related to https://github.com/rancher/terraform-provider-rancher2/issues/820?

rawmind0:
@bashofmann , yes, this is related to https://github.com/rancher/terraform-provider-rancher2/issues/820

rancher2/schema_cluster_v2_rke_config.go
thedadams: Updated.

rawmind0 (approved):
Thanks for the update @thedadams 

LGTM

PR #831, opened 2021-12-27
Policy validator add specialized verbs

zerotens:
Kubernetes has specialized verbs for specific use cases.

https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb

This PR adds the specialized verbs to the validator list and allows creating role templates using those specialized verbs.


rawmind0 (approved):
Thanks for the PR @zerotens 

LGTM

Issue #802, opened 2021-11-17
rancher2_cluster: inconsistent final plan produced


airbjorn:
It seems, that in some cases rancher2_cluster thinks there is something to do, although it isn't.

- We are using custom cacerts for authorization.
- Even without changes, calling `terraform plan` shows that there is something to do. The plan itself exits with 0.
- When calling `terraform apply` however, we get this error:
 ```
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for local_file.kubeconfig_cluster to include new values learned so far during apply, provider
│ "registry.terraform.io/hashicorp/local" produced an invalid new value for .content: inconsistent values for sensitive attribute.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
- trace log see [tf-trace.log](https://github.com/rancher/terraform-provider-rancher2/files/7553997/tf-trace.log)
- definitions of the Ressources "cluster" and "local_file" see [cluster-local_file.tf](https://github.com/rancher/terraform-provider-rancher2/files/7554002/cluster-local_file.log)
- Workaround / Quick Fix is shown in the file [cluster-local_file.diff](https://github.com/rancher/terraform-provider-rancher2/files/7561889/cluster-local_file.diff.txt)
- used versions:
Rancher 2.6.2
```
Terraform v1.0.11
on linux_amd64
+ provider registry.terraform.io/gavinbunney/kubectl v1.13.1
+ provider registry.terraform.io/hashicorp/dns v3.2.1
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/tls v3.1.0
+ provider registry.terraform.io/rancher/rancher2 v1.21.0
+ provider registry.terraform.io/rancher/rke v1.2.4
```

rawmind0:
Related with issue https://github.com/rancher/terraform-provider-rancher2/issues/789

rawmind0:
Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/837 to fix the issue

rawmind0:
The fix has been included on new release v1.22.2. Please, reopen issue if needed

Issue #789, opened 2021-11-02
Rancher User for provider gets a new Kubeconfig token (kube_config) each time ran


mitchellmaler:
We have noticed that the user used for this provider (providing the bearer token) gets a new Kubeconfig token generated each time we run. If we run plan then it would create two (plan and apply). We grab the kube_config from the rancher2_cluster datasource or created rancher2_cluster output. Is there a reason it generates a new kube_config each time or should it be using the existing one and one shouldn't be always created.

pwurbs:
Yes, this very annoying, because many tokens are created which are all valid.

rawmind0:
Issue reproduced at Rancher v2.6.x. 
Related with issue https://github.com/rancher/terraform-provider-rancher2/issues/802

rawmind0:
Submitted PR https://github.com/rancher/terraform-provider-rancher2/pull/837 to fix the issue

sgapanovich:
### Reproduction ###

Rancher version: 2.6.3
Rancher cluster type: HA
Terraform provider rancher 2 version:  "1.21.0"

----------

New Kubeconfig token is created every time when `terraform apply` or `terraform plan` is run with the following resources
- rancher2_cluster
- rancher2_cluster_v2
- rancher2_cluster_sync
![apply and plan](https://user-images.githubusercontent.com/40282141/148284922-28928784-b61e-4f88-b60c-f083a1047bac.jpg)

the above resources should generate tokens when run first time but should use existing ones when re-running 
- new token generated every time i run `terraform apply` or `terraform plan` (even when there is no changes)

rawmind0:
@sgapanovich  the fix has been included on new release v1.22.2.

sgapanovich:
### Test Environment: ###
Rancher version: 2.6.3
Rancher cluster type: HA
Terraform provider rancher 2 version: "1.22.2"

----------

### Testing: ###

`rancher2_cluster` and `rancher2_cluster_sync` and `rancher2_cluster_v2`:
when running `tf plan` first time - no token is created
when running `tf apply` first time - token is created 
when cluster is provisioned running `tf plan` or `tf apply` does not create new tokens 
if the generated kubeconfig-user token for the resource is deleted (in rancher)
-`tf plan` will generate a new token (will be creating new tokens every time until a token created by `tf apply` assigned to the resource)
-`tf apply` will generate a new token (the token will be assigned to the resource)

----

[...]

PR #837, opened 2022-01-05
Fix kube_config generation function for Rancher 2.6.0 and above

rawmind0:
Issue https://github.com/rancher/terraform-provider-rancher2/issues/789
Issue https://github.com/rancher/terraform-provider-rancher2/issues/802

### DIFF
--- a/rancher2/resource_rancher2_token.go
+++ b/rancher2/resource_rancher2_token.go
@@ -117,3 +117,24 @@ func resourceRancher2TokenDelete(d *schema.ResourceData, meta interface{}) error
 	d.SetId("")
 	return nil
 }
+
+func isTokenValid(c *Config, id string) (bool, error) {
+	if len(id) == 0 {
+		return false, nil
+	}
+
+	client, err := c.ManagementClient()
+	if err != nil {
+		return false, err
+	}
+
+	token, err := client.Token.ByID(id)
+	if err != nil {
+		if !IsNotFound(err) && !IsForbidden(err) {
+			return false, err
+		}
+		return false, nil
+	}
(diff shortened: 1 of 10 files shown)


rancher2/resource_rancher2_token.go
```diff
@@ -117,3 +117,24 @@ func resourceRancher2TokenDelete(d *schema.ResourceData, meta interface{}) error
 	d.SetId("")
 	return nil
 }
+
+func isTokenValid(c *Config, id string) (bool, error) {
+	if len(id) == 0 {
+		return false, nil
+	}
```
rmweir, line 124: Wouldn't a kubeconfig with CLI commands instead of a token fail this? There is a setting `kubeconfig-generate-token` that when ~enabled~ set to "false" CLI commands are added to kubeconfig to generate a token. Instead of embedding it into the kubeconfig it just lets the CLI handle it each time, CLI does cache so that won't be a problem. However, I think this may interpret that setup as not having a valid token.

rmweir (changes requested):

rancher2/resource_rancher2_token.go
rawmind0: Nope, i've tested the user case and is working fine. The token is replaced just if not empty https://github.com/rancher/terraform-provider-rancher2/pull/837/files#diff-035d1bb527d3ce1fde79a8bf6a191e1c441f7e2074f17caa106ba35a66fefa0bR697 , when `kubeconfig-generate-token=false` the `generateKubeconfig` action is not generating token nor diff at kube_config

rmweir (approved):
LGTM

Issue #800, opened 2021-11-16
Rancher delete existing node_groups and logging_types config  EKS imported cluster 


abelnieva:
I have created EKS cluster with terraform , then imported to rancher, but rancher change the config in imported clusters, i dont want to  have rancher killing all cluster nodes  

code 

resource "rancher2_cluster" "cluster" {
  name                      = lower(var.cluster_name)
  description               = "Terraform EKS cluster"
  enable_cluster_monitoring = var.enable_cluster_monitoring
  eks_config_v2 {
    cloud_credential_id = data.rancher2_cloud_credential.rancher2norcal.id
    name                = var.cluster_name
    region              = var.region
    **imported            = true**
  }
}

terraform plan 

      **~ eks_config_v2 {
            cloud_credential_id   = "cattle-global-data:cc-rgp2g"
            imported              = true
          ~ logging_types         = [
              - "api",
              - "audit",
              - "authenticator",
              - "controllerManager",
              - "scheduler",
            ]
            name                  = "cluster-ie-eks"
            private_access        = false
            public_access         = false
            public_access_sources = []
            region                = "us-west-2"
            secrets_encryption    = false
            security_groups       = []
            subnets               = []
            tags                  = {}

          ~ node_groups {
              ~ desired_size           = 1 -> 2
                disk_size              = 20
                gpu                    = false
              ~ instance_type          = "t2.large" -> "t3.medium"
                labels                 = {
                    "type" = "custom"
                }
              ~ max_size               = 12 -> 2
              ~ min_size               = 1 -> 2
                name                   = "test-spot"
                request_spot_instances = false
                resource_tags          = {}
                spot_instance_types    = []
                subnets                = [
                    "subnet-0466aa...36738",
                    "subnet-05440....a00",
                    "subnet-0e2d..6b6...3c",
                ]
                tags                   = {}
                version                = "1.21"
            }
          ~ node_groups {
              ~ desired_size           = 1 -> 2
                disk_size              = 20
                gpu                    = false
              ~ instance_type          = "t2.large" -> "t3.medium"
                labels                 = {
                    "type" = "primary-infra"
                }
              ~ max_size               = 10 -> 2
              ~ min_size               = 1 -> 2
                name                   = "us-west-2a"
                request_spot_instances = false
                resource_tags          = {}
                spot_instance_types    = []
                subnets                = [
                    "subnet-0466a...68bc36738",
                ]
                tags                   = {}
                version                = "1.21"
            }
[...]

PR #843, opened 2022-01-07
ignore existing eks node_groups, labels, and tags if it is an imported cluster

kalinon:
Should solve rancher/terraform-provider-rancher2#800

### DIFF
--- a/rancher2/structure_cluster_eks_config_v2.go
+++ b/rancher2/structure_cluster_eks_config_v2.go
@@ -121,8 +121,14 @@ func flattenClusterEKSConfigV2(in *managementClient.EKSClusterConfigSpec, p []in
 	if in.KubernetesVersion != nil && len(*in.KubernetesVersion) > 0 {
 		obj["kubernetes_version"] = *in.KubernetesVersion
 	}
+	if len(in.Region) > 0 {
+		obj["region"] = in.Region
+	}
 
 	obj["imported"] = in.Imported
+	if in.Imported {
+		return []interface{}{obj}
+	}
 
 	if in.KmsKey != nil && len(*in.KmsKey) > 0 {
 		obj["kms_key"] = *in.KmsKey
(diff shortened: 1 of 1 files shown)


thedadams:
I don't think this fixes the problem. If these fields are left out, then the will be passed as `nil` to the client. The problem is between the client and Rancher.

kalinon:
It does, because rancher auto imports changes from AWS on its side. By passing null, it doesn't update it. The rancher controller will pull the values, but if it's defined it will see a Delta in the configs and perform a change.

thedadams:
Sorry, let me be more clear.

If you leave these items out of the terraform file or if you make the change that you're making here, then you have achieved the same thing. So, if you put these items in the terraform file, then your change will take them out.

There is still a problem after that. For example, if you take out the `nodegroups`, then the client that the terraform provider uses will still send `nodegroups: []`. And, if your EKS cluster has any nodegroups in it, they will be deleted (because there is a delta as you described).

So, this change will protect a user if they mistakenly put in the wrong items, but there is still the problem that I am describing that happens with the client.

rawmind0:
The PR is incomplete. To fix the problem, the same logic has to be applied the the `flattenClusterEKSConfigV2` and to the `expandClusterEKSConfigV2` functions. 

Take a look how it's done at `aks_config_v2`, https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_aks_config_v2.go#L68 and 
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_aks_config_v2.go#L241

kalinon:
@rawmind0 i gave it a go! Please let me know if thats what you meant.

rawmind0 (changes requested):
Thanks for the update @kalinon , but it wasn't what i meant. Let me try to explain better.

If the eks cluster is imported, just `cloud_credential_id, `name`, `region` and  `imported` arguments should be filled.  So the PR should treat this fields and as [mentioned](https://github.com/rancher/terraform-provider-rancher2/pull/843#issuecomment-1020028406), return if cluster is imported. The code should be something like that

```
	//FILL required arguments for imported eks cluster
	if v, ok := in["imported"].(bool); ok {
		obj.Imported = v
		if obj.Imported {
			// Return if the cluster is imported
			return obj
		}
	}
	// FILL the rest of the arguments for NON imported eks cluster
```

Also, the cluster docs should be updated for `eks_config_v2`, with something like that, https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#aks_config_v2

kalinon:
@rawmind0 ah, i better understand now. thanks. Let me fix that up.

kkaempf:
closing stale PRs.

PR #847, opened 2022-01-13
Add support to import cloud credentials

BobVanB:
**Problem**

The resource in rancher2 already exists.
We use a remote state file, that i can't edit.

**Result**

```
Error: resource rancher2_cloud_credential doesn't support import
```

**Solution**

Support import for `rancher2_cloud_credentials`

```bash
terraform import rancher2_cloud_credential.resource_id rancher_id
```

The reason why i use a range loop is because the API does not give me a "driver" field to retrieve.
`d.Get("driver")` yields no results and ends up in the got no driver error.

**Better result**

```
rancher2_cloud_credential.resource: Importing from ID "cattle-global-data:resource"...
rancher2_cloud_credential.resource: Import prepared!
  Prepared rancher2_cloud_credential for import
rancher2_cloud_credential.resource: Refreshing state... [id=cattle-global-data:resource]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

### DIFF
--- a/docs/resources/cloud_credential.md
+++ b/docs/resources/cloud_credential.md
@@ -123,20 +123,13 @@ The following attributes are exported:
 
 ## Import
 
-Node Template can be imported using the Rancher Node Template ID
+Cloud Credential can be imported using the Cloud Credential ID and the Driver name.
 
 ```bash
-terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt;.&lt;driver&gt;
+terraform import rancher2_cloud_credential.foo &lt;CLOUD_CREDENTIAL_ID&gt;.&lt;DRIVER&gt;
 ```
 
-### Argument Reference
-
-The following arguments are supported:
-
-* `cloud_credential_id` - (Required) The ID of the Cloud Credential (string)
-* `driver` - (Required) The driver of the Cloud Credential (string)
-
-Supported drivers:
+The following drivers are supported:
 
 * amazonec2
 * azure
--- a/docs/resources/cloud_credential.md
+++ b/docs/resources/cloud_credential.md
@@ -126,7 +126,7 @@ The following attributes are exported:
 Node Template can be imported using the Rancher Node Template ID
 
 ```bash
-terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt.&lt;driver&gt;
+terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt;.&lt;driver&gt;
 ```
 
 ### Argument Reference
--- a/docs/resources/cloud_credential.md
+++ b/docs/resources/cloud_credential.md
@@ -120,3 +120,29 @@ The following attributes are exported:
 - `create` - (Default `10 minutes`) Used for creating cloud credentials.
 - `update` - (Default `10 minutes`) Used for cloud credential modifications.
 - `delete` - (Default `10 minutes`) Used for deleting cloud credentials.
+
+## Import
+
+Node Template can be imported using the Rancher Node Template ID
+
+```bash
+terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt.&lt;driver&gt;
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+* `cloud_credential_id` - (Required) The ID of the Cloud Credential (string)
+* `driver` - (Required) The driver of the Cloud Credential (string)
+
+Supported drivers:
+
+* amazonec2
+* azure
+* digitalocean
+* googlekubernetesengine
+* linode
+* openstack
+* s3
+* vmwarevsphere
--- a/rancher2/import_rancher2_cloud_credentials.go
+++ b/rancher2/import_rancher2_cloud_credentials.go
@@ -6,33 +6,23 @@ import (
 )
 
 func resourceRancher2CloudCredentialsImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	cloudCredentialID, driver := splitID(d.Id())
+	d.Set("driver", driver)
+
 	client, err := meta.(*Config).ManagementClient()
 	if err != nil {
 		return []*schema.ResourceData{}, err
 	}
 
 	cloudCredential := &CloudCredential{}
-	err = client.APIBaseClient.ByID(managementClient.CloudCredentialType, d.Id(), cloudCredential)
+	err = client.APIBaseClient.ByID(managementClient.CloudCredentialType, cloudCredentialID, cloudCredential)
 	if err != nil {
 		return []*schema.ResourceData{}, err
 	}
 
-	drivers := []string{
-		amazonec2ConfigDriver,
-		azureConfigDriver,
-		digitaloceanConfigDriver,
-		googleConfigDriver,
-		s3ConfigDriver,
-		vmwarevsphereConfigDriver,
-	}
-
-	// Missing "driver" field in api.
-	for _, driver := range drivers {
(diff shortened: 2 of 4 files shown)


rancher2/import_rancher2_cloud_credentials.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	// Missing "driver" field in api.
+	for _, driver := range drivers {
+		d.Set("driver", driver)
+		err = flattenCloudCredential(d, cloudCredential)
+		if err != nil {
+			return []*schema.ResourceData{}, err
+		}
+	}
```
rawmind0, line 26: How is this checking that the used `driver` is the proper one?? With this code, the for would always be executed by all `drivers`, isn't it?? 
On the import command the driver should be specified with the id and split  within the `resourceRancher2CloudCredentialsImport` function

rawmind0 (changes requested):
@BobVanB , thanks for the PR.

Cloud credential sensitive data can't be read from Rancher API. For example, for an `amazonec2`  cloud credential driver, `amazonec2_credential_config.secret_key` will not be available. That's the main reason due to import was not implemented for this resource. 

If this security limitation is still ok to you, please, take a look to changes requested. Also, resource documentation should be updated with the import command definition.

rancher2/import_rancher2_cloud_credentials.go
BobVanB: Hi @rawmind0,

Copy Paste

https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/import_rancher2_node_template.go#L20-L22

> Cloud credential sensitive data can't be read from Rancher API.

This doesn't matter, this is something you have to fix after the import anyway.
The main goal , for me at least, is to get the id imported. Because i can't change the statefile.

> How is this checking that the used driver is the proper one??

For other resources i can see there is a `driver: ...` field in the api of rancher.
Cloud Credential does not have this, this is why i just use a loop with all supported resources.
The better solution would be to add `driver` to the rancher api this way the code can target only one resource with `get('driver')`

That said, the empty resources return nothing and only the resource configured will be added.
Must be a hidden feature ;-)

> On the import command the driver should be specified with the id and split within the resourceRancher2CloudCredentialsImport function

I haven't seen any use of this before, can you point me to some examples?

rawmind0: > This doesn't matter, this is something you have to fix after the import anyway.
The main goal , for me at least, is to get the id imported. Because i can't change the statefile.

What do you mean with `...this is something you have to fix after the import anyway...` ?? The cloud credential should be added to the tf file before import and once imported, shouldn't generate diffs on apply.

> For other resources i can see there is a driver: ... field in the api of rancher.
Cloud Credential does not have this, this is why i just use a loop with all supported resources.
The better solution would be to add driver to the rancher api this way the code can target only one resource with get('driver')
> 
> That said, the empty resources return nothing and only the resource configured will be added.
Must be a hidden feature ;-)

May be i'm loosing something, but i don't get you point on this. The code is assigning the driver value to the resource data in every iteration, https://github.com/rancher/terraform-provider-rancher2/pull/847/files#diff-0308de0f04c805a9ddb5d67f5c1fe0bc5092e1af4ff918d0fd60136c8f4f412aR31 What is the final value of `driver` resourceData?? 

> I haven't seen any use of this before, can you point me to some examples?

Sure, you can take a look to [import_rancher2_app_v2](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/import_rancher2_app_v2.go#L7) . The import id is composed by `<CLUSTER_ID>.<APP_V2_NAME>` https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/app_v2#import . Here can be done something similar, `<CLOUD_CREDENTIAL_ID>.<DRIVER>` and forget the loop.

BobVanB: > What do you mean with ...this is something you have to fix after the import anyway... ?? The cloud credential should be added to the tf file before import and once imported, shouldn't generate diffs on apply.

https://spacelift.io/blog/importing-exisiting-infrastructure-into-terraform
We wanted to migrate from our own ansible modules to terraform.

*  We have an existing cluster, with filled resources.
* We want to create a terraform state that has no changes.
* I want to import the existing id into the terraform state.
* And then work away all the changes.
* I have the credentials, i don't need them from rancher.

BobVanB: > What is the final value of driver resourceData??

There is no final value for the driver. That resource don't have a driver.
That will be determent with the next `terraform plan/apply`.
This is just importing anything that will be found. And rancher does only allow one resource.
This will, for now, will always go correctly :D

BobVanB: > `<CLOUD_CREDENTIAL_ID>.<DRIVER>` and forget the loop.

Done

docs/resources/cloud_credential.md
```diff
@@ -120,3 +120,29 @@ The following attributes are exported:
 - `create` - (Default `10 minutes`) Used for creating cloud credentials.
 - `update` - (Default `10 minutes`) Used for cloud credential modifications.
 - `delete` - (Default `10 minutes`) Used for deleting cloud credentials.
+
+## Import
+
+Node Template can be imported using the Rancher Node Template ID
```
rawmind0: Typo here, not referencing node template

docs/resources/cloud_credential.md
```diff
@@ -120,3 +120,29 @@ The following attributes are exported:
...
+terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt;.&lt;driver&gt;
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+* `cloud_credential_id` - (Required) The ID of the Cloud Credential (string)
+* `driver` - (Required) The driver of the Cloud Credential (string)
+
```
rawmind0: These are not really terraform arguments, moreover import parameters (all required). Could you please use a format like https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/app_v2#import , including the parameters at the import definition??

docs/resources/cloud_credential.md
```diff
@@ -120,3 +120,29 @@ The following attributes are exported:
 - `create` - (Default `10 minutes`) Used for creating cloud credentials.
 - `update` - (Default `10 minutes`) Used for cloud credential modifications.
 - `delete` - (Default `10 minutes`) Used for deleting cloud credentials.
+
+## Import
+
+Node Template can be imported using the Rancher Node Template ID
+
+```bash
+terraform import rancher2_cloud_credential.foo &lt;cloud_credential_id&gt;.&lt;driver&gt;
```
rawmind0: For docs notation consistency, could you please set the variable names is capital letters??

docs/resources/cloud_credential.md
```diff
@@ -120,3 +120,29 @@ The following attributes are exported:
...
+Supported drivers:
+
+* amazonec2
+* azure
+* digitalocean
+* googlekubernetesengine
+* linode
+* openstack
+* s3
+* vmwarevsphere
```
rawmind0, line 141: This is good to me as defining posible values for driver.

rawmind0 (changes requested):
Thanks for the update @BobVanB 

Please, take a look to the requested changes

docs/resources/cloud_credential.md
BobVanB: I Removed the arguments part and just created a list.

rawmind0 (approved):
LGTM

PR #864, opened 2022-02-17
Add retry on conflict when deleting custom roles

mouellet:
This PR adds a Retry on deletion with retryable error on `IsConflict` API error, the provider will error after delete timeout.

When deleting custom roles that inherit other custom roles, you can *sometimes* get an 409 API error even if terraform 
deletes the hierarchy properly (childs are deleted before parents).


kkaempf:
closing stale PRs.

PR #876, opened 2022-02-25
Fix error on rke2 cluster for Azure

### LINKED ISSUE #875: Error when creating rancher2_machine_config_v2 using Azure
Problem: 
When using `rancher2_cluster_v2` with `rke_config` that references a `rancher2_machine_config_v2` using `azure_config` an error occurs, also when using an empty machine config.

Error:
`Error: Invalid address to set: []string{"azure_config", "0", "private_address_only"}`

Adding the parameter `private_address_only` also gives an error, because the parameter is not expected in the `rancher2_machine_config_v2` `azure_config` nested block.

02bensch:
This addresses and fixes rancher/terraform-provider-rancher2#875 
After adding the `private_address_only` to the schema, we were able to successfully create the resource.


phillipsj (approved):
I just ran all the tests and they are all passing.

superseb (changes requested):
I believe this needs a docs change at https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/machine_config_v2.md?

02bensch:
@superseb I have added the docs change. 
There is however - I believe - some overlap between `no_public_ip` and `private_address_only` within the `azure_config` arguments. Perhaps in future `no_public_ip` could be removed, in order to align `azure_config` more closely with `amazonec2_config` arguments.

eliyamlevy:
@02bensch I agree that we can probably drop the `no_public_ip` flag. Seems like overlapping functionality. I'd open another issue for removing it so it's written down somewhere.

eliyamlevy (approved):

superseb (approved):

02bensch:
> @02bensch I agree that we can probably drop the `no_public_ip` flag. Seems like overlapping functionality. I'd open another issue for removing it so it's written down somewhere.

@eliyamlevy I have opened https://github.com/rancher/terraform-provider-rancher2/issues/956. I will link the necessary PRs to it when I get a chance.

eliyamlevy:
@02bensch sounds good. I've passed it along to the right people to get it approved before merging. We also have an internal issue for this but feel free to link yours as well.

Issue #835, opened 2022-01-03
Provider produced inconsistent final plan, produced an invalid new value for │ .rke_config[0].machine_pools[0].cloud_credential_secret_name: was cty.StringVal(""), but now │ cty.StringVal("cattle-global-data:cc-694ng").


andersjohansson2021:
When terraforming a RKE2 cluster i receive the following:
│ Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for rancher2_cluster_v2.test-cluster to include new values learned so far during apply, provider
│ "registry.terraform.io/rancher/rancher2" produced an invalid new value for
│ .rke_config[0].machine_pools[1].cloud_credential_secret_name: was cty.StringVal(""), but now
│ cty.StringVal("cattle-global-data:cc-trrz8").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

This happends in version 1.22.1 not in 1.21.0 of the provider.
/Anders.

SURE-5412
SURE-4866

frouzbeh:
Any update on this? I'm seeing the same issue.
```
│ When expanding the plan for module.cluster.rancher2_cluster_v2.cluster to include new values learned so far
│ during apply, provider "registry.terraform.io/rancher/rancher2" produced an invalid new value for
│ .rke_config[0].machine_pools[0].cloud_credential_secret_name: was cty.StringVal(""), but now
│ cty.StringVal("cattle-global-data:cc-f9hbf").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

andersjohansson2021:
@frouzbeh I had too use an older version as stated above. That solved the issue for me. But having said that this bug need to be adressed for the provider.

frouzbeh:
@andersjohansson2021 Thank you, yes I tested with older version and it works, but as I remember older version had another issue with kube-config generation function which is fixed in 1.22.2. I hope they fix it soon.

frouzbeh:
@rawmind0 Would you please take a look at this issue. We really would like to make this work.

brudnak:
+1 seeing the same issue

git-ival:
+1 here as well

PrakashFromBunnings:
Hello , Any workaround or fix this issue . I am stuck at this issue .

PrakashFromBunnings:
> @frouzbeh I had too use an older version as stated above. That solved the issue for me. But having said that this bug need to be adressed for the provider.

[...]

PR #878, opened 2022-03-08
remove cloud_credential_secret_name from rke machine pool

ceesios:
I believe that the cloud_credential_secret_name should not be in the machine pool config. I couldn't find any rancher documentation to support this. The only clue was the absence of the option in the rancher ui for the credential under the machine pool config. The terraform apply also provides a valid plan with this proposed change off course.

Nevertheless I would recommend someone with deeper understanding of the v2 clusters to review this change.

rancher/terraform-provider-rancher2#835


kkaempf:
closing stale PRs.

PR #881, opened 2022-03-11
fixing  desc on how to use kms_key eks_config_v2

leonardoalvesprates:
Fixing the description on how to use properly the use of the values for kms_key parameter on eks_config_v2 options.


leonardoalvesprates:
I'm trying to understand if some characters is being blocked

Issue #869, opened 2022-02-23
[FEATURE] rke2/k3s Rotate Certificates


jakefhyde:
Implement certificate rotation according to https://github.com/rancher/rancher/issues/35343

rancher-max:
Validated that using the terraform provider to provision a cluster, it is possible to rotate either individual certs or all certs. Multiple rotations on the same cluster are also successful. My steps in validating were:

1. Clone this repo, using master branch commit `b875ecde` to validate
2. I updated scripts/version.sh to pin the version to `9.99.9` for my testing but this is probably optional
3. Run `make build`. Since I'm on OSX, I also had to update line 9 in scripts/gobuild.sh to use a lowercase a: `declare -a OS_ARCH_ARG`
4. Ensure the resulting binary is able to be used as a provider. I followed some steps from https://www.infracloud.io/blogs/developing-terraform-custom-provider/#:~:text=Steps%20to%20copy%20provider%20executable%20to%20plugins%20directory. The key piece here is: `cp bin/terraform-provider-rancher2 ~/.terraform.d/plugins/registry.terraform.io/rancher/rancher2/9.99.9/darwin_amd64`
5. Build the terraform. See below for my files (shown as one big file here)
```
# Implicit Local Provider
terraform {
  required_providers {
    rancher2 = {
      source  = "rancher/rancher2"
      version = "9.99.9"
    }
  }
}

# Variables
variable "rancher_url" {
  type        = string
  description = "URL for Rancher."
}

variable "rancher_token_key" {
  type        = string
  description = "API bearer token for Rancher."
  sensitive   = true
}

# main.tf
provider "rancher2" {
  api_url   = var.rancher_url
  token_key = var.rancher_token_key
  insecure = true
}

# Create amazonec2 cloud credential
resource "rancher2_cloud_credential" "maxtf" {
  name = "maxtf"
  amazonec2_credential_config {
    access_key = "<aws key>"
    secret_key = "<aws secret>"
  }
}

# Create amazonec2 machine config v2
resource "rancher2_machine_config_v2" "maxtfec2config" {
  generate_name = "maxtfconfig"
  amazonec2_config {
    ami =  "<ami>"
    region = "<region>"
    subnet_id = "<subnet id>"
    vpc_id = "<vpc id>"
    zone = "a"
    security_group = ["<sg name>"]
  }
}

# Create a new rancher v2 amazonec2 RKE2 Cluster v2
[...]

PR #882, opened 2022-03-11
added support for rke2 cluster cert rotation

paynejacob:
rancher/terraform-provider-rancher2#869 

- Added support for v2 cluster certificate rotation.
- upgraded go version for rancher compatibility
- fixed issues with harvester causing ci failures


jakefhyde (approved):

rancher2/schema_cluster_v2_rke_config.go
```diff
@@ -122,6 +122,15 @@ func clusterV2RKEConfigFields() map[string]*schema.Schema {
 				Schema: clusterV2RKEConfigETCDFields(),
 			},
 		},
+		"rotate_certificates": {
```
a-blender, line 125: Does `rotate_certificates` have a default bool value in rancher?

rancher2/schema_cluster_v2_rke_config_certificate_rotation.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+//Types
+
+func clusterV2RKEConfigRotateCertificatesFields() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		"generation": {
```
a-blender, line 11: Same question as above with these fields: what would happen if a cert rotation is done via terraform, and generation isn't set? Or services isn't set? If that would cause a bug in rancher, setting default values would be a good idea here

rancher2/structure_cluster_v2_rke_config_certificate_rotation.go
```diff
@@ -0,0 +1,41 @@
...
+import rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+
+func flattenClusterV2RKEConfigRotateCertificates(in *rkev1.RotateCertificates) []interface{} {
+	if in == nil {
+		return nil
+	}
+
+	obj := make(map[string]interface{})
+
+	if in.Generation > 0 {
```
a-blender, line 12: Seems like, from `lines 12` and `16` generation and services don't have to be set for a cert rotation to be successful. Is that true?
Sorry for all the questions :)

rancher2/structure_cluster_v2_rke_config_certificate_rotation.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if len(in.Services) > 0 {
+		obj["services"] = toArrayInterfaceSorted(in.Services)
+	}
+
+	return []interface{}{obj}
+}
+
+func expandClusterV2RKEConfigRotateCertificates(p []interface{}) *rkev1.RotateCertificates {
+	if p == nil || len(p) == 0 || p[0] == nil {
```
a-blender, line 24: Thought for a second here you could condense this, but an array's zero value can be non-nil unlike slices

rancher2/schema_cluster_v2_rke_config_certificate_rotation.go
paynejacob: generation is type int64, it will default to 0.  If one is not specified in terraform it will also default to zero.  The generation will not have changed and rancher will take no action.

rancher2/schema_cluster_v2_rke_config.go
paynejacob: No the object is nilable.

rancher2/structure_cluster_v2_rke_config_certificate_rotation.go
paynejacob: This function is for parsing the response from the rancher api and transforming it to terraform.  Given the certificate rotation object is nilable, we may not have a value to transform to tf.

rancher2/schema_cluster_v2_rke_config_certificate_rotation.go
a-blender: Ah, unspecified generation still results in 0 in rancher. Looks good then

a-blender (approved):

PR #883, opened 2022-03-11
fixing desc for using kms_key eks_config_v2

leonardoalvesprates:
Fixing the description of kms_key on how to use properly the values for eks_config_v2


leonardoalvesprates:
fixing some issue during bulid

PR #886, opened 2022-03-11
Fix rancher2 desc kms key

leonardoalvesprates:
Fixing the description of kms_key on how to use properly the values for eks_config_v2


leonardoalvesprates:
Tyring to submit a fix on docs

Issue #873, opened 2022-02-24
[FEATURE] rke2/k3s Drain Before Delete


jakefhyde:
Implement rke2/k3s "Drain Before Delete" according to https://github.com/rancher/rancher/issues/35274

a-blender:
## Testing template

#### Root cause

New feature: add `DrainBeforeDelete` support to custom RKE2/K3s clusters with Terraform.
 
#### What was fixed, or what changes have occurred

- Added drain_before_delete field to rke2/k3s [machine pool schema](https://github.com/rancher/terraform-provider-rancher2/blob/b875ecdefccfdf65f0603df177a77f739bc76f20/rancher2/schema_cluster_v2_rke_config_machine_pool.go#L44)
- Added drain_before_delete to the rke2/k3s [machine pool](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_v2_rke_config_machine_pool.go)
- Updated tests

#### Areas or cases that should be tested

1. Create an RKE2 cluster via Terraform

How to do step 1

- Clone https://github.com/rancher/terraform-provider-rancher2
- Compile a local terraform binary using this script

```
# When you run a `make bin` in the terraform-provider-rancher2 repo, it
# downloads a binary into your local cache from the specified source. To
# tell terraform to use a local binary, you need to overwrite the one in
# the cache.

# Add the following to your terraform config file. This tells terraform 
# to use the local cache https://terraform.io/cli/config/config-file.

: '
terraform {
  required_providers {
    rancher2 = {
      source = "terraform.example.com/local/rancher2"
      version = "1.0.0"
    }
  }
}
'

# Compile the local binary
make bin

# Remove lock file
rm -f .terraform.lock.hcl

# Overwrite the cached binary with your local one
cp terraform-provider-rancher2 ~/.terraform.d/plugins/terraform.example.com/local/rancher2/1.0.0/darwin_amd64/terraform-provider-rancher2
```

- Create a Terraform config to provision an RKE2 downstream cluster in EC2 with 2 nodes (1 cp/etcd with `drain_before_delete` not set, 1 worker with` drain_before_delete = true`)

```
terraform {
  required_providers {
    rancher2 = {
      source = "terraform.example.com/local/rancher2"
      version = "1.0.0"
    }
  }
}

provider "rancher2" {
  api_url    = "<url of local Rancher server>"
  token_key = "<Rancher API bearer token>"
  insecure = true
}

[...]

PR #890, opened 2022-03-18
Add drain before delete to cluster_v2 resource

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/873 

 
## Problem
 

This is a new feature. A user can specify a `DrainBeforeDelete` flag in the Rancher UI when creating an RKE/K3s cluster. This feature adds a `drain_before_delete` field to the [rancher2_cluster_v2](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster_v2) machine pool configuration so the same option can be set via Terraform when you provision an RKE/K3s cluster. Not setting the field adds a CAPI annotation on the machine deployment that excludes nodes from being drained. Setting it to true removes the CAPI annotation and will drain nodes on deletion (except for etcd nodes).
 
## Solution
 

* Added drain_before_delete field to rke2/k3s [machine pool schema](https://github.com/rancher/terraform-provider-rancher2/blob/b875ecdefccfdf65f0603df177a77f739bc76f20/rancher2/schema_cluster_v2_rke_config_machine_pool.go#L44)
* Added drain_before_delete to the rke2/k3s [machine pool](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_v2_rke_config_machine_pool.go)
* Updated tests

## Testing
 

* Create an RKE2 cluster via Terraform with 2 downstream nodes (1 cp/etcd with `drain_before_delete` not set, 1 worker with `drain_before_delete = true`)
* Create a K3s cluster with the same spec
* View each MachineDeployment in the Rancher UI. The MachineDeployment for cp/etcd nodes should have the CAPI annotation `machine.cluster.x-k8s.io/exclude-node-draining`set in the deployment spec. The MachineDeployment for the worker nodes should not have that annotation.

### DIFF
--- a/rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
+++ b/rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
@@ -45,11 +45,12 @@ func init() {
 	quantity := int32(10)
 	testClusterV2RKEConfigMachinePoolsConf = []provisionv1.RKEMachinePool{
 		{
-			Name:             "test",
-			DisplayName:      "test",
-			NodeConfig:       testClusterV2RKEConfigMachinePoolMachineConfigConf,
-			ControlPlaneRole: true,
-			EtcdRole:         true,
+			Name:              "test",
+			DisplayName:       "test",
+			NodeConfig:        testClusterV2RKEConfigMachinePoolMachineConfigConf,
+			ControlPlaneRole:  true,
+			EtcdRole:          true,
+			DrainBeforeDelete: true,
 			MachineDeploymentAnnotations: map[string]string{
 				"anno_one": "one",
 				"anno_two": "two",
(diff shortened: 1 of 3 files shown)


rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
```diff
@@ -79,6 +80,7 @@ func init() {
 			"machine_config":               testClusterV2RKEConfigMachinePoolMachineConfigInterface,
 			"control_plane_role":           true,
 			"etcd_role":                    true,
+			"drain_before_delete":			true,
```
jakefhyde: formatting, otherwise LGTM

jakefhyde (changes requested):

jakefhyde (approved):

paynejacob (approved):

thedadams (approved):

braunsonm:
Looks like @annablender forgot to update the documentation. This is now an undocumented argument and doesn't appear on registry.terraform.io

snasovich:
@braunsonm , looks like this was addressed in https://github.com/rancher/terraform-provider-rancher2/pull/903/files#diff-77c882c174e2591df55fb567b9198db2188df2dd80d62a0d2038c9a0e7c1a5f9R464, should be updated in published documentation with the next release.

Issue #824, opened 2021-12-14
Can't provision a downstream RKE2 node driver cluster with a standard user token


sgapanovich:
### Environment: ###
Rancher version: 2.6-head (confirmed it exists in 2.6.2 as well)
Rancher cluster type: HA
Docker version: 20.10

### Steps to reproduce: ###
1. Create a bearer token as a standard user
2. Use that token as `token_key` in the `provider "rancher2"` file
3. Create necessary files from docs https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster_v2#creating-rancher-v2-amazonec2-cluster-v2
4. run `terraform apply`

### Issue: ###
when a standard user token used I get an error back
```
rancher2_machine_config_v2.rke2conf: Creating...
╷
│ Error: [ERROR] waiting for machine config (fleet-default/nc-rke2conf-pjmdf) to be active: Getting Machine Config V2: Resource type [rke-machine-config.cattle.io.amazonec2config] can not be looked up by ID
│
│   with rancher2_machine_config_v2.rke2conf,
│   on rancher2_machine_config_v2_ec2.tf line 2, in resource "rancher2_machine_config_v2" "rke2conf":
│    2: resource "rancher2_machine_config_v2" "rke2conf" {
│
╵
```
when I pass an admin token a RKE2 cluster is created 

![user](https://user-images.githubusercontent.com/40282141/146242888-efb2f308-ac9c-4f1f-8048-72d09e48f63c.jpeg)

NOTE:
sometimes there is no error and `rancher2_machine_config_v2` never finishes to create a resource 

I can create a RKE downstream cluster using `rancher2_cluster` resource with a standard user token so I would expect to be able to create a RKE2 downstream cluster using `rancher2_cluster_v2` resource with a standard user token as well

rawmind0:
Is the same RKE2 cluster deployment with the same standard user working from the Rancher UI?? Have you tested using  the `clusters-create` global role instead??

sgapanovich:
Standard user can create a RKE2 cluster on UI but can't do it using terraform. No machines can't be created using a standard user token (even with `clusters-create` role added). It takes 10 min to provision but never finishes it

![machine](https://user-images.githubusercontent.com/40282141/146244233-62d39256-5a42-4021-905f-6f2f3359ebbb.jpg)


If an admin token is used you can see that it took 1 sec to create a machine 
[...]

PR #898, opened 2022-04-05
Fix RKE2 EC2 cluster creation with standard user token

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/824 
 
 
## Problem
 

The problem was that Terraform would throw an error for RKE2 clusters provisioned with a standard user token. The cluster was being prevented from becoming active on refresh if the terraform provider couldn't look up the machine config by ID. The root cause is that that RBAC is not configured to allow standard users access to the machine configs in `rke-machine-config.cattle.io.` Instead, the [authprovisioning controller](https://github.com/rancher/rancher/blob/ac92c2a38a122dddd9a859259775d31602b98eb0/pkg/controllers/management/authprovisioningv2/machineconfigs.go#L22) in Rancher gives a user access to the specific machine config that it creates, after provisioning. This caused Terraform to error out because it would try to create the cluster before Rancher set the machine config permissions.
 
## Solution
 

Add code in the Terraform provider to retry 5 times if the machine config state returns an error; specifically, restart RBAC if the machine config can't be looked up by ID.

Here
https://github.com/annablender/terraform-provider-rancher2/blob/50446e14852764b47192b745aeb187636c95412c/rancher2/resource_rancher2_machine_config_v2.go#L51

## Testing
 

Verify these scenarios work _the first time_ without error
* Provision an RKE2 EC2 cluster via terraform using an admin token
* Provision an RKE2 EC2 cluster via terraform using a standard user token (click the `Create Clusters` box when creating the user. It is not added by default)

### DIFF
--- a/rancher2/resource_rancher2_cluster_v2.go
+++ b/rancher2/resource_rancher2_cluster_v2.go
@@ -96,7 +96,7 @@ func resourceRancher2ClusterV2Read(d *schema.ResourceData, meta interface{}) err
 
 	cluster, err := getClusterV2ByID(meta.(*Config), d.Id())
 	if err != nil {
-		if IsNotFound(err) || IsForbidden(err) || IsNotLookForByID(err) {
+		if IsNotFound(err) || IsForbidden(err) || IsNotAccessibleByID(err) {
 			log.Printf("[INFO] Cluster V2 %s not found", d.Id())
 			d.SetId("")
 			return nil
@@ -165,7 +165,7 @@ func clusterV2StateRefreshFunc(meta interface{}, objID string) resource.StateRef
 	return func() (interface{}, string, error) {
 		obj, err := getClusterV2ByID(meta.(*Config), objID)
 		if err != nil {
-			if IsNotFound(err) || IsForbidden(err) || IsNotLookForByID(err) {
+			if IsNotFound(err) || IsForbidden(err) || IsNotAccessibleByID(err) {
 				return obj, "removed", nil
 			}
 			return nil, "", err
@@ -273,10 +273,10 @@ func waitForClusterV2State(c *Config, id, state string, interval time.Duration) 
 		obj, err := getClusterV2ByID(c, id)
 		if err != nil {
 			log.Printf("[DEBUG] Retrying on error Refreshing Cluster V2 %s: %v", id, err)
-			if !IsNotFound(err) && !IsForbidden(err) && !IsNotLookForByID(err) {
+			if !IsNotFound(err) && !IsForbidden(err) && !IsNotAccessibleByID(err) {
 				return nil, fmt.Errorf("Getting cluster V2 ID (%s): %v", id, err)
 			}
-			if IsNotLookForByID(err) {
+			if IsNotAccessibleByID(err) {
 				// Restarting clients to update RBAC
 				c.RestartClients()
 			}
--- a/rancher2/resource_rancher2_machine_config_v2.go
+++ b/rancher2/resource_rancher2_machine_config_v2.go
@@ -1,6 +1,7 @@
 package rancher2
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"time"
@@ -18,7 +19,7 @@ func resourceRancher2MachineConfigV2() *schema.Resource {
 		Delete: resourceRancher2MachineConfigV2Delete,
 		Schema: machineConfigV2Fields(),
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(10 * time.Minute),
+			Create: schema.DefaultTimeout(3 * time.Minute),
 			Update: schema.DefaultTimeout(10 * time.Minute),
 			Delete: schema.DefaultTimeout(10 * time.Minute),
 		},
@@ -38,20 +39,42 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
 
 	d.SetId(newObj.ID)
 	d.Set("kind", newObj.TypeMeta.Kind)
-	stateConf := &resource.StateChangeConf{
-		Pending:    []string{},
-		Target:     []string{"active"},
-		Refresh:    machineConfigV2StateRefreshFunc(meta, newObj.ID, newObj.TypeMeta.Kind),
-		Timeout:    d.Timeout(schema.TimeoutCreate),
-		Delay:      1 * time.Second,
-		MinTimeout: 3 * time.Second,
-	}
-	_, waitErr := stateConf.WaitForState()
-	if waitErr != nil {
-		return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, waitErr)
+
+	err = waitForMachineConfigV2(d, meta.(*Config), d.Timeout(schema.TimeoutCreate))
+	if err != nil {
+		return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, err)
 	}
 
 	return flattenMachineConfigV2(d, newObj)
+}
+
+func waitForMachineConfigV2(d *schema.ResourceData, config *Config, interval time.Duration) error {
+	log.Printf("[INFO] Waiting for state Machine Config V2 %s", d.Id())
+
+	ctx, cancel := context.WithTimeout(context.Background(), interval)
+	defer cancel()
+	for {
+		kind := d.Get("kind").(string)
+		_, err := getMachineConfigV2ByID(config, d.Id(), kind)
+		if err == nil {
+			return nil
+		}
+		log.Printf("[INFO] Retrying on error Refreshing Machine Config V2 %s: %v", d.Id(), err)
+		if IsNotFound(err) || IsForbidden(err) {
+			d.SetId("")
+			return fmt.Errorf("Machine Config V2 %s not found: %s", d.Id(), err)
+		}
+		if IsNotAccessibleByID(err) {
+			// Restarting clients to update RBAC
+			config.RestartClients()
+		}
+
+		select {
+		case <-time.After(rancher2RetriesWait * time.Second):
+		case <-ctx.Done():
+			return fmt.Errorf("Timeout waiting for machine config V2 ID %s", d.Id())
+		}
+	}
 }
 
 func resourceRancher2MachineConfigV2Read(d *schema.ResourceData, meta interface{}) error {
@@ -60,7 +83,7 @@ func resourceRancher2MachineConfigV2Read(d *schema.ResourceData, meta interface{
 	kind := d.Get("kind").(string)
 	obj, err := getMachineConfigV2ByID(meta.(*Config), d.Id(), kind)
 	if err != nil {
-		if IsNotFound(err) || IsForbidden(err) || IsNotLookForByID(err) {
+		if IsNotFound(err) || IsForbidden(err) || IsNotAccessibleByID(err) {
 			log.Printf("[INFO] Machine Config V2 %s not found", d.Id())
 			d.SetId("")
 			return nil
@@ -102,7 +125,7 @@ func resourceRancher2MachineConfigV2Delete(d *schema.ResourceData, meta interfac
 
 	obj, err := getMachineConfigV2ByID(meta.(*Config), d.Id(), kind)
 	if err != nil {
-		if IsNotFound(err) || IsForbidden(err) || IsNotLookForByID(err) {
+		if IsNotFound(err) || IsForbidden(err) || IsNotAccessibleByID(err) {
 			d.SetId("")
 			return nil
 		}
@@ -137,7 +160,7 @@ func machineConfigV2StateRefreshFunc(meta interface{}, objID, kind string) resou
 				return obj, "removed", nil
 			}
 			// This is required to allow standard user to use this resource
-			if !IsNotLookForByID(err) {
+			if !IsNotAccessibleByID(err) {
 				return obj, "active", nil
 			}
 			return nil, "", err
--- a/rancher2/util.go
+++ b/rancher2/util.go
@@ -341,7 +341,7 @@ func IsUnknownSchemaType(err error) bool {
 	return strings.Contains(err.Error(), "Unknown schema type")
 }
 
-func IsNotLookForByID(err error) bool {
+func IsNotAccessibleByID(err error) bool {
 	return strings.Contains(err.Error(), "can not be looked up by ID")
 }
 


rancher2/resource_rancher2_cluster_v2.go
```diff
@@ -165,9 +165,13 @@ func clusterV2StateRefreshFunc(meta interface{}, objID string) resource.StateRef
 	return func() (interface{}, string, error) {
 		obj, err := getClusterV2ByID(meta.(*Config), objID)
 		if err != nil {
-			if IsNotFound(err) || IsForbidden(err) || IsNotLookForByID(err) {
+			if IsNotFound(err) || IsForbidden(err) {
 				return obj, "removed", nil
 			}
+			// This is required to allow standard user to use this resource
+			if IsNotAccessibleByID(err) {
```
thedadams: If a user cannot access something by ID, then they should not be given access to that item. This seems like privilege escalation.

I think it should be determined why the user doesn't have access to it and fix the permissions.

thedadams (changes requested):

rancher2/resource_rancher2_cluster_v2.go
a-blender: Ah I see - I thought Terraform should provision anyway because a standard user not being able to access the machine config was the bug. I'll look into this a little more

a-blender: Here's the new PR https://github.com/rancher/rancher/pull/37230

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -47,8 +47,23 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
-		return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, waitErr)
+		if IsNotAccessibleByID(waitErr) {
+			var numberOfRetries = 0
+			for numberOfRetries < 3 {
+				_, err = stateConf.WaitForState()
+				if err != nil {
+					numberOfRetries++
+					continue
+				}
+				return flattenMachineConfigV2(d, newObj)
```
jakefhyde:
```suggestion
			for i := 0; i < 3; i++ {
				_, err = stateConf.WaitForState()
				if err != nil {
					continue
				}
				return flattenMachineConfigV2(d, newObj)
```
I feel like we can pull the loop counter into the loop scope and give it a simpler name, otherwise logic looks good.

jakefhyde (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -47,8 +47,21 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+		if IsNotAccessibleByID(waitErr) {
+			for i := 0; i < 3; i++ {
+				_, err = stateConf.WaitForState()
+				if err != nil {
+					continue
+				}
+				return flattenMachineConfigV2(d, newObj)
+			}
+		}
+		return fmt.Errorf("[ERROR] Waiting for machine config (%s) to be active: %s", newObj.ID, waitErr)
```
jakefhyde: Can we just replace `waitErr` with `err`? Sorry should've caught this first pass, just worried about the case where neither `waitErr` nor `err` were nil but for potentially different reasons.

jakefhyde (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -47,8 +47,21 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
 		MinTimeout: 3 * time.Second,
 	}
 	_, waitErr := stateConf.WaitForState()
```
thedadams: Can we get rid of this call and just do the retry? Maybe increase the retry to 4 and move this into the for-loop.

thedadams (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
a-blender: I'm going to condense the logic so I'll just have one `err` within the for loop. Pending

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -46,9 +46,20 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+	for i := 0; i < 4; i++ {
+		_, err = stateConf.WaitForState()
+		if err != nil {
+			if IsNotAccessibleByID(err) {
+				log.Printf("[INFO] Retrying state for Machine Config V2 %s", newObj.ID)
+				continue
+			}
+			return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, err)
+		}
+		return flattenMachineConfigV2(d, newObj)
```
jakefhyde: This logic doesn't seem right to me, should we still exit the loop with `IsNotAccesibleByID(err)` still being true after 4 tries?

jakefhyde (dismissed):

rancher2/resource_rancher2_machine_config_v2.go
a-blender: From what I've seen the machine config v2 state reconciles after 1 try in terraform, so doing 4 is just for good measure but yeah, ideally we'd want to return an err saying if the machine config _still_ isn't accessible then return an error from the create function. Thought?

jakefhyde: Yeah I think we should do that, because we could potentially have a problem with the controller that is supposed to give the user access to that machine config, and in that case we would want to return the error even if we can never error after 4 tries in the happy path.

a-blender: Ah ok makes sense. I'll add an extra check

a-blender: Added! Thanks for me catching me on this one @jakefhyde

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -46,9 +46,28 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+	// Terraform may finish creating the machine config v2 before rancher has set get
+	// permissions on the resource. Retry a few times to see if the state will return
+	// without error.
+	var machineConfigErrorState bool
+
+	for i := 0; i < 4; i++ {
+		_, err = stateConf.WaitForState()
+		if err != nil {
+			machineConfigErrorState = true
+			if IsNotAccessibleByID(err) {
```
thedadams: Currently, if this returns `false`, then we won't retry because we will return on line 64. Is this the expect flow?

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -46,9 +46,28 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+		return flattenMachineConfigV2(d, newObj)
+	}
+
+	// If we retry four times and the machine config v2 is still inaccessible or in an
+	// error state, return err.
+	if machineConfigErrorState {
+		return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, err)
 	}
 
 	return flattenMachineConfigV2(d, newObj)
```
thedadams: This line is unreachable given the current logic.

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -46,9 +46,28 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+				log.Printf("[INFO] Retrying state for Machine Config V2 %s", newObj.ID)
+				continue
+			}
+		}
+		return flattenMachineConfigV2(d, newObj)
+	}
+
+	// If we retry four times and the machine config v2 is still inaccessible or in an
+	// error state, return err.
+	if machineConfigErrorState {
```
thedadams: Currently, this check is not needed because, if we reach this point, then `machineConfigErrorState` is `true`.

thedadams (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
a-blender: Last line is now reachable!

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -38,29 +39,53 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), interval)
+	defer cancel()
+	for {
+		kind := d.Get("kind").(string)
+		_, err := getMachineConfigV2ByID(meta.(*Config), d.Id(), kind)
+		if err == nil {
+			return nil
+		}
+		if err != nil {
```
thedadams: At this point, we already know that the error is not nil.

thedadams (approved):
One nit, otherwise looks good.

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -38,29 +39,51 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+	for {
+		kind := d.Get("kind").(string)
+		_, err := getMachineConfigV2ByID(meta.(*Config), d.Id(), kind)
+		if err == nil {
+			return nil
+		}
+		log.Printf("[DEBUG] Retrying on error Refreshing Machine Config V2 %s: %v", d.Id(), err)
+		if IsNotFound(err) || IsForbidden(err) {
+			d.SetId("")
+			return fmt.Errorf("[INFO] Machine Config V2 %s not found: %s", d.Id(), err)
```
thedadams: An error prefixed with `[INFO]` is likely a copy-pasta leftover.

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -38,29 +39,51 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
-		Refresh:    machineConfigV2StateRefreshFunc(meta, newObj.ID, newObj.TypeMeta.Kind),
-		Timeout:    d.Timeout(schema.TimeoutCreate),
-		Delay:      1 * time.Second,
-		MinTimeout: 3 * time.Second,
-	}
-	_, waitErr := stateConf.WaitForState()
-	if waitErr != nil {
-		return fmt.Errorf("[ERROR] waiting for machine config (%s) to be active: %s", newObj.ID, waitErr)
+
+	err = waitForMachineConfigV2State(meta.(*Config), d, meta, d.Timeout(schema.TimeoutCreate))
```
thedadams: If I understand correctly, this create timeout is 20 or 30 minutes. That is appropriate for clusters, but this timeout should likely be shorter: 30-60 seconds.

Additionally, the name of this function doesn't explain what it does. It implies that there is some state the machine config is going to transition to.

thedadams (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
a-blender: `[INFO]` errs are commonly listed throughout the terraform provider. @Oats87 pointed out to me that the log message on line 62 should actually be an INFO message

What's the ask here?

rancher2/resource_rancher2_machine_config_v2.go
a-blender: I'm using the machine config default create timeout https://github.com/annablender/terraform-provider-rancher2/blob/ca13e143dedd568ef010339bc611041aaedd04aa/rancher2/resource_rancher2_machine_config_v2.go#L21, which is 10 mins. Is changing it to 1 min ok?

The func name is cohesive with the wait func for the cluster_v2 resource https://github.com/annablender/terraform-provider-rancher2/blob/ca13e143dedd568ef010339bc611041aaedd04aa/rancher2/resource_rancher2_cluster_v2.go#L265. If I change one, would I have to change both? I guess maybe `waitForValidMachineConfigV2State` would be more specific.

thedadams: I think 10 minutes is too long. There is nothing that has to happen when the machine config is created besides the RBAC getting applied. I think 1 to 3 minutes is plenty.

The cluster function takes in a `state` parameter, so the name makes sense there. The machine config one does not. Also, you are passing the same parameter twice in your implementation. Is it possible to only pass `meta` in once?

rancher2/resource_rancher2_machine_config_v2.go
thedadams: This is the only place I am seeing in the terraform provider where an error is returned with a prefix of `[INFO]` in its message. It doesn't seem correct that an error should have such a prefix.

The ask is to remove the prefix of `[INFO]` on this error.

thedadams (changes requested):
My previous requests were not addressed.

rancher2/resource_rancher2_machine_config_v2.go
a-blender: Ah yes, that makes more sense. Will do

rancher2/resource_rancher2_machine_config_v2.go
a-blender: Right, and with this implementation the machine config restarts RBAC once within 20 secs and terraform succeeds in getting the object. We don't want to underestimate the timeout though, so 3 mins sounds fair. For the name, how about `waitForMachineConfigV2` or `getMachineConfigV2?` Yeah, thanks for catching that. I only have to pass in `meta` and can grab `meta.(*Config)` inside the func.

thedadams (approved):

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -38,29 +39,51 @@ func resourceRancher2MachineConfigV2Create(d *schema.ResourceData, meta interfac
...
+			return nil
+		}
+		log.Printf("[INFO] Retrying on error Refreshing Machine Config V2 %s: %v", d.Id(), err)
+		if IsNotFound(err) || IsForbidden(err) {
+			d.SetId("")
+			return fmt.Errorf("Machine Config V2 %s not found: %s", d.Id(), err)
+		}
+		if IsNotAccessibleByID(err) {
+			// Restarting clients to update RBAC
+			config.RestartClients()
```
jiaqiluo, line 69: I do not understand why we need to restart the client to update RBAC. Is it because we cannot refresh the cache for RBAC or something in this case?  Is it the way Terraform works that I am not aware of? 
Excuse me if my questions do not make sense.

jiaqiluo (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
a-blender: In this case, the bug is a race condition. Terraform finishes creating the machine config object and starts trying to create the cluster _before_ Rancher has given GET method access to the machine config object for a standard user. RBAC is the service that checks those permissions and returns whether or not access is allowed (high level-y), so restarting RBAC here refreshes the service and enables RBAC to see that Rancher has given access to that obj. Then terraform will provision correctly.

Does that make sense? This is basically the root of my fix.

a-blender: This same thing is also done when terraform tries to create the cluster v2 https://github.com/rancher/terraform-provider-rancher2/blob/ca13e143dedd568ef010339bc611041aaedd04aa/rancher2/resource_rancher2_cluster_v2.go#L281

jiaqiluo (approved):
LGTM

PR #899, opened 2022-04-11
Bump Rancher and Docker for acceptance tests

superseb:


superseb:
This needs a bigger overhaul

Issue #902, opened 2022-04-11
[FEATURE] rke2/k3s Drain Before Delete Node Drain Timeout


jakefhyde:
Follow up to https://github.com/rancher/terraform-provider-rancher2/issues/873

Implement `NodeDrainTimeout` as added in https://github.com/rancher/rancher/pull/36835 for maximum drain wait before force deleting. Additionally, updates documentation here https://github.com/rancher/terraform-provider-rancher2/blob/9669e77f0b5e6a2c2f08157eb8cb2b21bd88c5c1/docs/resources/cluster_v2.md?plain=1#L459-L468 for both `drainBeforeDelete` and `nodeDrainTimeout`.

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type: Terraform
----------

### Testing: ###
Tested this issue by adding `node_drain_timeout` in terraform config file.

**Result** - Pass
Terraform accepts node drain timeout parameter.
- rke2
  - node_drain_timeout set to 1 - node gets deleted before drain is complete
  - node_drain_timeout set to 0 - node gets drained successfully
- k3s
  - node_drain_timeout set to 1 - node gets deleted before drain is complete
  - node_drain_timeout set to 0 - node gets drained successfully

PR #903, opened 2022-04-12
Add Drain Before Delete Timeout to cluster_v2 resource

HarrisonWAffel:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/902, https://github.com/rancher/terraform-provider-rancher2/issues/873

Problem:
A user can specify a `Drain Timeout` if they select `Drain Nodes` in the Rancher UI. This PR add the `node_drain_timeout` to the rancher2_cluster_v2 machine pool configuration so that this property can be set via Terraform. Supplying a value for this field denotes the number of seconds Rancher should wait for nodes to drain before deleting them. `0` indicates that it will wait indefinitely.

Solution:
 + add `node_drain_timeout` field to machine pool schema
 + add `node_drain_timeout` to the machine pool
 + update tests
 + update go.mod


go.mod
```diff
@@ -23,7 +23,7 @@ replace (
 	github.com/crewjam/saml => github.com/crewjam/saml v0.4.1
 	github.com/docker/distribution => github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker => github.com/docker/docker v20.10.6+incompatible
-	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20220309231411-e4af2465c5b4
+	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20220411181203-b44d54ae2e81
 	github.com/rancher/rancher/pkg/client => github.com/rancher/rancher/pkg/client v0.0.0-20220309231411-e4af2465c5b4
```
jakefhyde: We should probably update the client here as well

rancher2/schema_cluster_v2_rke_config_machine_pool.go
```diff
@@ -78,6 +78,12 @@ func clusterV2RKEConfigMachinePoolFields() map[string]*schema.Schema {
 			Default:     false,
 			Description: "Machine pool drain before delete",
 		},
+		"node_drain_timeout": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     0,
+			Description: "Machine pool drain timeout",
```
jakefhyde: We should add seconds to description, and include what it is used for in description

docs/resources/cluster_v2.md
```diff
@@ -461,6 +461,8 @@ The following attributes are exported:
 * `machine_config` - (Required) Machine pool node config (list)
 * `control_plane_role` - (Optional) Machine pool control plane role? (bool)
 * `etcd_role` - (Optional) Machine pool etcd role? (bool)
+* `drain_before_delete` - (Optional) Machine Pool Drain Before Delete? (bool)
+* `node_drain_timeout` - (Optional) Machine Pool Drain Timeout? (int)
```
jakefhyde: Can we add `node_startup_timeout_seconds`, `unhealthy_node_timeout_seconds`, `max_unhealthy`, and `unhealthy_range` as well? And also a more descriptive message for `node_drain_timeout` would be nice. Also I know it was messed up before, but if we could unify the indentation here that would be nice as well.

jakefhyde (changes requested):

docs/resources/cluster_v2.md
HarrisonWAffel: I can add the missing fields, but are we sure we want unified indentation? This would be the only place in the file where that is done, and I'm not sure if it effects any generated documentation.

go.mod
```diff
@@ -23,8 +23,8 @@ replace (
 	github.com/crewjam/saml => github.com/crewjam/saml v0.4.1
 	github.com/docker/distribution => github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker => github.com/docker/docker v20.10.6+incompatible
-	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20220309231411-e4af2465c5b4
-	github.com/rancher/rancher/pkg/client => github.com/rancher/rancher/pkg/client v0.0.0-20220309231411-e4af2465c5b4
+	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20220411181203-b44d54ae2e81
```
jakefhyde, line 26: I think these should probably be the same commit, otherwise looks good.

jakefhyde (changes requested):

jakefhyde (approved):

rancher2/schema_cluster_v2_rke_config_machine_pool.go
```diff
@@ -78,6 +78,12 @@ func clusterV2RKEConfigMachinePoolFields() map[string]*schema.Schema {
 			Default:     false,
 			Description: "Machine pool drain before delete",
 		},
+		"node_drain_timeout": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     0,
+			Description: "seconds to wait for machine pool drain to complete before machine deletion",
```
a-blender, line 85: Pedantic, but capitalize description ;)

a-blender (approved):
LGTM Harrison!

PR #910, opened 2022-04-24
Allow import on node template config drivers for all drivers that allow it without additional modification

staerion:
Why this pull request?

When importing node-template the config driver configuration are not included. For example an imported node-template with a vsphere diver configuration will generate a statefile like below:

```json
  "resources": [
    {
      "mode": "managed",
      "type": "rancher2_node_template",
      "name": "myname",
      "provider": "provider[\"registry.terraform.io/rancher/rancher2\"]",
      "instances": [
        {
          ...
         "attributes": {
            "use_internal_ip_address": true,
            "vsphere_config": []
          },
          "sensitive_attributes": [],
          "private": "myprivatekey"
        }
      ]
    }
  ]
```
Allowing the driver config to be set will correctly import the driver configuration as well:

```json
  "resources": [
    {
      "mode": "managed",
      "type": "rancher2_node_template",
      "name": "myname",
      "provider": "provider[\"registry.terraform.io/rancher/rancher2\"]",
      "instances": [
        {
          ...
          "attributes": {
            "use_internal_ip_address": true,
            "vsphere_config": [
              {
                "boot2docker_url": "",
                "cfgparam": [
                  "disk.enableUUID=TRUE"
                ],
                ...
                "vcenter": "",
                "vcenter_port": "443"
              }
            ]
          },
          "sensitive_attributes": [],
          "private": "myprivatekey"
        }
      ]
    }
  ]
```
Azure config driver has been excluded for now because it appears that are some attributes still configured in terraform which no longer exist in rancher. As a result the tests fail when running against rancher version 2.3.6+.

Openstack and opennebula have also be excluded as I can't find any reference of these driver configs in rancher version 2.3.6+.

I'm open to a different approach, but currently I suggest to solve the azure config driver in a separate pull request and leaving the openstack and opennebula config drivers out.

PR #911, opened 2022-04-24
Allow to include azure driver config when importing node-templates

staerion:
Why this pull request?

This pull request allows to include the azure driver config into the statefile when importing a node-template based on azure. The pull request is related to https://github.com/rancher/terraform-provider-rancher2/pull/910 in which the azure driver config has been skipped because of the complications outlined below:

To accomplish including the attribute azure_config in the terraform imports of node-templates, three extra changes need to happen:

1. PrivateAddressOnly has been removed as an attribute. This attribute is nowhere to be found in the rancher api for the azure_config resource schema and it breaks the test.
2. DockerPort was missing as an attribute for the azure_config, also breaking the test as it is created with a default. It has been added.
3. The default value "docker-machine-nsg" for the new nsg attribute has been removed. This is potentially breaking existing terraform configurations. Since this attribute does not exist in rancher 2.3.6 or 2.4.2 it is however also breaking diffs / updates to rancher clusters still running with these versions. Omit the attribute when empty and only push it to rancher when explicitly defined seemed like the best solution for this kind of attributes. However I'd be happy to hear about other solutions.

### DIFF
--- a/rancher2/schema_node_template_azure.go
+++ b/rancher2/schema_node_template_azure.go
@@ -11,34 +11,34 @@ const (
 //Types
 
 type azureConfig struct {
-	AvailabilitySet    string   `json:"availabilitySet,omitempty" yaml:"availabilitySet,omitempty"`
-	ClientID           string   `json:"clientId,omitempty" yaml:"clientId,omitempty"`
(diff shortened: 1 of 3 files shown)


kkaempf:
closing stale PRs.

Issue #872, opened 2022-02-24
[FEATURE] rke2/k3s Etcd Snapshot Restore


jakefhyde:
Implement etcd snapshot restore according to https://github.com/rancher/rancher/issues/34417

a-blender:
Not blocked by S3 snapshot or UI testing. Will implement this feature.

a-blender:
## Testing template

#### Root cause

Implement rke2/k3s etcd snapshot create and restore in terraform.
 
#### What was fixed, or what changes have occurred

Added the following functionality to rke2/k3s v2 provisioning

* etcd_snapshot_create (generation) - creates a snapshot of the cluster's etcd db
* etcd_snapshot_restore (name, generation, rke_restore_config) - restores the cluster to the snapshot specified in the name field
* Update terraform docs

So now, if you add etcd_snapshot_create or etcd_snapshot_restore to the rke_config field in a terraform config file, terraform will pass that RKE configuration to Rancher and it will be set on the cluster object spec. When Rancher checks for config updates, it will kick off/restore to a snapshot.

```
etcd_snapshot_create {
  generation = 2
}
```

```
etcd_snapshot_restore {
  name = "ablender-rke2-downstream-on-demand-ip-172-31-31-249.us-ea-dcf6b"
  generation = 1
}
```

#### Areas or cases that should be tested

Test steps

* Provision an rke2 3 node cluster in AWS EC2 via terraform on rainbow `2.6-head`
* Test `Snapshot Now` button to make sure that snapshots work. Use rke2 version `v1.22.9-rc4+rke2r1`. I did not see snapshots appear before this version.
* Add deployment/pod
* Add etcd snapshot create to terraform config, run terraform apply to create snapshot
* Remove deployment/pod
* Add etcd snapshot restore to terraform config, run terraform apply to restore to the previous snapshot. Verify that your data is restored.

Tf config

```
terraform {
  required_providers {
    rancher2 = {
      source = "rancher/rancher2"
      version = "1.23.0"
    }
  }
}

provider "rancher2" {
  api_url = "https://rainbow.cp-dev.rancher.space"
  token_key = "<Rancher API bearer token>"
  insecure = true
}

# Create amazonec2 machine config v2
resource "rancher2_machine_config_v2" "foo" {
  generate_name = "machine-config"
  amazonec2_config {
    ami =  "ami-020b9012dfce5d885"
    region = "us-east-2"
    security_group = []
[...]

PR #920, opened 2022-05-04
Add k3s/rke2 etcd snapshot restore support

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/872 

 
## Problem
 

Implement rke2/k3s etcd snapshot create and restore in terraform.

## Solution
 

Added the following functionality to rke2/k3s v2 provisioning
* `etcd_snapshot_create` (generation) - creates a snapshot of the cluster's etcd db
* `etcd_snapshot_restore` (name, generation, rke_restore_config) - restores the cluster to the snapshot specified in the `name` field
* Update terraform docs

So now, if you add `etcd_snapshot_create` or `etcd_snapshot_restore` to the `rke_config` field in a terraform config file, terraform will pass that RKE configuration to Rancher and it will be set on the cluster object spec. When Rancher checks for config updates, it will kick off/restore to a snapshot.

```
etcd_snapshot_create {
  generation = 2
}
```
 
```
etcd_snapshot_restore {
  name = "ablender-rke2-downstream-on-demand-ip-172-31-31-249.us-ea-dcf6b"
  generation = 1
}
```

## Testing
 

Test steps
* Tested on rainbow cluster `rancher 2.6-head`
* Provision an rke2 3 node cluster in AWS EC2 via terraform
* Test `Take Snapshot` Now button to make sure that snapshots work. Use rke2 version `v1.22.9-rc4+rke2r1.` I did not see snapshots appear before this version.
* Add deployment/pod
* Add etcd snapshot create to terraform config, run `terraform apply` to create snapshot
* Remove deployment/pod
* Add etcd snapshot restore to terraform config, run `terraform apply` to restore to the previous snapshot. Verify that your data is restored.

### DIFF
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -420,6 +420,9 @@ The following attributes are exported:
 * `machine_selector_config` - (Optional/Computed) Cluster V2 machine selector config (list)
 * `registries` - (Optional) Cluster V2 docker registries (list maxitems:1)
 * `etcd` - (Optional) Cluster V2 etcd (list maxitems:1)
+* `rotate_certificates` (Optional) Cluster V2 certificate rotation (list maxitems:1)
+* `etcd_snapshot_create` (Optional) Cluster V2 etcd snapshot create (list maxitems:1)
+* `etcd_snapshot_restore` (Optional) Cluster V2 etcd snapshot restore (list maxitems:1)
 
 #### `local_auth_endpoint`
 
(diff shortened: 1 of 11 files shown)


docs/resources/cluster_v2.md
```diff
@@ -563,6 +566,27 @@ The following attributes are exported:
 * `region` - (Optional) ETCD snapshot S3 region (string)
 * `skip_ssl_verify` - (Optional) Disable ETCD skip ssl verify. Default: `false` (bool)
 
+##### `rotate_certificates`
+
+###### Arguments
+
+* `generation` - (Optional) Desired certificate rotation generation (int)
```
jakefhyde: I just noticed that generation for cert rotation is optional, I wonder if it should be required instead, or if etcd related ones should be optional? cc @paynejacob

jakefhyde (changes requested):

docs/resources/cluster_v2.md
a-blender: So what @jakefhyde and I concluded on was that the top level rke_config fields (cert rotation, snapshot create, snapshot restore) should be Optional but the proper subfields should be Required because for ex: you don't want to request a restore with an empty name and generation. If the generation has to be set for cert rotation to work, I think it should be Required.

paynejacob: I think this makes sense and we can require it.

paynejacob (approved):

jakefhyde (approved):

PR #932, opened 2022-05-19
Add rancher2_custom_user_token resource

### LINKED ISSUE #315: Create Provider using Username/Password to be able to generate tokens for other users

sboschman:
See rancher/terraform-provider-rancher2#315 for different use-cases.

The original proposed solution by rawmind0 was to modify the rancher2_user resource to have it optionally return a token.  A preview of that solution is available in rancher/terraform-provider-rancher2#926.

As there is also the request for multiple tokens for a user, and a bunch of user accounts to manage from tf this alternative solution is up for discussion. Reasons are juggling multiple rancher2_user resources linked to user specific rancher2 provider instances (with alias) to generate additional tokens is complicated, and the current tf sdk does not support nested blocks with logical keys, making it hard/impossible to expose multiple tokens from the rancher2_user resource.

Hence a new resource, which takes a username and password to perform a login, to generate a token for an arbitrary user. The existing rancher2_token resource only generates tokens for the provider user.

This solution duplicates the need to provide the username/password from the `rancher2_user` to the new `rancher2_custom_user_account_token` resource and in doing so this new resource uses it's own credentials to communicate with the Rancher api, instead of the provider credentials. Even with these drawbacks I think this is the better of the two approaches (#926 vs rancher/terraform-provider-rancher2#932). Less complicated code and more flexibility to use tf constructs to generate many user / service acoounts, each with multiple tokens if desired.


Shocktrooper:
Thank you for doing this @sboschman let me know if you need anything as I have previous experience on working with development of providers

Shocktrooper:
@annablender Can we get a review for this so we can potentially add it to the v2.6.6 milestone?

JWackerbauer:
Bump, this feature would be very useful!

sboschman:
Is there anything I can do to get a review going @annablender or @HarrisonWAffel ? 

If there are different ideas on how to solve the issues described in #315 with Rancher + Terraform than please let us know.

docs/resources/service_account_token.md
```diff
@@ -0,0 +1,83 @@
...
+    rancher2_global_role_binding.foo-login
+  ]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `username` - (Required/ForceNew) The user username (string)
```
a-blender: Why are these args `ForceNew?` I actually haven't run into this before but as I understand it, you would use `ForceNew` if you didn't want/need to create an Update function for a resource. Is that the case here? I know tokens can't be updated

rancher2/resource_rancher2_service_account_token.go
```diff
@@ -0,0 +1,189 @@
...
+			log.Printf("[INFO] Token ID %s not found.", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	err = client.Token.Delete(token)
+	if err != nil {
+		return fmt.Errorf("Error removing Token: %s", err)
```
a-blender: Add `[ERROR]` to the front of str

rancher2/resource_rancher2_service_account_token_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	})
+}
+
+func testAccCheckRancher2ServiceAccountTokenExists(n string, token *managementClient.Token) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
```
a-blender: Add `[ERROR]` to all fmt.Errorf object strings returned from a function

rancher2/schema_service_account_token.go
```diff
@@ -0,0 +1,100 @@
...
+	s := map[string]*schema.Schema{
+		"password": {
+			Type:      schema.TypeString,
+			Required:  true,
+			Sensitive: true,
+		},
+		"username": {
+			Type:     schema.TypeString,
+			Required: true,
+			ForceNew: true,
```
a-blender: Same question about `ForceNew`

rancher2/schema_service_account_token.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+//Schemas
+
+func serviceAccountTokenFields() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		"password": {
```
a-blender: Add `Description: <description>` to password, username, temp token, and temp token id fields

rancher2/schema_service_account_token.go
```diff
@@ -0,0 +1,100 @@
...
+			Type:      schema.TypeString,
+			Required:  true,
+			Sensitive: true,
+		},
+		"username": {
+			Type:     schema.TypeString,
+			Required: true,
+			ForceNew: true,
+		},
+		"access_key": {
```
a-blender: I think `access_key` is a Sensitive field as well?

a-blender (changes requested):

a-blender:
In general, I agree with the approach and good job on the implementation and error handling. There may be security concerns around being able to create a new user with a custom username/password to access the Rancher API for a specific cluster id via terraform so I'd discuss this with my team before moving forward. @HarrisonWAffel if you have any other ideas for the approach here let me know

In general, if this is approved we will also need
* Test instructions (`make testacc? How did you run the tests and do they work?`)
* Test provisioning an RKE/K3s cluster on the latest Rancher v2.7-head with a custom non-default user and verify that it works for the requested RFE https://github.com/rancher/terraform-provider-rancher2/issues/315

docs/resources/service_account_token.md
sboschman: Yes, tokens can't be updated, any change to the resource basically becomes a recreate operation. See the existing rancher2_token resource for reference, it also uses the `ForceNew` behaviour.
A change to the `username` field means a token is requested for a completely different user(account). Hence I  marked them as `ForceNew` as well.

rancher2/schema_service_account_token.go
sboschman: Tried to keep it the same as the `rancher2_token` [fields](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/token#access_key) as much as possible. Besides, the 'access_key' acts as the username, and usernames shouldn't be treated as secrets afaik.

Rancher UI message after you create a token:
> Access Key and Secret Key can be sent as the username and password for HTTP Basic auth to authorize requests. You can also combine them to use as a Bearer token:

sboschman:
> There may be security concerns around being able to create a new user with a custom username/password to access the Rancher API for a specific cluster id via terraform so I'd discuss this with my team before moving forward. 

My 2c, the provider itself requires an admin token, so a new user with token created with terraform has at most the same admin permissions as the provider. Which does not grant you any more permissions than you already have. In practise we want to be able to create users+tokens with less permissions.

a-blender:
@sboschman Yep ok, I discussed with my team as well and have adjusted/responded to my comments. I saw there's been a new force push, let me know when it is ready for re-review. As a note: please _add_ commits for any additional changes. It is easier to see what has changed for reviewers. Don't worry about additional commits. Those can be squashed into one when we merge into the provider

rancher2/schema_service_account_token.go
annablender: Ah you are right. The access_key acts as the username, LGTM

docs/resources/service_account_token.md
a-blender: Thank you for the docs link in the other comment. This makes sense to me, since `username` and `password` are part of the token_resource here then they should also be `ForceNew` because they will be included in the recreate operation.

HarrisonWAffel (approved):

HarrisonWAffel:
outside of the other changes requested by Anna this  lgtm, i can reapprove once the other changes have been made

sboschman:
Changes made in response to the review:
1) rebased on current `master`

2) Add [ERROR] to all fmt.Errorf object strings returned from a function

    `rancher2/resource_rancher2_service_account_token.go`: [line 120](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-9edd0033c1e445e7d514f47687ae232287d9333115398ca63422be55be26d404R120)

    `rancher2/resource_rancher2_service_account_token_test.go`: [line 113](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-ec2be5f5cc8b50d792b3368ad6a0499c199c2afc8f4042fa8eb592b6f8c2ceddR113), [line 117](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-ec2be5f5cc8b50d792b3368ad6a0499c199c2afc8f4042fa8eb592b6f8c2ceddR117), [line 121](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-ec2be5f5cc8b50d792b3368ad6a0499c199c2afc8f4042fa8eb592b6f8c2ceddR121) and [line 140](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-ec2be5f5cc8b50d792b3368ad6a0499c199c2afc8f4042fa8eb592b6f8c2ceddR140)

3) Add Description: <description> to password, username, temp token, and temp token id fields

    `rancher2/schema_service_account_token.go`: [username desc](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-105315353b1d2e67ac62178ba99fad05d301a73806fa292e52d2fabd957761b6R15), [password desc](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-105315353b1d2e67ac62178ba99fad05d301a73806fa292e52d2fabd957761b6R21), [temp_token](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-105315353b1d2e67ac62178ba99fad05d301a73806fa292e52d2fabd957761b6R90) and [temp_token_id](https://github.com/rancher/terraform-provider-rancher2/pull/932/files#diff-105315353b1d2e67ac62178ba99fad05d301a73806fa292e52d2fabd957761b6R95)

swinn37:
Hello, do you know when this MR will be validated ?

a-blender (approved):
LGTM! Apologies for the delay. My last thoughts are that we may want to add a retry for create/delete a new user's token in TF in case either operation gets stuck in Rancher. Like, we don't want to add preliminary retries to work around bugs that are not an issue yet and this is just creating a token not provisioning infra, but a retry may be needed in the future since this PR https://github.com/rancher/terraform-provider-rancher2/pull/1033 is adding retries to most other v2 resources.

docs/resources/service_account_token.md
```diff
@@ -0,0 +1,83 @@
+---
+page_title: "rancher2_service_account_token Resource"
+---
+
+# rancher2\_service\_account\_token Resource
+
+Provides a Rancher v2 Token resource, specifically to create service account tokens. Service accounts tokens are tokens for other users (service accounts) than the Rancher v2 provider user. To create a service account token the username/password for the Rancher User must be known.
```
a-blender: Please update the naming for this resource as it's misleading. This feature creates a Rancher v2 API token for a new user not a kubernetes service account token.

docs/resources/service_account_token.md
```diff
@@ -0,0 +1,83 @@
...
+* `token` - (Computed/Sensitive) Token value (string)
+* `user_id` - (Computed) Token user ID (string)
+* `annotations` - (Computed) Annotations of the token (map)
+* `labels` - (Computed) Labels of the token (map)
+* `temp_token` - (Computed) Generated API temporary token as helper. Should be empty (string)
+* `temp_token_id` - (Computed) Generated API temporary token id as helper. Should be empty (string)
+
+## Timeouts
+
+`rancher2_service_account_token` provides the following
```
a-blender: Same as above, updating naming

rancher2/resource_rancher2_service_account_token.go
```diff
@@ -0,0 +1,189 @@
...
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	norman "github.com/rancher/norman/types"
+	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
+)
+
+func resourceRancher2ServiceAccountToken() *schema.Resource {
```
a-blender: Same as above, update naming. My suggestion would be to call this resource `resourceRancher2UserAccountAPIToken` and this func `resourceRancher2UserAccountAPITokenCreate` and then propagate that naming to the rest of the code

rancher2/resource_rancher2_service_account_token_test.go
```diff
@@ -0,0 +1,149 @@
...
+}
+`
+}
+
+func TestAccRancher2ServiceAccountToken_basic(t *testing.T) {
+	var user *managementClient.User
+	var globalRole *managementClient.GlobalRoleBinding
+	var token *managementClient.Token
+
+	testAccRancher2ServiceAccountToken = testAccRancher2User + testAccRancher2GlobalRoleBinding + testAccRancher2ServiceAccountToken
```
a-blender: Update naming in the tests as well

a-blender (changes requested):

rancher2/resource_rancher2_service_account_token.go
sboschman: Do we want to keep the naming in line with the existing `rancher2_user` (functions `resourceRancher2User() / resourceRancher2UserCreate()`) and `rancher2_token` (functions `resourceRancher2Token() / resourceRancher2TokenCreate()`) resources?

If so, perhaps we can consider just simply naming it `rancher2_user_token` (function `resourceRancher2UserToken`). Otherwise we introduce the wording 'user' vs 'user account' and  'api token' vs 'token'.

a-blender: @sboschman I had a discussion with a few folks about the naming, and I think it makes more sense to _split the resources_ you have implemented into separate `userAccount` and `token` resources because in Rancher, those resources are different entities.

Then you can have the "login to userAccount" action and create/update/delete operations in different resource files. That makes more sense because we are logging into Rancher with the new user and and creating a token... for the same "Create" func which is confusing.

So let's have

`resourceRancher2User` `resourceRancher2Token` (existing resource names for default user and API token)
`resourceRancher2CustomUser` `resourceRancher2CustomUserToken` (your resource names)

How does this sound?

sboschman: We only have 3 (terraform) resources, not 4 as I seem to understand from your comment.

`resourceRancher2User`: This existing resource allows user management, not just management of the provider user we use to connect to the Rancher API.
```
# Create a new rancher2 User
resource "rancher2_user" "foo" {
  name = "Foo user"
  username = "foo"
  password = "changeme"
  enabled = true
}
```

`resourceRancher2Token`: This existing resource only allows creating tokens for the provider user (the user we use to provision Rancher, which has super admin permissions)
```
# Create a new rancher2 Token
resource "rancher2_token" "foo" {
  description = "foo token"
  ttl = 1200
}
```

`resourceRancher2<something>Token`: This is a new resource which allows creating tokens for any Rancher User. It does not create the user itself. The user still has to be created with the `rancher2_user` resource. Which allows us to create a Rancher User with limited permission and add a token to this user. 

Unfortunately, at the time this PR was made, the Rancher API did not offer any endpoint to create tokens for a user other than the user the token used to access the Rancher API belongs to.
The login flow you mention is a workaround. The only way to acquire a Rancher API token is to log in using username+password, unless you already have a token for the user of course.

So, from a terraform provider point of view, the only resource we want to expose is a single resource which can be used to create tokens for any existing Rancher user.

Internally the terraform provider has to do a trick to make this happen:
1. Use the username+password of the user we want to create tokens for to login to Rancher and acquire a Rancher API token bound to this user
1. Make a call to the Rancher API Token endpoint authenticating with the token from step 1 to create an additional token, this is the token we are returning as 'the token' for the terraform resource. 
1. Logout the temporarily token we acquired in step 1.

It clearly is a different "flow" of things compared to all the other resources, which all can simply use the provider token to authenticate to the Rancher API. The `doUserLogin`, `doUserLogout`, etc are just helper functions very specific to this new resource. We do not return the temporary token acquired in step 1 as it has a limited lifespan, controlled by Rancher, and we want the option to create more than one token.

I am happy to name the new resource `resourceRancher2CustomUserToken` though, even if I don't really understand the reasoning behind the 'custom' part.

Usage would look like this:
```hcl
# Create a rancher2 Token
resource "rancher2_user" "foo" {
  name = "foo"
  username = "foo"
  password = "changeme"
  enabled = true
}
resource "rancher2_global_role_binding" "foo-login" {
  name = "foo-login-binding"
  global_role_id = "user-base"
  user_id = rancher2_user.foo.id
}
resource "rancher2_custom_user_token" "foo" {      <<< what does this 'custom' mean?
  username = rancher2_user.foo.username
  password = rancher2_user.foo.password
  description = "foo token"
  ttl = 0
  depends_on = [
    rancher2_global_role_binding.foo-login
  ]
}
```

a-blender: @sboschman Thank you for the clarification here.

Yes, please rename the resource to be `resourceRancher2CustomUserToken` so that it's clear this resource is only to generate a new Rancher API token for a new user (a custom user not a token for the default / admin user, that is why I think custom is good wording).

Could you please also draft a test plan and attach it to the GH issue with the _usage_ included? That will be necessary for our QA team to test this feature.

rancher2/resource_rancher2_custom_user_token.go
```diff
@@ -10,14 +10,25 @@ import (
...
-func resourceRancher2ServiceAccountToken() *schema.Resource {
+/*
+The Rancher API can only create tokens for the user owning the token used to authenticate the API call. To acquire
+the initial token to connect to the Rancher API a username+password login is needed. The lifespan of this token is
+controlled by Rancher.
+
+The following steps are used to create a Terraform controlled Rancher API token:
+1) Acquire a temporarily Rancher API token for the custom user by doing a username+password login
+2) Create a Terraform controlled token by authenticating using the (temporarily) token from step 1
+3) Logout the temporarilu token acquired in step 1
```
a-blender, line 21: Nit `Logout using the temporarily acquired token in step 1`

rancher2/schema_custom_user_token.go
```diff
@@ -6,19 +6,19 @@ import (
...
 
-func serviceAccountTokenFields() map[string]*schema.Schema {
+func customUserTokenFields() map[string]*schema.Schema {
 	s := map[string]*schema.Schema{
 		"password": {
 			Type:        schema.TypeString,
 			Required:    true,
 			Sensitive:   true,
-			Description: "The user username",
+			Description: "The user password",
```
a-blender, line 15: Good catch!

a-blender (approved):

rancher2/resource_rancher2_custom_user_token.go
sboschman: opened #1130 for it

PR #941, opened 2022-05-27
[Docs] Update wording for amazonec2_config userdata

dkeightley:
It appears this argument does not read a file, and expects userdata inline

### DIFF
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -123,7 +123,7 @@ The following attributes are exported:
 * `tags` - (Optional) AWS Tags (e.g. key1,value1,key2,value2) (string)
 * `use_ebs_optimized_instance` - (Optional) Create an EBS optimized instance. Default `false` (bool)
 * `use_private_address` - (Optional) Force the usage of private IP address. Default `false` (bool)
-* `userdata` - (Optional) Path to file with cloud-init user data (string)
+* `userdata` - (Optional) UserData content of cloud-init (string)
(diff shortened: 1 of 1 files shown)


a-blender (changes requested):
`userdata` is located in the docs in many more places than just this one line. To update the docs, please
* Add your test steps to verify that `user data` only expects data inline and not via a file. Did you just see this behavior with EC2? With other providers?
* Update all instances of this field in the docs.

kkaempf:
closing stale PRs.

dkeightley:
Sorry, just got alerted to this.

> `userdata` is located in the docs in many more places than just this one line. To update the docs, please
> 
>     * Add your test steps to verify that `user data` only expects data inline and not via a file. Did you just see this behavior with EC2? With other providers?

If my interpretation is accurate, the provider is not copying the file or contents, but [passing the value as a string](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_machine_config_v2_amazonec2.go#L50) to the API, so providing a file would be unusual, the [API schema](https://github.com/rancher/rancher/blob/release/v2.8/pkg/client/generated/management/v3/zz_generated_node_group.go) appears to align.

Here's an example I built some time back: https://gist.github.com/dkeightley/6e8e3de9567e2ca9c63294171d45c828#file-main-tf-L70-L76

> 
>     * Update all instances of this field in the docs.

I'll check through the other node drivers if you agree this is a desired change @a-blender?

dkeightley:
Reopening as per the above, please close if this change is not desired.

Issue #943, opened 2022-06-01
http_endpoint and http_tokens arguments missing from amazonec2_config node template block


axeal:
Rancher EC2 Node Templates support IMDSv2 configuration via the http_endpoint and http_tokens arguments (https://github.com/rancher/rancher/issues/25078).

Currently these options are exposed for the amazonec2_config block of the [machine_config_v2](https://github.com/rancher/terraform-provider-rancher2/blob/v1.24.0/docs/resources/machine_config_v2.md#amazonec2_config) resource in the Rancher Terraform Provider, but not for the amazonec2_config block of the [node_template resource](https://github.com/rancher/terraform-provider-rancher2/blob/v1.24.0/docs/resources/node_template.md#amazonec2_config).

SURE-4765

a-blender:
Terraform tests for these EC2 fields will be added in a second PR. Then, once that is merged I believe QA can test this using the rancher2 provider.

HarrisonWAffel:
The tests have been implemented in https://github.com/rancher/terraform-provider-rancher2/commit/9f7ba67f6467cfc8f8e1381b0be98d10b6124e52, (which is misnamed 944 vs 943), this should be good to test

HarrisonWAffel:
This issue can be tested with the latest terraform release. This change is related to https://github.com/rancher/rancher/issues/37943#event-7066651715, but does not require the IMDSv2 AWS policy to be enabled on your IAM user as this change relates to node templates. 

1. Create a new `.tf` file for a cluster on AWS
2. When defining the `amazonec2_config` block ensure that you add both the `http_endpoint` and `http_tokens` field. the value of `http_endpoint` must be `enabled`, and the value of `http_tokens` must be `required`. 
3. Create the cluster and navigate to the AWS dashboard
4. Find the EC2 instance created and ensure it was created properly
5. Navigate to the Rancher UI and find the newly created node template 
6. Edit the node template and scroll to section 4, `instance`
7. Check that `HTTP Endpoint` is equal to `Enabled` and that `HTTP Tokens` is equal to `Required`.

Josh-Diamond:
Ticket #943 - Test Results - ✅ 
---

With Docker on a single-node instance:

Verified with rancher2 provider `v1.24.1` on rancher `v2.7-bb9012ee0e66196136403d026b9e38c757c06c8c-head`:
1. Fresh install of rancher `v2.7-head`
[...]

PR #944, opened 2022-06-01
Add ec2 http_endpoint and http_tokens

axeal:
Add the ability to set http_endpoint and http_tokens for EC2 Node Templates 
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/943


HarrisonWAffel (approved):
This LGTM, however I'm unsure if we need to expand test coverage for these two new fields. I cannot see any `_test.go` files for the schema or structure files but perhaps @annablender knows if other tests should be enhanced for these two fields.

a-blender (changes requested):
@HarrisonWAffel You are right! If there aren't tests for the structure file, one should be added for the flattener and expander to make sure that tf is converting the terraform config to a plan in the correct way. Use any of the other test files as a template for this

a-blender (approved):
Tests will be added in another PR

Issue #949, opened 2022-06-16
No provision to pass labels to machine pools and nodes 


PrakashFromBunnings:
Hello , 

We have a requirement to have multiple machine pools and deploy different kind of workload to each pools. 

I was thinking to apply some labels to these pools , and pass my deployment to have nodeSelector .
So that  my deployment goes to selected nodeSelector label only . 

I cant see any options in terraform modules rancher2_cluser_v2  https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster_v2#machine_pools

Does any one have bright idea how can I achieve that?

Thanks

PrakashFromBunnings:
I can see adding labels option is available in GUI ( in Advance option ). Not sure if I am missing anything in terraform module.

azafairon:
Hi,

We have tried also to set the labels on machine pools which it seems that is supported:
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_cluster_v2_rke_config_machine_pool.go#L143

The above will populate the metadata on **MachineDeployment** but will not create the nodes with labels as the specs under templates are not correct.

Are we missing something ?

Thanks

m4rCsi:
Hi, 

we ran into this as well and were able to figure it out.

**Flow**

The flow (backwards) to get labels passed to the eventual nodes:
- (A) MachineDeployment needs to have it in `spec.template.metadata.labels`
- (B) A controller in rancher takes the a [RKEMachinePool](https://github.com/rancher/rancher/blob/c5af1ff21ce809fe298de160bd440a074fb484cc/pkg/apis/provisioning.cattle.io/v1/rke.go#L10-L31)  object and creates the MachineDeployments. ([link to relevant file](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go)). It is looking at data from provisioninong.cattle.io/v1/Cluster , specifically the field: `spec.rkeConfig.machinePools`
- (C) The `provisioninong.cattle.io/v1/Cluster` with machinePools inside is sent from this provider to rancher. And [this is the file](https://github.com/rancher/terraform-provider-rancher2/blob/34683372c426e490640529b842c39fd3f6043074/rancher2/structure_cluster_v2_rke_config_machine_pool.go) where the flattening/expanding between the RKEMachinePool and schema is happening.
[...]

PR #951, opened 2022-06-22
Allow setting labels on nodes with rke2

m4rCsi:
See https://github.com/rancher/terraform-provider-rancher2/issues/949#issuecomment-1163236185 for a detailed writeup.

This is to allow setting labels on nodes with rke2.

The implemented solution creates a new variable `machine_labels` that can be set inside `machine_pools` resource.
The existing `labels` of `machine_pools` is applied to the DeploymentMachine Resource directly (not the eventual nodes through the spec)

addresses rancher/terraform-provider-rancher2#949

### DIFF
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -475,7 +475,8 @@ The following attributes are exported:
 * `unhealthy_node_timeout_seconds` - (Optional) Seconds an unhealthy node has to become active before it is replaced (int)
 * `max_unhealthy` - (Optional) Max unhealthy nodes for automated replacement to be allowed (string)
 * `unhealthy_range` - (Optional) Range of unhealthy nodes for automated replacement to be allowed (string)
-* `machine_labels` - (Optional) Machine pool labels. Node labels (list)
+* `machine_labels` - (Optional) Labels for Machine pool nodes (map)
+* `labels` - (Optional) Labels for Machine Deployment Resource (map)
 
 ##### `machine_config`
 
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -475,6 +475,7 @@ The following attributes are exported:
 * `unhealthy_node_timeout_seconds` - (Optional) Seconds an unhealthy node has to become active before it is replaced (int)
 * `max_unhealthy` - (Optional) Max unhealthy nodes for automated replacement to be allowed (string)
 * `unhealthy_range` - (Optional) Range of unhealthy nodes for automated replacement to be allowed (string)
+* `machine_labels` - (Optional) Machine pool labels. Node labels (list)
 
 ##### `machine_config`
 
--- a/rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
+++ b/rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
@@ -1,12 +1,14 @@
 package rancher2
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"reflect"
 	"testing"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	provisionv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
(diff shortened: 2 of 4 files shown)


HarrisonWAffel (commented):
@m4rCsi, thanks for you contribution! This LGTM, and tests are passing, but input from @annablender would be greatly appreciated. 

We should also probably add the documentation for the existing argument for labeling machineDeployments.

docs/resources/cluster_v2.md
```diff
@@ -475,6 +475,7 @@ The following attributes are exported:
 * `unhealthy_node_timeout_seconds` - (Optional) Seconds an unhealthy node has to become active before it is replaced (int)
 * `max_unhealthy` - (Optional) Max unhealthy nodes for automated replacement to be allowed (string)
 * `unhealthy_range` - (Optional) Range of unhealthy nodes for automated replacement to be allowed (string)
+* `machine_labels` - (Optional) Machine pool labels. Node labels (list)
```
a-blender: I think this should be changed to `(Optional) Labels for a machine pool deployment (list)` to be more clear about what this is.

rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
```diff
@@ -62,6 +64,12 @@ func init() {
 				"label_one": "one",
 				"label_two": "two",
 			},
+			RKECommonNodeConfig: rkev1.RKECommonNodeConfig{
```
a-blender, line 67: Spacing

rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
```diff
@@ -97,6 +105,10 @@ func init() {
 				"label_one": "one",
 				"label_two": "two",
 			},
+			"machine_labels": map[string]interface{}{
```
a-blender, line 108: Fix bracket spacing in tests (same as above) ^

a-blender (changes requested):
Thanks @HarrisonWAffel! LGTM, this needs testing with provisioning an rke2 cluster in rancher with the tf rancher2 provider with 2+ labels added to at least one machine pool. Verify in the rancher ui that the labels show up in the deployment yaml (if this has already been done, let me know)

docs/resources/cluster_v2.md
m4rCsi: I have updated the documentation to document both `labels` and `machine_labels`. I hope this is clearer now. Happy to revise again. I'm having trouble coming up with something that is clear :)

rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
m4rCsi: I looked at this, and I couldn't figure out where or how the spacing is wrong. It seems to be consistent with this and other files in this repo?

rancher2/structure_cluster_v2_rke_config_machine_pool_test.go
m4rCsi: I looked at this, and I couldn't figure out where or how the spacing is wrong. It seems to be consistent with this and other files in this repo?

m4rCsi:
> Thanks @HarrisonWAffel! LGTM, this needs testing with provisioning an rke2 cluster in rancher with the tf rancher2 provider with 2+ labels added to at least one machine pool. Verify in the rancher ui that the labels show up in the deployment yaml (if this has already been done, let me know)

I ran tests when I was developing this PR, but only with 1 Label. 
The following images show, how I made a change in terraform, and how it shows up in the UI in rancher:

HarrisonWAffel (approved):

a-blender:
@m4rCsi did you also add `node-role: control-plane` in your terraform changes during testing? I see that label was also added to a machine pool.

a-blender (approved):

m4rCsi:
> @m4rCsi did you also add `node-role: control-plane` in your terraform changes during testing? I see that label was also added to a machine pool.

Yes, I also did have some other node-roles during testing.

Martin-Weiss:
FYI - just tried this with 1.24.2 and it works like a charm! Thanks so much :-)

Issue #347, opened 2020-05-20
Namespace container resource limit ignored when used with empty values


anouarchattouna:
* Create a namespace with empty values for `container_resource_limit`
```tf
# Create a new rancher2 Project
resource "rancher2_project" "foo" {
  name = "foo"
  cluster_id = "c-d9lbr"
  resource_quota {
    project_limit {
      limits_cpu = ""
      limits_memory = ""
      requests_storage = ""
    }
    namespace_default_limit {
      limits_cpu = ""
      limits_memory = ""
      requests_storage = ""
    }
  }
  container_resource_limit {
    limits_cpu = ""
    limits_memory = ""
    requests_cpu = ""
    requests_memory = ""
  }
}
# Create a new rancher2 Namespace
resource "rancher2_namespace" "foo" {
  name = "foo"
  project_id = rancher2_project.foo.id
  description = "foo namespace"
  resource_quota {
    limit {
      limits_cpu = ""
      limits_memory = ""
      requests_storage = ""
    }
  }
  container_resource_limit {
    limits_cpu = ""
    limits_memory = ""
    requests_cpu = ""
    requests_memory = ""
  }
}
```
```shell
rancher2_project.foo: Creating...
rancher2_project.foo: Creation complete after 4s [id=c-d9lbr:p-smbgq]
rancher2_namespace.foo: Creating...
rancher2_namespace.foo: Creation complete after 1s [id=foo]

Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
```
* Execute your plan a second time
```shell
  # rancher2_namespace.foo will be updated in-place
  ~ resource "rancher2_namespace" "foo" {
        annotations      = {
            "cattle.io/status"                          = jsonencode(
                {
                    Conditions = [
                        {
                            LastUpdateTime = "2020-05-20T13:55:06Z"
                            Message        = ""
                            Status         = "True"
                            Type           = "InitialRolesPopulated"
                        },
                        {
                            LastUpdateTime = "2020-05-20T13:55:05Z"
                            Message        = ""
                            Status         = "True"
                            Type           = "ResourceQuotaValidated"
                        },
                        {
                            LastUpdateTime = "2020-05-20T13:55:05Z"
                            Message        = ""
                            Status         = "True"
                            Type           = "ResourceQuotaInit"
                        },
                    ]
                }
            )
            "lifecycle.cattle.io/create.namespace-auth" = "true"
        }
        description      = "foo namespace"
[...]

PR #958, opened 2022-07-15
Only set namespace `container_resource_limit` when it is not nil and empty

manhtukhang:
Like **[structure_project/flattenProject](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_project.go#L123-L129)** func, **[structure_namespace/flattenNamespace](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_namespace.go#L123-L127)** should only set `container_resource_limit` if `ContainerDefaultResourceLimit` return from Rancher API is **not nil** and **not empty**

This resolves issue https://github.com/rancher/terraform-provider-rancher2/issues/347 ([`container_resource_limit` block is empty](https://github.com/rancher/terraform-provider-rancher2/issues/347#issuecomment-632556721))

### DIFF
--- a/rancher2/structure_namespace.go
+++ b/rancher2/structure_namespace.go
@@ -120,14 +120,17 @@ func flattenNamespace(d *schema.ResourceData, in *clusterClient.Namespace) error
 	d.Set("name", in.Name)
 	d.Set("description", in.Description)
 
-	containerLimit := flattenNamespaceContainerResourceLimit(in.ContainerDefaultResourceLimit)
-	err := d.Set("container_resource_limit", containerLimit)
-	if err != nil {
-		return err
+	emptyContainerResourceLimit := clusterClient.ContainerResourceLimit{}
+	if in.ContainerDefaultResourceLimit != nil && *in.ContainerDefaultResourceLimit != emptyContainerResourceLimit {
+		containerLimit := flattenNamespaceContainerResourceLimit(in.ContainerDefaultResourceLimit)
+		err := d.Set("container_resource_limit", containerLimit)
+		if err != nil {
+			return err
+		}
 	}
 
 	resourceQuota := flattenNamespaceResourceQuota(in.ResourceQuota)
-	err = d.Set("resource_quota", resourceQuota)
+	err := d.Set("resource_quota", resourceQuota)
 	if err != nil {
 		return err
 	}


manhtukhang:
@HarrisonWAffel Could you review this PR, please?

HarrisonWAffel:
After reading the linked issue, I'm still not sure what the use case is for defining an empty block of `container_resource_limit `. @manhtukhang could you please elaborate on what you're trying to do for my own understanding? 

Also, I see that unit tests are passing but have you tested this by creating a live cluster? In general, I think having consistent behavior in the provider is a good thing, but understanding your use case would be helpful

manhtukhang:
Thanks for your feedback @HarrisonWAffel,

Let me make more clear:

- > I'm still not sure what the use case is for defining an empty block of `container_resource_limit`

  In my use case, I write a TF module that takes `resource_quota` and `container_resource_limit`'s attributes dynamic, like this:
  + Module:
    ```hcl
    resource "rancher2_namespace" "main" {
      provider   = rancher2
      depends_on = [rancher2_project.main]
      for_each   = { for v in var.rancher_cluster_namespaces : v.name => v }
      name       = each.key
      project_id = rancher2_project.main["${each.value.project}"].id
      resource_quota {
        limit {
          limits_cpu       = try(each.value.resource_quota.limits_cpu, null)
          limits_memory    = try(each.value.resource_quota.limits_memory, null)
          requests_cpu     = try(each.value.resource_quota.requests_cpu, null)
          requests_memory  = try(each.value.resource_quota.requests_memory, null)
          requests_storage = try(each.value.resource_quota.requests_storage, null)
        }
      }
      container_resource_limit {
        limits_cpu      = try(each.value.container_resource_limit.limits_cpu, null)
        limits_memory   = try(each.value.container_resource_limit.limits_memory, null)
        requests_cpu    = try(each.value.container_resource_limit.requests_cpu, null)
        requests_memory = try(each.value.container_resource_limit.requests_memory, null)
      }
    }
    ```
  + Variable inputs:
    ```yaml
    namespaces:
      - name: test-a
        container_resource_limit:
          limits_cpu: 10000
          limits_memory: 20480
          requests_cpu: 2000
          requests_memory: 4096
      - name: test-b
    ```
  as you see, in the case of namespace `test-b`, these attributes do not provide, `container_resource_limit`  block will be rendered as empty

- > Also, I see that unit tests are passing but have you tested this by creating a live cluster?

  Yes. I have been using it in my production environment

HarrisonWAffel:
Okay, I think I have a better understanding of your use-case, thank you for elaborating. In essence, when using dynamic blocks to render namespaces some namespaces will not have `container_resource_limit` defined, at which point you would like to have the block omitted entirely as it is no longer relevant to that namespace. Please correct me if I'm wrong. This sounds reasonable to me, I've requested that others from my team also review this in case I've missed anything.

manhtukhang:
Hi @HarrisonWAffel,

I've found a solution to the problem, therefore this PR is not necessary.
Please help me close the PR!

Thanks for your support!

Issue #922, opened 2022-05-11
Add RKE Support for ACI Network Provider to Terraform Provider


a-blender:
Add RKE support for the [ACI Network Provider](https://www.cisco.com/c/en/us/solutions/data-center-virtualization/application-centric-infrastructure/index.html) to the rancher2 terraform provider. The [aci_network_provider](https://registry.terraform.io/providers/rancher/rke/latest/docs/resources/cluster#aci_network_provider) is already supported in the docs. It just needs to be added to the provider.

a-blender:
Discussed with @Oats87 since I noticed the `aci_network_provider` field is documented in the terraform docs for the [rke_cluster](https://registry.terraform.io/providers/rancher/rke/latest/docs/resources/cluster#aci_network_provider) resource but not in the [rancher2_cluster](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#network) resource. This is not a blocker. It will be implemented in the rancher 2 terraform provider, but ACI is not supported as an option in the UI.

annablender:
I had to bump several packages that were out of date in the tf provider go.mod. Golang has also been updated to 1.17. In addition, drone was failing with errors when using the latest golang alpine image `golang 1.17.3-alpine3.16 so @kinarashah and I updated the drone image to golang 1.17.3 and the build image to golang 1.17.3-alpine3.13 because of this issue https://github.com/docker-library/php/issues/1177.

We're suspicious there may be a problem with how drone runs the tf tests so I've commented out the tests in the GNU makefile.

a-blender:
In order for ACI network provider support to work properly, the tf rancher2 provider had to have the rancher/rancher and go client dependencies bumped, as well as go updated to 1.17. The tf provider now uses go 1.17-alpine3.13 instead of 3.14 due to this issue https://github.com/Mellanox/k8s-rdma-shared-dev-plugin/issues/49. Drone was also having issues running tests with that image, so drone now uses go 1.17.3.

The dependency bump PR https://github.com/rancher/terraform-provider-rancher2/pull/959 has been merged. I rebased the updates into  https://github.com/rancher/terraform-provider-rancher2/pull/912 (ACI network provider support) which should fix the CI.

a-blender:
https://github.com/rancher/terraform-provider-rancher2/pull/912 has been merged

a-blender:
[...]

PR #959, opened 2022-07-16
Bump rancher/rancher and go client deps

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/922

I'm not _exactly_ sure if this is the right way to do this, but we need to bump rancher to pull in the latest rke updates so ACI network provider support will work. @kinarashah


HarrisonWAffel (approved):
The build failure doesn't seem to be related to the changes made here (its having a problem reaching the docker-socket, I restarted it for you). I found an older PR where [the rancher dep is bumped]( https://github.com/rancher/terraform-provider-rancher2/commit/b4d5bcf1156c9f97172855598e36f56a5c13964f#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6) and this looks to being doing the same thing, so LGTM as long as build passes and it works locally 

edit: build is still failing, looks like it may be related to these changes

jakefhyde (approved):

a-blender:
@kinarashah It appears that the build is failing with the `sharedInformer` error because go client is not getting updated. Trying to fix this on my end

edit: My go.sum was not updated due to the difference in how go 1.16 and 1.17 handle dependencies. I ran a `go mod tidy -compat=1.17` to remove the go 1.16 checksums from the go.mod file and that worked. Needs a re-review to make sure that this change is ok to merge.

a-blender:
I had to bump several packages that were out of date in the tf provider go.mod. Golang has also been updated to 1.17. In addition, drone was failing with errors when using the latest golang alpine image `golang 1.17.3-alpine3.16 so @kinarashah and I updated the drone image to golang 1.17.3 and the build image to golang 1.17.3-alpine3.13 because of this issue https://github.com/docker-library/php/issues/1177.

We're suspicious there may be a problem with how drone runs the tf tests so I've commented out the tests in the GNU makefile.

annablender:
@kinarashah Could you confirm that this is ok to merge even with the commented-out tests?

a-blender:
So the build passed with commented out tests, which means there's an issue with how drone runs the tf tests. Will need to debug this and fix this before merging.

a-blender:
Issue was flaky tests, unsurprisingly :)

PR #983, opened 2022-09-01
Add a note about Azure AD auth configuration migration

maxsokolovsky:
This is relevant for setups which were created before v2.6.7 but which had the underlying Rancher image updated to v2.6.7+.


a-blender (approved):
Docs LGTM

HarrisonWAffel (approved):

PR #992, opened 2022-09-09
Don't mark certificates as sensitive

tmatilai:
The certificates are not sensitive, only the certificate keys are.

My small annoyance came from the `rancher2_cluster` data source, but while at it, I tried to remove the sensitivity flag from all CA and normal certificates.

It looks like there are also many username/ID/etc. attributes marked as sensitive even though most shouldn't needed to be. But those I left for someone else for now. 🙂


kkaempf:
closing stale PRs.

tmatilai:
Oh, I didn’t know I was expected to do something as there was no reaction or feedback whatsoever. But I guess you are not interested. 😅

PR #994, opened 2022-09-09
Clarify source of `image_name` for `harvester_config` blocks

anoldguy:
This one stumped me for a bit. Finding out that the `image_name` it wanted was the one in the API, `default/image-abcde` and *not* the one that I had created in harvester, `default/ubuntu-22.04-server-cloudimg-amd64.img`. 

It might not be an issue if you're using the harvester TF provider and the `harvester_image` data source, but the documentation could help point the user in the correct direction.

### DIFF
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -189,7 +189,7 @@ The following attributes are exported:
 * `memory_size` - (Optional) Memory size (in GiB), Default `4` (string)
 * `disk_size` - (Optional) Disk size (in GiB), Default `40` (string)
 * `disk_bus` - (Optional) Disk bus, Default `virtio` (string)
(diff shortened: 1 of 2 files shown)


futuretea:
What this field expects is the name of the image, like `image-abcde`
If the image is not in the same namespace as the vm, namespace specification is also supported like `default/image-abcde`
The terraform-provider-harvester use namespace/name as resource id
So this field can be filled with the terraform provider harvester id of the image
The name on the UI `ubuntu-22.04-server-cloudimg-amd64.img` is actually the displayName of the image, the real name of the image needs to be checked through view yaml.

futuretea (approved):
LGTM. @snasovich can you help to review and merge this PR? Thanks

kkaempf (approved):
lgtm

kkaempf:
Thanks for your contribution !

kkaempf:
But was solved before via an example

Issue #867, opened 2022-02-18
Occasionally managed rke1 cluster re-provisioning fails using terraform provider


riuvshyn:
**Rancher Server Setup**
- Rancher version: 2.6.2
- Installation option (Docker install/Helm Chart): 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: 1.21.5
- Cluster Type (Local/Downstream): local
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**
Occasionally managed rke1 cluster re-provisioning fails using terraform provider
provider version 1.22.2

**To Reproduce**
- provision managed rke1 cluster using terraform provider
- destroy it
- re-provision

**Result**
Provisioning fails with following error:
```
module.cluster.rancher2_cluster.this: Creating...
│ Error: Bad response statusCode [409]. Status [409 Conflict]. Body: [baseType=error, code=AlreadyExists, message=clusterregistrationtokens.management.cattle.io "default-token" already exists] from [https://api.rancher.test.com/v3/clusterregistrationtokens]
```

**Expected Result**
Managed cluster provisioned sucessfully

**Screenshots**

**Additional context**
After destroy the `default-token` goes away from from `ClusterRegistrationTokens` along with namespace that was created for the new cluster but it still fails with already exists error.

iTaybb:
I get this sometimes on brand new clusters.

phillamb168:
I've gotten it twice in a row just now.

elvinasp:
Can confirm, we have seen this too, when deploying new cluster. When applying for the second time, TF finishes successfully.  
Rancher 2.6.2 and 2.6.5
K8s: v1.21.5-rancher1-1
TF provider:
```
...
  required_providers {
    rancher2 = {
      source  = "rancher/rancher2"
      version = ">= 1.12.0, < 2.0.0"
    }
```

timBeehexa:
> Can confirm, we have seen this too, when deploying new cluster. When applying for the second time, TF finishes successfully. Rancher 2.6.2 and 2.6.5 K8s: v1.21.5-rancher1-1 TF provider:
> 
> ```
> ...
>   required_providers {
>     rancher2 = {
>       source  = "rancher/rancher2"
[...]

PR #997, opened 2022-09-20
Use existing cluster registration token if conflict

jakefhyde:
rancher/terraform-provider-rancher2#867
https://github.com/rancher/terraform-provider-rancher2/issues/1098

Added log messages and logic to reuse cluster registration token if present. The token doesn't exist the first time we check it, but it does when we try to create it, but erroring her doesn't make any sense since we want it created anyway.

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -537,6 +537,7 @@ func findFlattenClusterRegistrationToken(client *managementClient.Client, cluste
 }
 
 func findClusterRegistrationToken(client *managementClient.Client, clusterID string) (*managementClient.ClusterRegistrationToken, error) {
+	log.Printf("[TRACE] Finding cluster registration token for %s", clusterID)
 	for i := range clusterRegistrationTokenNames {
 		regTokenID := clusterID + ":" + clusterRegistrationTokenNames[i]
 		for retry, retries := 1, 10; retry <= retries; retry++ {
@@ -545,18 +546,24 @@ func findClusterRegistrationToken(client *managementClient.Client, clusterID str
 				if !IsNotFound(err) {
 					return nil, err
 				}
+				log.Printf("[TRACE] Cluster registration token %s not found for %s", regTokenID, clusterID)
 				break
 			}
 			if (len(regToken.Command) > 0 && len(regToken.NodeCommand) > 0) || retry == retries {
+				log.Printf("[INFO] Found existing cluster registration token for %s", clusterID)
 				return regToken, nil
 			}
+			log.Printf("[DEBUG] Sleeping for 3 seconds before checking cluster registration token for %s", clusterID)
 			time.Sleep(3 * time.Second)
 		}
 	}
+	log.Printf("[TRACE] Cluster registration token not found for %s", clusterID)
 	return createClusterRegistrationToken(client, clusterID)
 }
 
 func createClusterRegistrationToken(client *managementClient.Client, clusterID string) (*managementClient.ClusterRegistrationToken, error) {
+	log.Printf("[DEBUG] Creating cluster registration token for %s", clusterID)
+
 	regToken, err := expandClusterRegistationToken([]interface{}{}, clusterID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


rancher2/resource_rancher2_cluster.go
```diff
@@ -545,25 +546,36 @@ func findClusterRegistrationToken(client *managementClient.Client, clusterID str
...
 					return nil, err
 				}
+				log.Printf("[TRACE] Cluster registration token %s not found for %s", regTokenID, clusterID)
 				break
 			}
 			if (len(regToken.Command) > 0 && len(regToken.NodeCommand) > 0) || retry == retries {
+				log.Printf("[INFO] Found existing cluster registration token for %s", clusterID)
 				return regToken, nil
 			}
+			log.Printf("[DEBUG] Sleeping for 3 seconds  cluster registration token for %s", clusterID)
```
HarrisonWAffel: reads weird, maybe `[DEBUG] Sleeping for 3 seconds before checking cluster registration token for %s`?

rancher2/resource_rancher2_cluster.go
```diff
@@ -583,6 +595,7 @@ func createClusterRegistrationToken(client *managementClient.Client, clusterID s
 	if err != nil {
 		return nil, err
 	}
+	log.Printf("[INFO] Creatied cluster registration token %s for %s", newRegToken.ID, clusterID)
```
HarrisonWAffel: `Creatied` -> `Created`

HarrisonWAffel (approved):

rancher2/resource_rancher2_cluster.go
```diff
@@ -537,6 +537,7 @@ func findFlattenClusterRegistrationToken(client *managementClient.Client, cluste
 }
 
 func findClusterRegistrationToken(client *managementClient.Client, clusterID string) (*managementClient.ClusterRegistrationToken, error) {
+	log.Printf("[TRACE] Finding cluster registration token for %s", clusterID)
 	for i := range clusterRegistrationTokenNames {
 		regTokenID := clusterID + ":" + clusterRegistrationTokenNames[i]
 		for retry, retries := 1, 10; retry <= retries; retry++ {
```
a-blender, line 543: Could we try 3-4 times to find the cluster registration token before we use an existing? Did retrying 10 times work better in testing?

jakefhyde: We try 10 times for each token (2 max, generally 1 though) before creating a new one. This works better IMO, as we want to use the generated one, but occasionally, trying to create it would fail because the generated one would be created before we could create one. I think in that case it's better to just return that one, so the total logic is as follows:
- for each token
  - try ten times to get the token
    - if we found it, return it
    - if it doesn't exist, sleep 3 seconds before trying again
    - if any other error, return
- if no tokens were found
  - create a new token, and return it
  - if this token fails to create because it already exists, return that
  - any other error, return
 
We might even do well to bump the retries up to 15 to prevent thrashing, but I'm not sure of the potential side effects of not waiting long enough to create the token.

jakefhyde: @annablender I think an easier explanation is that since the token may be populated after we stop waiting for it to be created, its a race condition. Does that make sense?

a-blender: Yes, I think I understand here. So, during those retries the token may be generated, in which case we stop retrying and use the generated one? And it's better to have more retries (say 10), than less retries because we want to give as much time as possible for rancher to _generate_ the token.. before we hit said scenario: token creation + generation at the same time. Yes?

a-blender: Ok, that sounds fine to me then

a-blender (approved):

rancher2/resource_rancher2_cluster.go
jakefhyde: Yes, that correct. The other issue is that these operations can't be done atomically, i.e. we can wait forever, but we still run the risk of rancher generating the token after we try to create it, which is why the change was made to just return that in the event it already exists.

PR #1001, opened 2022-10-12
Pass machine labels to kubernetes nodes

lukeelten:
According to the [documentation](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/machine_config_v2#labels) the labels set on the machine configs are passed to the kubernetes nodes.
Currently this is not happening because the provider only sets `MachineDeploymentLabels` which is not propagated to the kubernetes nodes. The correct attribute would be `labels`.


lukeelten:
Has already been fixed in #951

PR #1012, opened 2022-10-24
Implement Azure Tags Support

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/26955

This PR adds the `tags` field to the node template, machine config, and cloud provider for Azure. This allows tags to be specified for particular VM instances using either the node template or machine config, as well as adding tags to supporting infrastructure via the tags field within the Azure cloud provider. 

Testing: 
Manual testing is in progress, however I've added unit tests for the Azure node template and improved prior unit tests for the AmazonEC2 node template. 

I'm raising this PR as a draft as the `tags` field within the Azure Cloud Provider needs to be generated in rancher/rancher for this PR to work. Because of this dependency the build will fail


HarrisonWAffel:
Closed in favor of https://github.com/rancher/terraform-provider-rancher2/pull/1023

PR #1014, opened 2022-10-27
Updated tf rancher2 release notes

a-blender:
Issue: https://github.com/rancher/terraform-provider-rke/issues/364

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ $ EXPOSE_HOST_PORTS=true make docker-testacc
 Releasing the Provider
 ---------------------------
 
+The provider should be released 'out of band' from Rancher, but can be loosely tied to a Rancher release to track issues. For example: Terraform v2.6.9 should be released a few days after Rancher v2.6.9 and Terraform fixes/features in the release are said to be included in Rancher v2.6.9 (and will work if you provision with Terraform).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,6 +87,12 @@ $ EXPOSE_HOST_PORTS=true make docker-testacc
 Releasing the Provider
 ---------------------------
 
+The provider should be released 'out of band' from Rancher, but can be loosely tied to a Rancher release to track issues. For example: Terraform v2.6.9 should be released a few days after Rancher v2.6.9 and Terraform fixes/features in the release are said to be included in Rancher v2.6.9 (and will work if you provision with Terraform).
+
+The [RKE provider](https://github.com/rancher/terraform-provider-rke) should also be checked for changes and released 'out of band' along with Rancher. If there are no updates, do not release.
```
jakefhyde: Might want to update this to say that the rke tf provider should be released after every rke/kdm release. While RKE can pull KDM dynamically (https://github.com/rancher/rke/blob/89246c37859ebc5bc90e6d22e695e773c9562edd/metadata/metadata.go#L62), it doesn't do so by default, so I don't think the provider will do that by default either: https://github.com/rancher/terraform-provider-rke/blob/b64dcadbb8502fd64108534a1f98592c9e0d03e3/rke/schema_rke_cluster.go#L171

annablender: I agree with this

jakefhyde (approved):

sbstp (approved):

PR #1023, opened 2022-11-04
[v2.7.1] Add support for new azure features

HarrisonWAffel:
### Testing can be tracked in this ticket: https://github.com/rancher/terraform-provider-rancher2/issues/1071

This PR adds support for the following Azure features 
1. The Azure cloud provider now accepts tags. Expanders and flatteners now support that field
2. The Azure machine config and node template now accept tags which are added to Azure VM Instances via rancher-machine. These tags 
3. The Azure machine config and node template now allow you to denote if you want to create a network interface utilizing Accelerated Networking.
4. The Azure machine config and node template now allow you to denote if you want to create the Azure VM in a particular Availability Zone.
5. The Azure machine config and node template now allow you to specify if you want to use the standard SKU for the public IP created for an Azure VM. This must be enabled for the creation of an Azure VM in a particular Availability Zone. 

This PR also updates the version of rancher used as a dependency (and its indirect dependencies) to the latest commit of dev-v2.6, in order to pick up the new tags field in the Azure cloud provider. 

Edit 1-06-2023: 

I've resolved a build error that was blocking this PR. The drone runner was using go 1.17, however the most recent version of the Rancher dependency uses indirect dependencies which require at least 1.18 since they utilize generics. I've updated the drone runner to use go 1.19, as well as updated the version of docker being used to [get around an alpine issue which was fixed in v3.14](https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2).


a-blender:
@HarrisonWAffel Generally we say [DNM v2.7.1] if a PR can only be merged pending a release branch existing ;) CI is also being flaky so I restarted it.

HarrisonWAffel:
Now that the Azure features have been validated in the Rancher frontend and backend I've circled back on this PR for terraform support. I've run into similar issues as @annablender when bumping dependencies within `go.mod`, and found that some of the issues are caused when bringing in rancher 2.7. Bumping the rancher version to the latest commit of 2.6 resolves a majority of the issues regarding structured logging and other areas, but one is still present which is preventing this PR from proceeding. 


I'm still looking into why this worked before the bump, and how to resolve the issue

edit: 
It looks like some of the build steps for this PR are not using go 1.18, but new dependencies brought in by rancher require at least 1.18 as they contain generics.

.drone.yml
```diff
@@ -20,7 +20,7 @@ steps:
   privileged: true
 
 - name: build-all-binaries
-  image: golang:1.17.3
```
jakefhyde, line 23: I'd bump to 1.19 if possible

HarrisonWAffel: Sounds good, even bumping to `1.18` has caused other errors in the build process, so there's likely more I have to do to get it all working properly

HarrisonWAffel: bumped version to `v1.19.4`, builds are now passing

a-blender (approved):
LGTM now

rancher2/structure_node_template_azure_test.go
```diff
@@ -0,0 +1,192 @@
...
+		err := flattenNodeTemplate(output, tc.Input)
+		if err != nil {
+			t.Fatalf("[ERROR] on flattener: %#v", err)
+		}
+		expectedOutput := map[string]interface{}{}
+		for k := range tc.ExpectedOutput {
+			expectedOutput[k] = output.Get(k)
+		}
+		if !reflect.DeepEqual(expectedOutput, tc.ExpectedOutput) {
+			t.Fatalf("Unexpected output from flattener. \nExpected: %#v\nGiven:::: %#v", expectedOutput, tc.ExpectedOutput)
```
jakefhyde:
```suggestion
			t.Fatalf("Unexpected output from flattener.\nExpected: %#v\nGiven: %#v", expectedOutput, tc.ExpectedOutput)
```
or
    
```suggestion
			t.Fatalf("Unexpected output from flattener.\nExpected: %#v\nGiven:    %#v", expectedOutput, tc.ExpectedOutput)
```
They seem to be used interchangeably.

rancher2/structure_node_template_azure_test.go
```diff
@@ -0,0 +1,192 @@
...
+			Input:          testNodeTemplateExpandAzureConfInterface,
+			ExpectedOutput: testAzureNodeTemplateConf,
+		},
+	}
+
+	for _, tc := range cases {
+		inputData := schema.TestResourceDataRaw(t, nodeTemplateFields(), tc.Input)
+		output := expandNodeTemplate(inputData)
+		if !reflect.DeepEqual(output, tc.ExpectedOutput) {
+			t.Fatalf("Unexpected output from expander.\nExpected: %#v\nGiven:::: %#v", tc.ExpectedOutput, output)
```
jakefhyde: same

jakefhyde (changes requested):
Just a few nits

rancher2/structure_node_template_azure_test.go
HarrisonWAffel: Oh, that must have sneaked back in during some resets while testing the build issues

rancher2/structure_node_template_azure_test.go
HarrisonWAffel: fixed

jakefhyde (approved):

PR #1034, opened 2022-11-18
[Q3] Add script to build local binary

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1156

Devs and customers do _a lot_ of local testing with terraform. There's two ways to test locally: 1) configure Terraform to use a local binary that you've generated, or 2) use an `in-dev` version of the provider with `dev-overrides` [relevant article](https://nicovibert.com/2021/03/29/dev-overrides-terraform/), thank you @git-ival. This script is for automating the first option.

~~I only have the script configured to work with a Mac fs. Feel free to add to the script, add params, or configure it for where Terraform is installed on other systems.~~

I have the script configured to work with most operating sys compatible with Rancher, Mac/linux and win. The user can also specify -i for an additional `terraform init` and/or -a for a `terraform apply` which I've found greatly increases the efficiency of local testing.

## Manual Testing

Tested script to make sure all possible usages run correctly. Example:

```
./scripts/build_local_binary -i -a
```

### DIFF
--- a/scripts/build_local_binary.sh
+++ b/scripts/build_local_binary.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+<<comment
+When you run a `make bin` in the terraform-provider-rancher2 repo, it
+downloads a binary into your local cache from the specified source. To
+tell Terraform to use a local binary, you need to overwrite the one in
+the cache.
+
+Add the following to your Terraform config file. This tells Terraform
+to use the local cache https://terraform.io/cli/config/config-file.
+
+terraform {
+required_providers {
+    rancher2 = {
+      source = "terraform.local/local/rancher2"
+      version = "1.0.0"
+    }
+  }
+}
+comment
+
+# Compile the local binary
+make bin
+
+opsys=windows
+if [[ "$OSTYPE" == linux* ]]; then
+  opsys=linux
+elif [[ "$OSTYPE" == darwin* ]]; then
+  opsys=darwin
+fi
+
+# Supported values of 'arch': amd64, arm64, ppc64le, s390x
+case $(uname -m) in
+x86_64)
+    arch=amd64
+    ;;
+arm64)
+    arch=arm64
+    ;;
+ppc64le)
+    arch=ppc64le
+    ;;
+s390x)
+    arch=s390x
+    ;;
+*)
+    arch=amd64
+    ;;
+esac
+
+LOCAL_BINARY_PATH=~/.terraform.d/plugins/terraform.local/local/rancher2/1.0.0/${opsys}_${arch}
+
+if [ ! -d $LOCAL_BINARY_PATH ]
+then
+  echo "Directory ${LOCAL_BINARY_PATH} does not exist. Creating..."
+  mkdir -p $LOCAL_BINARY_PATH
+fi
+
+# Overwrite the cached binary with your local one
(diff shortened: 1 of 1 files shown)


scripts/build_local_binary.sh
```diff
@@ -0,0 +1,30 @@
...
+    }
+  }
+}
+```
+
+# Compile the local binary
+make bin
+
+# Overwrite the cached binary with your local one
+cp terraform-provider-rancher2 ~/.terraform.d/plugins/terraform.example.com/local/rancher2/1.0.0/darwin_amd64/terraform-provider-rancher2
```
stormqueen1990: Suggestion: make version and platform customizable, so people using Apple Silicon and/or Linux can also use it. Example: `"${HOME}/.terraform.d/plugins/terraform.example.com/local/rancher2/${VERSION}/$(uname | tr '[:upper:]' '[:lower:]')_$(arch)"`

stormqueen1990 (commented):
This is awesome! 🎉

scripts/build_local_binary.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Initialize terraform modules
+# terraform init
```
snasovich: Is it intentional to leave these steps commented out? If it's something that may only be needed sometimes, supporting option flags is probably a good idea (not sure if `terraform init / apply` is something we want/need to do as part of script named `build_local_binary.sh`).

scripts/build_local_binary.sh
sbstp: Note: On amd64 `$(arch)` outputs `x86_64` instead of `amd64`. And on M1 Macs I think it will output `aarch64` instead of `arm64`. I'm not familiar with terraform plugins, the suggestion might work fine. But I know rancher uses `amd64` and `arm64` instead of `x86_64` and `aarch64`.

stormqueen1990: Interesting! `uname -m` might be a good alternative. My M1 Mac outputs `arm64` for that, but it does output `aarch64` for Linux (slight difference from Darwin to Linux).

stormqueen1990: It seems `uname -m` also outputs `x86_64` for `amd64` on Linux so I guess I have to update my suggestion to having a `case` statement instead 😞

sbstp: Yeah it's a mess! I don't think it's possible to use a single command, it might be possible to use a one liner with sed like, `$(arch | sed s/x86_64/amd64/ | sed s/aarch64/arm64/)` though.

stormqueen1990: I'd vote on having a case statement in this situation so we can cover more scenarios in the future if needed be. Like the kustomize install script: https://github.com/kubernetes-sigs/kustomize/blob/78978723db943fe5fe4baebd043dd9ab02ab144d/hack/install_kustomize.sh#L108-L124

scripts/build_local_binary.sh
a-blender: I include those commands in my own process. I find automating the process from creating the binary to running an apply is super helpful for the efficiency of local testing! With those commands included, I call this script `./run.sh` But for `build_local_binary.sh` we don't need them no. For option flags, I could add `-i` and `-a` flags for running an additional init and apply.

thatmidwesterncoder (approved):

jakefhyde (approved):

PR #1041, opened 2022-12-05
fix(docs): add 'annotations' property

24601:
The machine_pools object supports adding annotations (required, for example, when using the rancher provider in the cluster autoscaler) but the documentation does not show this. We had to goto the source code to find this and figured others would have the same issue, adding it into the documentation can make it easier for those who do know know Go or are not programmers to find how to annotate the Machine Deployment object when doing tasks like this easily.


jakefhyde (approved):

a-blender (approved):
Thanks @24601 ! Also, if your GitHub handle is a Les Miserable musical reference, that is officially my favorite thing I have seen all week haha

PR #1044, opened 2022-12-06
WIP (Do Not Merge): Generic Auto-Scaling API/Schema & AWS EC2 ASG Implementation

24601:
**What**

The goal of this PR once it is fully baked is to introduce an implementation-agnostic API at the node-pool level to define auto-scaling semantics and implement it in the AWS EC2 provider (so it creates ASGs rather than EC2 nodes directly). 

This PR is an "open early in case anyone wants to help/advise a first time contributor" item and is a work in progress and not being submitted to merge. I have not been able to find a "contributing guidelines" doc for this repo, so if this is not acceptable or outside accepted convention for this repo/Rancher/org, please let me know. The intention here is to share and gain advice from contributors so the work we have done/are doing for our own use on this provider is sourced back in a timely manner, but also, any convention/design for this to code to be useful to others can be advised early before code is ready for merge in the absence of a well codified contributing spec. 

Given that the rancher provider cluster autoscaler provider for the Kubernetes auto scaler provides auto scaling for bare metal based implementations of RKE2 (via the cluster.x-k8s.org CAPI rancher provider), there is no need to address that in our estimation, however, AWS-based clusters created with this provider are our focus to create nodes and ASGs in such a way that cluster autoscaler can be used with the AWS provider pursuant with https://docs.ranchermanager.rancher.io/how-to-guides/new-user-guides/manage-clusters/install-cluster-autoscaler/use-aws-ec2-auto-scaling-groups.

**Why**

We use RKE2 in production across a variety of clouds/bare metal and the lack of easy/straightforward auto-scaling configuration across varying implementations is tough for us. While the rancher provider for the Kubernetes out-of-tree auto-scaler is nice, it does not function/play well with AWS ASGs and the amazonec2 provider in this repo directly creates Nodes in a way that are largely incompatible with autoscaling groups, reducing the usefulness of this provider. Also, the AKS autoscaling is implemented in a relatively AKS-specific manner that specifies scaling preferences at a cluster level, not a node-pool level (you may not want to auto-scale all node pools, for example). 

**How**

The design intent is to surface the common elements of auto-scaling in an implementation non-specific schema while providing appropriate escape hatches for provider-specific features or idiosyncrasies at a node-pool level with minimal changes to the API (e.g. additive) and the least trauma to the codebase to facilitate a separate branch indefinitely in case changes are not accepted into the main repo.


kkaempf:
closing stale PRs.

PR #1055, opened 2023-01-09
Add example for authenticated cluster level private registry

HarrisonWAffel:
### Problem: 
The terraform documentation does not provide clear examples on how to properly configure cluster level private registries from which core system images should be pulled from. This causes a problem, as the process of properly configuring these registries for rancher v2 cluster resources is not as straight forward as it is for RKE1 clusters, leading to confusion. 

### Solution: 
Add additional documentation on how to properly configure an authenticated cluster level registry using terraform. Add additional text around the example to also explain requirements for unauthenticated cluster level registries (simply omit values if not needed).

### DIFF
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -207,6 +207,36 @@ EOF
 }
 ```
 
+### Creating Rancher V2 cluster using a cluster level authenticated `system-default-registry`
+
+```hcl
+resource "rancher2_cluster_v2" "foo_cluster_v2" {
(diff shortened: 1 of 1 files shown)


a-blender (approved):
Super helpful, thank you! LGTM

jiaqiluo (approved):

matteogazzadi:
Sorry but example is not clear. 
How can you specify the name of a Secret in the downstream cluster while you are creating the cluster itself ?

I mean in RKE1, in the private_registries section is possible to specify url, username and password

Issue #1052, opened 2023-01-06
Terraform build fails with Rancher dependency v2.7


HarrisonWAffel:
Terraform fails to build when using Rancher commits from 2.7, but is able to build when using commits from 2.6. It seems some types were moved around in 2.7, breaking the following structure files: 

`structure_cluster_logging.go` `structure_cluster_scan.go` `structure_logging_custom_target_config.go`

Builds will fail with the following error 

```
# github.com/rancher/terraform-provider-rancher2/rancher2
rancher2/structure_cluster_logging.go:12:73: undefined: client.ClusterLogging
rancher2/structure_cluster_logging.go:153:71: undefined: client.ClusterLogging
rancher2/structure_cluster_scan.go:10:55: undefined: client.CisScanConfig
rancher2/structure_cluster_scan.go:31:52: undefined: client.ClusterScanConfig
rancher2/structure_cluster_scan.go:42:70: undefined: client.ClusterScan
rancher2/structure_cluster_scan.go:78:68: undefined: client.CisScanConfig
rancher2/structure_cluster_scan.go:108:65: undefined: client.ClusterScanConfig
rancher2/structure_cluster_scan.go:121:67: undefined: client.ClusterScan
rancher2/structure_logging_custom_target_config.go:9:60: undefined: client.CustomTargetConfig
rancher2/structure_logging_custom_target_config.go:42:74: undefined: client.CustomTargetConfig
rancher2/structure_logging_custom_target_config.go:42:74: too many errors
```

This error seems to be due to the fact that `"github.com/rancher/rancher/pkg/client/generated/management/v3"` package no longer contains the following types

`CustomTargetConfig`, `ClusterLogging`, `ClusterScan`, `CisScanConfig`, and `ClusterScanConfig`

The resolution of this issue may be as simple as fixing import paths, but could also be more involved depending on the changes that these types have undergone in 2.7.

HarrisonWAffel:
Root causes: 

https://github.com/rancher/rancher/pull/39656

https://github.com/rancher/rancher/issues/37318

a-blender:
**Root cause**

The [root cause](https://github.com/rancher/terraform-provider-rancher2/issues/1052#issuecomment-1374159354) of this issue is that Logging and cis v1 scan code support were removed in rancher 2.7 but not in the TF rancher2 provider so the provider is trying to reference types that don't exist.

**Design discussion summary**

[...]

PR #1056, opened 2023-01-12
Add k8s 1.25 support | Remove logging cis v1 support

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1052

 
## Problem

 
The [root cause](https://github.com/rancher/terraform-provider-rancher2/issues/1052#issuecomment-1374159354) of this issue is that Logging and cis v1 scan code support were removed in rancher 2.7 but not in the TF rancher2 provider so the provider is trying to reference types that don't exist.

## Solution

I have branched the TF provider, per [this discussion](https://github.com/rancher/terraform-provider-rancher2/issues/1052#issuecomment-1380685936) into release/v2 for rancher 2.6 and still `master` for rancher 2.7. To fix the TF build, I am doing the following in this PR

* Bump go to 1.19
* Bump rancher/rancher dependency to `v2.7-head` commit
* Bump k8s packages to 1.25
* Remove logging and cis v1 support from code, tests, and docs

### Manual Testing

Tested this by creating an Amazon EC2 cluster with k8s 1.25.5 on Rancher `2.7-head` instance provisioned with my local copy of the Terraform rancher2 provider. Verified cluster came up active and was healthy.


HarrisonWAffel (approved):
Looks good, thanks for splitting changes into separate commits

jakefhyde (approved):

PR #1058, opened 2023-01-16
Fix intermittent 409 Conflict when creating new rancher2_project w/ PSPTID

mouellet:
When creating a new `rancher2_project` w/ `pod_security_policy_template_id`, you'll sometimes get the following error:
```
Error: Bad response statusCode [409]. Status [409 Conflict]. Body: [baseType=error, code=Conflict, message=error updating PSPT ID] from [https://localhost/v3/projects/c-rpwfk:p-t4f4x?action=setpodsecuritypolicytemplate]
```

As explained [here](https://github.com/rancher/dashboard/blob/14286fd7a856a396f46fe71665385d0bccf93f16/shell/models/management.cattle.io.project.js#L109-L116), the Project resource is updated right after it's created, some controllers must be watching this resource. And as shown in https://github.com/rancher/rancher/issues/22834#issuecomment-766100209,  409 is returned when calling `ActionSetpodsecuritypolicytemplate` with an outdated resource version.

This PR refreshes the project resource, updating the resource version, before calling `ActionSetpodsecuritypolicytemplate` and retries if needed.

### DIFF
--- a/rancher2/resource_rancher2_project.go
+++ b/rancher2/resource_rancher2_project.go
@@ -98,12 +98,22 @@ func resourceRancher2ProjectCreate(d *schema.ResourceData, meta interface{}) err
 		pspInput := &managementClient.SetPodSecurityPolicyTemplateInput{
 			PodSecurityPolicyTemplateName: pspID,
 		}
-		_, err = client.Project.ActionSetpodsecuritypolicytemplate(newProject, pspInput)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


a-blender (approved):
I'd prefer a comment above this block or above line 102 stating that we are trying to refresh the project resource before updating the PST to avoid conflict errors, to make it clear why we are retrying at that point. Otherwise LGTM thanks for putting up this PR

HarrisonWAffel (approved):

PR #1060, opened 2023-01-17
Update docs with branching strategy and compatibility matrix

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1052

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,14 +84,29 @@ Due to [network limitation](https://docs.docker.com/docker-for-mac/networking/#k
 $ EXPOSE_HOST_PORTS=true make docker-testacc
 ```
 
+Branching the Provider
+---------------------------
+
+The provider is branched into two release lines that have major version alignment with Rancher 2.6 and 2.7. The `release/v2` branch with 2.0.0+ is aligned with Rancher 2.6 and `master` with 3.0.0+ is aligned with Rancher 2.7. Terraform fixes and new features will be available on `master` but only bug fixes will be backported to `release/v2.`
+
+Aligning major provider releases with minor Rancher releases means,
+
+* We can follow semver 
+* We can cut patch/minor versions on an as-needed basis to fix bugs or add new resources 
+* We have 'out of band' flexibility and are only tied to releasing a new version of the provider when we get a new 2.x Rancher minor version.
+
+See the [compatibility matrix](https://github.com/rancher/terraform-provider-rancher2/tree/master/docs/compatibility-matrix.md) for details.
+
+If you are using Terraform to provision clusters on instances of Rancher 2.6 and 2.7, you must have a separate configuration in a separate dir for each provider. Otherwise, Terraform will overwrite the `.tfstate` file every time you switch versions.
+
 Releasing the Provider
 ---------------------------
 
-The provider should be released 'out of band' from Rancher, but can be loosely tied to a Rancher release to track issues. For example: Terraform v2.6.9 should be released a few days after Rancher v2.6.9 and Terraform fixes or features in the release are said to be included in Rancher v2.6.9 (and will work if you provision with Terraform).
+As of Terraform 2.0.0 and 3.0.0, the provider is tied to Rancher minor releases but can be released 'out of band' within that minor version. For example, 3.0.0 will be released a few days after Rancher 2.7.x and fixes and features in the 3.0.0 release will be supported for clusters provisioned via Terraform on Rancher 2.7.x. A critical bug fix can be released 'out of band' as 3.0.1 and backported to `release/v2` as 2.0.1. A new feature can also be released 'out of band' as 3.1.0 but not backported. Terraform 4.0.0 must be released with Rancher 2.8.
 
-The [RKE provider](https://github.com/rancher/terraform-provider-rke) should be released after every RKE/KDM release. For example: If RKE v1.3.15 was released, bump the RKE version to v1.3.15 and release the provider.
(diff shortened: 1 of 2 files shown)


HarrisonWAffel (approved):

snasovich (changes requested):
I will have some suggestions for this later, so just marking with "Request changes" for now to block merging.

docs/compatability-matrix.md
```diff
@@ -0,0 +1,15 @@
+# Rancher2 Provider
```
snasovich: There is a typo in fule name

docs/compatability-matrix.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Version compatibility matrix
+
+#### Rancher 2.6
+
+| Terraform Version | 2.0.0         | 2.x.x  | 2.x.x         |
```
snasovich: I think we want to transpose this table and list TF release versions as rows and list min/max Rancher versions as columns. With the current approach, we're adding a new column for each TF provider release which eventually will make a very wide table.

snasovich (changes requested):
Couple initial comments to address, will need to review again later.

docs/compatability-matrix.md
snasovich: Discussed with @annablender offline and we've agreed on a more flexible approach where we list "reference" Racher<->TF provider pairs and make disclaimers of using combinations where TF provider version is higher than reference Rancher version and vice-versa.

README.md
```diff
@@ -84,14 +84,29 @@ Due to [network limitation](https://docs.docker.com/docker-for-mac/networking/#k
 $ EXPOSE_HOST_PORTS=true make docker-testacc
 ```
 
+Branching the Provider
+---------------------------
+
+The provider is branched into two release lines that have major version alignment with Rancher 2.6 and 2.7. The `release/v2` branch with 2.0.0+ is aligned with Rancher 2.6 and `master` with 3.0.0+ is aligned with Rancher 2.7. Terraform fixes and new features will be available on `master` but only bug fixes will be backported to `release/v2.`
```
snasovich: Don't want to necessarily commit to backporting all fixes, so adding a bit of a disclaimer. Also minor formatting fix.

```suggestion
The provider is branched into two release lines that have major version alignment with Rancher 2.6 and 2.7. The `release/v2` branch with 2.0.0+ is aligned with Rancher 2.6 and `master` with 3.0.0+ is aligned with Rancher 2.7. Terraform provider fixes and new features will be available on `master` but only bug fixes will be backported to `release/v2` as necessary.
```

README.md
```diff
@@ -84,14 +84,29 @@ Due to [network limitation](https://docs.docker.com/docker-for-mac/networking/#k
...
+
+Aligning major provider releases with minor Rancher releases means,
+
+* We can follow semver 
+* We can cut patch/minor versions on an as-needed basis to fix bugs or add new resources 
+* We have 'out of band' flexibility and are only tied to releasing a new version of the provider when we get a new 2.x Rancher minor version.
+
+See the [compatibility matrix](https://github.com/rancher/terraform-provider-rancher2/tree/master/docs/compatibility-matrix.md) for details.
```
snasovich: Suggesting to use relative link (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes#relative-links-and-image-paths-in-readme-files) here as it's more resilient to changes to URLs and will work if we backport this docs change to another branch without needing to change it (unless we don't want to port actual `compatibility-matrix.md` file and always point to it in the master 🤷 ).

docs/compatibility-matrix.md
```diff
@@ -0,0 +1,23 @@
...
+#### Rancher 2.7
+
+| Terraform provider version | Rancher |
+|----------------------------------------|:-------:|
+| 3.0.0                                  | 2.7.2   |
+| 3.0.1 (3.1.0 for features)             | 2.7.x   |
+
+**Can I use an earlier Terraform version?** Yes, but Terraform may not support all features and fields supported in your Rancher version so provisioning may be limited.
+
+**Can I use a later Terraform version?** Yes, but you must NOT use any new features and fields that your Rancher version does not support.
```
snasovich, line 23: Nice, I like this wording!

Maybe add a section like "FAQ" above it so these remarks are technically in their own section and not with "Rancher 2.7"?

docs/compatibility-matrix.md
```diff
@@ -0,0 +1,23 @@
...
+### Version compatibility matrix
+
+The version matrix specifies the Terraform provider version _recommended_ to use with the associated minor Rancher version that it was released for. When updating the version matrix, add a row for each Terraform version released with a minor Rancher release.
+
+#### Rancher 2.6
+
+| Terraform provider version | Rancher |
+|----------------------------------------|:-------:|
+| 2.0.0                                  | 2.6.11  |
+| 2.0.1 (2.1.0 for features)             | 2.6.x   |
```
snasovich: Recommending to not list not-yet-released versions, only the ones that we already released. I understand it's an example, but don't want it to be misleading. The same applies to the 2.7 table below.

snasovich (changes requested):
Looks good, some minor comments to address and it should be good to go from my point of view.

docs/compatibility-matrix.md
a-blender: Do you mean to remove the entire second row here or just modify to list

`2.0.x (2.x.0 for features)`

snasovich: Just remove, we explain the approach to versioning in Readme already. This matrix is for released(to be released) versions.

snasovich (approved):

PR #1062, opened 2023-01-18
[2.7] Add missing implementation of Azure features for RKE2

HarrisonWAffel:
### Problem:
This PR addresses two issues with the initial implementation of various Azure features within the provider.

1. The expanders and flatteners for the `v2` cluster were not implemented. The schema elements were added however they were not used when crafting API requests to the Rancher API. This has been fixed and smoke tested, so these features should work on RKE2 clusters now. RKE1 implementation is complete and not missing these expanders and flatteners. 

2. The JSON annotation used for the `UsePublicIpStandardSKU` fields have been modified to match the argument name expected by Rancher machine. Previously the JSON Annotation matched the field name, however this resulted in the Rancher API improperly converting this JSON annotation into an unexpected rancher-machine argument.  

### Solution

Add missing expanders and flatteners for v2 clusters

Modify the JSON annotation for the `UsePublicIPStandardSKU` field to match the expected argument name in rancher-machine. This was updated for both RKE1 and RKE2 clusters, as both distributions craft rancher-machine arguments in the same way. 

### Testing

+ Create an RKE2 cluster via terraform on Azure, ensure that the newly added fields are defined in the `azure_config`. 

### Engineer testing

I've created an RKE2 cluster with all of these properties enabled and confirmed that it provisioned properly and became active. I also confirmed that these features were enabled in the Azure Portal's UI. 

### QA considerations

1. When testing Availability Zones ensure you have the following three properties defined in your `azure_config`
   `use_public_ip_standard_sku = true`
   `static_public_ip = true`
   `managed_disks = true`
2. Azure does not allow tags to be prefixed with the following reserved keywords, `azure`, `microsoft`, and `windows`.

### DIFF
--- a/rancher2/schema_node_template_azure.go
+++ b/rancher2/schema_node_template_azure.go
@@ -42,7 +42,7 @@ type azureConfig struct {
 	Tags                   string   `json:"tags,omitempty" yaml:"tags,omitempty"`
 	AcceleratedNetworking  bool     `json:"acceleratedNetworking,omitempty" yaml:"acceleratedNetworking,omitempty"`
 	AvailabilityZone       string   `json:"availabilityZone,omitempty" yaml:"availabilityZone,omitempty"`
-	UsePublicIPStandardSKU bool     `json:"usePublicIPStandardSKU,omitempty" yaml:"usePublicIPStandardSKU,omitempty"`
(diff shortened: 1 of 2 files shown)


a-blender (approved):
LGTM thank you!

jiaqiluo (approved):
LGTM

PR #1070, opened 2023-02-08
Update SDK and make machine pool cloud_credential_secret_name optional to fix plan bug

a-blender:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/835
 
## Problem

When creating an RKE2 cluster via Terraform on any hosted provider (Amazon EC2, Azure, Linode driver so far), Terraform computes a new value for a duplicate field `cloud_credential_secret_name` in the machine pool and then throws an error on a terraform apply pertaining to that new value. It works on a second run of terraform apply but is frustrating and widely seen by many customers.

## Solution

That field currently exists in both `rancher_cluster_v2` and `rancher_cluster_v2.machine_pool`. I couldn't find docs on why the cloud credential secret was added to the machine pool. Rancher pulls the cloud credential secret from the [RKECommonNodeConfig](https://github.com/rancher/rancher-operator/blob/df96e4e74f8ee0dc4f6861a00ad9a26127e8a675/pkg/apis/rke.cattle.io/v1/machine.go#L8) and sets it on the newly created machine [here](https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go#L168-L173) as well as the cluster management Spec but doesn't do anything further with it. This looks like a legacy-related change from [this PR](https://github.com/rancher/rancher-operator/commit/df96e4e74f8ee0dc4f6861a00ad9a26127e8a675) but I don't fully know the context of the design decision here.

Rancher allows you to mix and match secrets, possibly to support the use case where a customer has machine pools in separate AWS/cloud accounts. The TF provider rancher2 _must_ have parity with Rancher so this field cannot be removed (as I originally thought) because that would cause a regression.

Update 2/10: I believe this is a bug/compatibility issue in the Terraform plugin SDK. See comment https://github.com/rancher/terraform-provider-rancher2/issues/835#issuecomment-1426491075 for further details.

This PR has the following fix
* Update `machine_pool.cloud_credential_secret_name` to be Optional. This keeps parity with Terraform and fixes the plan bug
* Update docs
* Update Terraform plugin SDK to 1.17.2
 
### Manual Testing

Tested with RKE2 clusters on Amazon EC2. Each cluster provisions successfully on the first try of terraform apply. If you do not specify a cloud credential, the cluster hangs with `Waiting for viable init node` trying to create the machine. I am open to discussion on this behavior.

Testing

- [x] only machine_pool.cloud_credential_secret_name
- [x] only global cloud_credential_secret_name
- [x] both global cloud_credential_secret_name and machine_pool.cloud_credential_secret_name
- [x] no cc (cannot create nodes as expected)


jakefhyde (approved):

HarrisonWAffel (approved):
The code change looks fine, though the PR description talks about docs updates which I don't see yet

jakefhyde (approved):

PR #1082, opened 2023-03-01
Add script for QA to test Terraform rcs locally

a-blender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1083

### DIFF
--- a/setup-provider.sh
+++ b/setup-provider.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# QA has a new process to test Terraform using rcs. We will not publish rcs
+# onto the Terraform registry because Hashicorp could potentially block our
+# testing by being slow to publish.
+
+# Instead, we will test using a downloaded binary from the rc. This script
+# sets up the correct binary using a defined <provider> <version> to test
+# updates locally.
+
+# Example
+# ./setup-provider.sh rancher2 v3.0.0-rc1
+
+# Validate user input
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo -e "Please specify a Terraform provider and version tag."
+    echo -e "./setup-provider.sh <provider> <version>"
+    exit 1
+fi
+
+# Set global vars
+
+PROVIDER=$1
+VERSION=$2
+VERSION_TAG=$(echo $2 | cut -c 2-)
+BASE=~/.terraform.d/plugins/terraform.example.com/local/${PROVIDER}/${VERSION_TAG}
+
+# Install gzip
+
+brew install gzip
+
+# Pull and copy the binary
+
+case $OSTYPE in
+
+  "linux-gnu"*)
+    curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_linux_amd64.zip | gunzip -c - > terraform-provider-${PROVIDER}_${VERSION}
+    if [ ! $? -eq 0 ]; then
+      echo "Error: incomplete download. Did you specify the correct provider and version?"
+      rm terraform-provider-${PROVIDER}_${VERSION}
+      exit 1
+    fi
+    DIR=${BASE}/linux_amd64
+    mkdir -p $DIR && cp terraform-provider-${PROVIDER}_${VERSION} $DIR/terraform-provider-${PROVIDER}
+    ;;
+
+  "darwin"*)
+    curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_darwin_amd64.zip | gunzip -c - > terraform-provider-${PROVIDER}_${VERSION}
+    if [ ! $? -eq 0 ]; then
+      echo "Error: incomplete download. Did you specify the correct provider and version?"
+      rm terraform-provider-${PROVIDER}_${VERSION}
+      exit 1
+    fi
+    DIR=${BASE}/darwin_amd64
+    mkdir -p $DIR && cp terraform-provider-${PROVIDER}_${VERSION} $DIR/terraform-provider-${PROVIDER}
+    ;;
+esac
+
+# Clean up
+
+rm terraform-provider-${PROVIDER}_${VERSION}
+
+echo -e "Terraform provider ${PROVIDER} ${VERSION} is ready to test!
+Please update the required_providers block in your Terraform config file. Example:
+
+terraform {
+  required_providers {
+    rancher2 = {
+      source = "terraform.example.com/local/rancher2"
+      version = "3.0.0-rc1"
+    }
+  }
+}"
--- a/setup-provider.sh
+++ b/setup-provider.sh
@@ -11,12 +11,13 @@
 # Example
 # ./setup-provider.sh rancher2 v3.0.0-rc1
 
+set -e 
+
 # Validate user input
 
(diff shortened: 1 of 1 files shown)


setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+    curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_darwin_amd64.zip | gunzip -c - > terraform-provider-${PROVIDER}_${VERSION}
+    if [ ! $? -eq 0 ]; then
+      echo "Error: incomplete download. Did you specify the correct provider and version?"
+      rm terraform-provider-${PROVIDER}_${VERSION}
+      exit 1
+    fi
+    DIR=${BASE}/darwin_amd64
+    mkdir -p $DIR && cp terraform-provider-${PROVIDER}_${VERSION} $DIR/terraform-provider-${PROVIDER}
+    ;;
+esac
```
jakefhyde:
```suggestion
OS_PLATFORM=$(uname -sp | tr '[:upper:] ' '[:lower:]_' | sed 's/x86_64/amd64/')
DIR=~/.terraform.d/plugins/terraform.local/local/${PROVIDER}/${VERSION_TAG}/${OS_PLATFORM}/terraform-provider-${PROVIDER}_${VERSION}

curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_${OS_PLATFORM}.zip | gunzip -c - > ${DIR}/terraform-provider-${PROVIDER}_${VERSION}
```

setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  "darwin"*)
+    curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_darwin_amd64.zip | gunzip -c - > terraform-provider-${PROVIDER}_${VERSION}
+    if [ ! $? -eq 0 ]; then
+      echo "Error: incomplete download. Did you specify the correct provider and version?"
+      rm terraform-provider-${PROVIDER}_${VERSION}
+      exit 1
+    fi
+    DIR=${BASE}/darwin_amd64
+    mkdir -p $DIR && cp terraform-provider-${PROVIDER}_${VERSION} $DIR/terraform-provider-${PROVIDER}
```
jakefhyde: What is the purpose of this directory? Why not download it directly there?

setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+# Example
+# ./setup-provider.sh rancher2 v3.0.0-rc1
+
+# Validate user input
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo -e "Please specify a Terraform provider and version tag."
+    echo -e "./setup-provider.sh <provider> <version>"
+    exit 1
+fi
```
jakefhyde:
```suggestion
if [ $# -ne 2 ]; then
  echo "Usage: $0 <provider> <version>"
  exit 1
fi
```

setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Instead, we will test using a downloaded binary from the rc. This script
+# sets up the correct binary using a defined <provider> <version> to test
+# updates locally.
+
+# Example
+# ./setup-provider.sh rancher2 v3.0.0-rc1
+
+# Validate user input
+
```
jakefhyde, line 17:
```suggestion
set -e
```
This is just a personal preference but I prefer scripts to auto fails with errors.

setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+# Set global vars
+
+PROVIDER=$1
+VERSION=$2
+VERSION_TAG=$(echo $2 | cut -c 2-)
+BASE=~/.terraform.d/plugins/terraform.example.com/local/${PROVIDER}/${VERSION_TAG}
+
+# Install gzip
+
+brew install gzip
```
jakefhyde:
```suggestion
if ! command -v "gzip" &> /dev/null; then
  echo "Missing gzip"
  exit 1
fi
```

setup-provider.sh
```diff
@@ -0,0 +1,74 @@
...
+Please update the required_providers block in your Terraform config file. Example:
+
+terraform {
+  required_providers {
+    rancher2 = {
+      source = "terraform.example.com/local/rancher2"
+      version = "3.0.0-rc1"
+    }
+  }
+}"
```
jakefhyde: This is also an option: https://pet2cattle.com/2021/10/terraform-override-providers, might be easier depending on what QA is going to test with so they won't have to update any of their existing terraform scripts to use the local one.

jakefhyde (dismissed):

setup-provider.sh
markusewalker: Bit of a nit. Can we add a conditional to check the if it's already installed? If not, then install `gzip`. I am assuming QA will use a VM with this script (if not, different story lol), but in the scenario they don't, this may be an issue. Some members use Windows as their local, so this installation and the case statement below is an issue

I don't see it being an issue, just calling it out.

markusewalker: Agreed, I left a similar comment on this below.

setup-provider.sh
markusewalker: I would prefer this as well, I don't think we necessarily need a new directory and the one-liner for `OS_PLATFORM` should cover darwin and linux distros. QA potentially could use Windows, but I would not expect it.

setup-provider.sh
a-blender: This dir in `~/.terraform.d` may not exist on their local system until they create the path for the tag they are trying to test with so I created it first

jakefhyde: Yeah but why not `mkdir -p` the directory as part of the script?

markusewalker (approved):
lgtm

setup-provider.sh
a-blender: True! I discussed with QA and tried out that option, and there's slightly more preference to update the `source` line in main.tf since there's only one per repo where Terraform tracks the state and they will be updating that file anyway for testing. @markusewalker said it's fine as long as there's docs on how to use the script which are included in the file here. I've also updated the path to `terraform.local/local/rancher2` since that appears to be the standard way to reference a local binary.

setup-provider.sh
```diff
@@ -0,0 +1,62 @@
...
+then
+  OS_PLATFORM="linux_amd64"
+fi
+
+DIR=~/.terraform.d/plugins/terraform.local/local/${PROVIDER}/${VERSION_TAG}/${OS_PLATFORM}
+mkdir -p $DIR
+curl -sfL https://github.com/rancher/terraform-provider-${PROVIDER}/releases/download/${VERSION}/terraform-provider-${PROVIDER}_${VERSION_TAG}_${OS_PLATFORM}.zip | gunzip -c - > ${DIR}/terraform-provider-${PROVIDER}
+
+echo -e "Terraform provider ${PROVIDER} ${VERSION} is ready to test!
+Please update the required_providers block in your Terraform config file
```
slickwarren, line 53: not necessary for now, but would be cool to have an optional variable to add this to the user's config file for them. i.e.
```
usage: $0 <provider> <version> <terraform_config_path>
```

slickwarren (approved):

Josh-Diamond (approved):
LGTM - ✅

PR #1089, opened 2023-03-23
Remove macAddress in Harvester node driver docs

futuretea:
This PR removes the description of MAC addresses from the document

## Issue: 

  https://github.com/harvester/harvester/issues/3698

## Problem

 The document contains a description of MAC addresses. If users are not careful, they may configure MAC addresses that conflict with each other
## Solution

 temporarily remove the description of MAC addresses from the document.
## Testing

No，Just document change


guangbochen (approved):
LGTM, this is just a doc change for the Harvester provider, @a-blender  can we PTAL and merge it, thanks.

a-blender (approved):

PR #1092, opened 2023-03-26
doc: Change attribute name for resource ID

Wykiki:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1104
 
## Problem

I tried to use the attribute `cluster_id`, but the provider complained it didn't exists.
 
## Solution

I tried with `id`, and it worked as expected.
 
## Testing

I think it's possible to update the TestCase for the resource, but I don't have any knowledge about it.

### Manual Testing

Change `rancher2_cluster_v2.resource.cluster_id` to `rancher2_cluster_v2.resource.id`
Run `terraform validate`

### DIFF
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -674,7 +674,7 @@ The following attributes are exported:
 
 #### Attributes
 
-* `cluster_id` - (Computed) Cluster ID (string)
+* `id` - (Computed) Cluster ID (string)
 * `name` - (Computed) Name of cluster registration token (string)
 * `command` - (Computed) Command to execute in a imported k8s cluster (string)
 * `insecure_command` - (Computed) Insecure command to execute in a imported k8s cluster (string)


a-blender (approved):

a-blender (commented):
@Wykiki Thank you for this community PR. A few things,

* Is there an open GH issue that this resolves?
* If not, can you please provide repro steps and your main.tf to reproduce this bug in a separate issue?

If we're going to update the `cluster_v2` docs, they need to match the schema, and the schema shows two different fields `id`and `cluster_id` for clusterRegistrationToken https://github.com/rancher/terraform-provider-rancher2/blob/f443d110371fa93bc3b12005c1b5c02e174b6517/rancher2/schema_cluster.go#L36-L45. It looks like `cluster_id` is a duplicate and possibly needs to be removed but I'd recommend more investigation on this in case both those fields are there for a reason.

HarrisonWAffel:
Could we get an example of how you are trying to use this resource? I would be surprised if the schema included duplicate fields like this, not that it can't happen but without an example I don't think we can justify changing the docs or schema

Wykiki:
Everything should be good, tell me if I missed something, thanks !

a-blender:
Docs were misread as per [this comment](https://github.com/rancher/terraform-provider-rancher2/issues/1104#issuecomment-1511409259). Closing this PR.

Issue #1069, opened 2023-02-07
[Backport release/v2] No way to configure vmAffinity for Harvester clusters 


a-blender:
Backport for https://github.com/rancher/terraform-provider-rancher2/issues/1009

### Is your feature request related to a problem? Please describe.

Maybe I'm missing the config somewhere, but I've looked all over and can't find a way to configure node scheduling rules for Harvester deployments. Here's a screenshot from the UI of what I'm trying to configure:

<img width="1744" alt="image" src="https://user-images.githubusercontent.com/5328577/197367874-83ce2c5c-e088-4f3b-87c0-1562eed7a416.png">

I was expecting to find a spot to configure this in `rancher2_machine_config_v2`, since it seems this setting gets applied to the property `vmAffinity` for the resulting `HarvesterConfig`.  E.g. value:

```
vmAffinity: eyJub2RlQWZmaW5pdHkiOnsicmVxdWlyZWREdXJpbmdTY2hlZHVsaW5nSWdub3JlZER1cmluZ0V4ZWN1dGlvbiI6eyJub2RlU2VsZWN0b3JUZXJtcyI6W3sibWF0Y2hFeHByZXNzaW9ucyI6W3sia2V5IjoidG9wb2xvZ3kua3ViZXJuZXRlcy5pby96b25lIiwib3BlcmF0b3IiOiJJbiIsInZhbHVlcyI6WyJ5eXoyIl19XX1dfX19
```

which decodes to
```
{"nodeAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"topology.kubernetes.io/zone","operator":"In","values":["yyz2"]}]}]}}}
```

This is important so I can ensure nodes deploy across availability zones. Right now, they're all going into one. Thanks for any help!

### Describe the solution you'd like

`vmAffinity` field added to the Terraform rancher2 provider.

a-blender:
See test plan for original issue https://github.com/rancher/terraform-provider-rancher2/issues/1009#issuecomment-1435309647.

a-blender:
Blocked -- waiting on Terraform 2.0.0 for Rancher v2.6.x.

snasovich:
Same as https://github.com/rancher/terraform-provider-rancher2/issues/1009#issuecomment-1437795825

a-blender:
[...]

PR #1102, opened 2023-04-07
[release/v2] Harvester vmAffinity feature revert

a-blender:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1069
 
## Problem

 
This is to fix an oversight. vmAffinity was a community PR that is a _feature_ and we are only releasing new TF features in `master` which is the Rancher 2.7 line. After this is merged and a new rc is cut, we can release TF 2.0.0 for 2.6.11 and users should not use vmAffinity on Rancher 2.6 via tf because it is not supported.


HarrisonWAffel (approved):
Because these changes have only been included in RC versions this schema change LGTM

snasovich (approved):

PR #1113, opened 2023-05-03
Fix 'unexpected end of JSON input' error setting pspt on new project

mouellet:
## Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1127

When creating a new 'rancher2_project' and specifying the 'pod_security_policy_template_id' attribute, we get this error:
```
│ Error: unexpected end of JSON input
│ 
│   with rancher2_project.this,
│   on main.tf line 1, in resource "rancher2_project" "this":
│    1: resource "rancher2_project" "this" {
```
 
## Problem

With rancher/terraform-provider-rancher2#1058, I removed skipping some API error on ActionSetpodsecuritypolicytemplate (sorry, my bad! 😅).

## Solution

Partial revert
 
### Manual Testing

Manual test successful on internal stack.

### Automated Testing

Couldn't add automated tests as the cluster used in CI doesn't have pod security policies enabled, which is fine since they are deprecated anyway...

### DIFF
--- a/rancher2/resource_rancher2_project.go
+++ b/rancher2/resource_rancher2_project.go
@@ -108,12 +108,15 @@ func resourceRancher2ProjectCreate(d *schema.ResourceData, meta interface{}) err
 				if IsConflict(err) || IsForbidden(err) {
 					return resource.RetryableError(err)
 				}
-				return resource.NonRetryableError(err)
+				// Checking error due to ActionSetpodsecuritypolicytemplate() issue
+				if error.Error(err) != "unexpected end of JSON input" {
+					return resource.NonRetryableError(err)
+				}
 			}
 			return nil
 		})
 		if err != nil {
-			return err
+			return fmt.Errorf("[ERROR] waiting for pod_security_policy_template_id (%s) to be set on project (%s): %s", pspID, newProject.ID, err)
 		}
 	}
 


rancher2/resource_rancher2_project.go
```diff
@@ -108,12 +108,15 @@ func resourceRancher2ProjectCreate(d *schema.ResourceData, meta interface{}) err
...
+				// Checking error due to ActionSetpodsecuritypolicytemplate() issue
+				if error.Error(err) != "unexpected end of JSON input" {
+					return resource.NonRetryableError(err)
+				}
 			}
 			return nil
 		})
 		if err != nil {
-			return err
+			return fmt.Errorf("[ERROR] waiting for pod_security_policy_template_id (%s) to be set on project (%s): %s", pspID, newProject.ID, err)
```
snasovich, line 119: Would generally prefer `%w` used for wrapped errors but then I see it's not used throughout the file so should leave it "as is". https://stackoverflow.com/a/61287626
It's also weird that error is prefixed with `[ERROR]` which doesn't seem to make sense (it makes sense for `log.Printf`) but again it's an established practice in this file - so going for consistency.

snasovich (approved):
Looks like a proper way of reintroducing the logic that got inadvertently removed by https://github.com/rancher/terraform-provider-rancher2/pull/1058

@a-blender @HarrisonWAffel , FYI - please review too.

HarrisonWAffel (approved):

Issue #1003, opened 2022-10-12
Intermittently imports of EKS clusters never finish


tulanian:
**Versions**
- Rancher version: 2.6.8
- Rancher Terraform provider: 1.24.0
- Terraform: 1.2.2

**Information about the Cluster**
- Kubernetes version: 1.21
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Hosted EKS

**Describe the bug**

Sometimes importing an EKS cluster will never complete (saying "Still creating..." for 30 min then time-out), but the cluster is active in the Rancher instance.  Other times it finishes in seconds. 
**To Reproduce**

Using this code to import the cluster. The aws-auth configMap has already been updated with the user referred to by the cloud_credential. 
```
resource "rancher2_cloud_credential" "this" {
  name        = var.name_prefix
  description = "Credentials used for managing ${var.name_prefix}"
  amazonec2_credential_config {
    access_key = aws_iam_access_key.rancher.id
    secret_key = aws_iam_access_key.rancher.secret
  }
}

resource "rancher2_cluster" "imported_eks_cluster" {
  name        = var.cluster_id
  description = "Terraform EKS cluster"
  eks_config_v2 {
    cloud_credential_id = rancher2_cloud_credential.this.id
    name                = var.cluster_id
    region              = var.region
    imported            = true
  }
}
```
**Result**
Sometimes this happens until the time-out but the cluster is active in Rancher:
```
module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster: Still creating... [10m40s elapsed]
module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster: Still creating... [10m50s elapsed]
module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster: Still creating... [11m0s elapsed]
module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster: Still creating... [11m10s elapsed]
...
Error: [ERROR] waiting for cluster (c-xfbkg) to be created: timeout while waiting for state to become 'pending' (last state: 'active', timeout: 30m0s)
│ 
│   with module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster,
│   on .terraform/modules/import_to_rancher/main.tf line 27, in resource "rancher2_cluster" "imported_eks_cluster":
│   27: resource "rancher2_cluster" "imported_eks_cluster" {

[...]

PR #1114, opened 2023-05-04
Consider all possible cluster states before passing them to StateChangeConf

furkatgofurov7:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1003
https://github.com/rancher/eks-operator/issues/84
 
## Problem

When importing an EKS cluster using the terraform provider, sometimes rancher cluster reconciliation loop can be fast and go ahead with cluster creation (going from `"pending"` to `"active"` too fast) and the provider can miss it, resulting on provider waiting for the cluster to be in the "pending" state even though it went pass that state long before and provider simply missed that
 
## Solution

Instead of predefining the `expectedState` beforehand and overwriting it later if need to be, define it as slice literal with keeping the same state ("active") and append the new states later if need to be (i.e.` if cluster.Driver == clusterDriverImported || (cluster.Driver == clusterDriverEKSV2 && cluster.EKSConfig.Imported)` == `True`) and pass it to   
`StateChangeConf` struct. Anyways, StateChangeConf struct expects `Target` to be `[]string` so nothing is changed behaviour wise but just all possible cluster states are considered before passing them.

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -129,14 +129,14 @@ func resourceRancher2ClusterCreate(d *schema.ResourceData, meta interface{}) err
 
 	log.Printf("[INFO] Creating Cluster %s", cluster.Name)
 
-	expectedState := "active"
+	expectedState := []string{"active"}
 
 	if cluster.Driver == clusterDriverImported || (cluster.Driver == clusterDriverEKSV2 && cluster.EKSConfig.Imported) {
-		expectedState = "pending"
+		expectedState = append(expectedState, "pending")
 	}
 
 	if cluster.Driver == clusterDriverRKE || cluster.Driver == clusterDriverK3S || cluster.Driver == clusterDriverRKE2 {
-		expectedState = "provisioning"
+		expectedState = append(expectedState, "provisioning")
 	}
 
 	// Creating cluster with monitoring disabled
@@ -164,7 +164,7 @@ func resourceRancher2ClusterCreate(d *schema.ResourceData, meta interface{}) err
 
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{},
-		Target:     []string{expectedState},
+		Target:     expectedState,
 		Refresh:    clusterStateRefreshFunc(client, newCluster.ID),
 		Timeout:    d.Timeout(schema.TimeoutCreate),
 		Delay:      1 * time.Second,


richardcase (approved):

mjura (approved):
LGTM

a-blender (changes requested):
@furkatgofurov7 I see that you were unable to repro this bug after multiple attempts https://github.com/rancher/terraform-provider-rancher2/issues/1003#issuecomment-1534879150 and the original author of the issue saw a 30 sec timeout in terraform apply but the cluster still became active in about a minute. This was with using the latest terraform v3.0.0.

1. Is it still an issue for them?

2. What evidence is there that not predefining the `expectedState` in the provider fixes this particular issue? I'm fine to have it go in as a noted enhancement but I don't see how this explicitly resolves https://github.com/rancher/terraform-provider-rancher2/issues/1003.

furkatgofurov7:
> 1. Is it still an issue for them?

Hey @a-blender thanks for review, as per https://github.com/rancher/terraform-provider-rancher2/issues/1003#issuecomment-1536012174 it should be still an issue

> 2. What evidence is there that not predefining the expectedState in the provider fixes this particular issue? I'm fine to have it go in as a noted enhancement but I don't see how this explicitly resolves https://github.com/rancher/terraform-provider-rancher2/issues/1003.

Based on the logs from the linked issue description:

`Error: [ERROR] waiting for cluster (c-xfbkg) to be created: timeout while waiting for state to become 'pending' (last state: 'active', timeout: 30m0s)`

and this [comment](https://github.com/rancher/terraform-provider-rancher2/issues/1003#issuecomment-1488267067), considering all possible cluster states should not hurt and in case of provider missing to catch the state of the rancher it will still be able to move on. 
Also, I forgot to mention that, the piece of code where `expectedState` is predefined beforehand appears only in this case, in all other cases, i.e clusterUpdate the [target](https://github.com/rancher/terraform-provider-rancher2/blob/f23fa72fca94a205fda457af1a76e8e6a77ab662/rancher2/resource_rancher2_cluster.go#LL385C3-L385C9) state is a slice with bunch of different states, so from that perspective this should align also with that. 

Since the reproduction was not possible, maybe we should take this in as an improvement to the provider, WDYT?

furkatgofurov7:
@a-blender can we merge this, based on https://github.com/rancher/terraform-provider-rancher2/issues/1003#issuecomment-1578267034 it helps to fix the issue in #1003 ?

tulanian:
Using v3.1.1 of the provider and still experiencing this problem:

 Error: [ERROR] waiting for cluster (c-w8rvv) to be created: timeout while waiting for state to become 'active, pending' (last state: 'waiting', timeout: 5m0s)
│ 
│   with module.import_to_rancher[0].rancher2_cluster.imported_eks_cluster,
│   on .terraform/modules/import_to_rancher/main.tf line 27, in resource "rancher2_cluster" "imported_eks_cluster":
│   27: resource "rancher2_cluster" "imported_eks_cluster" {

Issue #1112, opened 2023-05-01
[RFE] Add support for PodSecurityAdmissionConfigurationTemplate


mouellet:
### Is your feature request related to a problem? Please describe.
Rancher v2.7.2 now comes w/ a new CRD for [Pod Security Admission (PSA) Configuration Templates](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates) that can only be provisioned manually.

### Describe the solution you'd like
1. Add a new resource/datasource for [`rancher2_pod_security_admission_configuration_template`](https://github.com/rancher/rancher/blob/release/v2.7/pkg/client/generated/management/v3/zz_generated_pod_security_admission_configuration_template.go).
2. Add a the field [`defaultPodSecurityAdmissionConfigurationTemplateName`](https://github.com/rancher/rancher/blob/release/v2.7/pkg/client/generated/management/v3/zz_generated_cluster.go#LL123C3-L123C3) to the `rancher_cluster` and `rancher_cluster_v2` resources.

### Additional context

SURE-6290

lazyfrosch:
Anyone working on that? I noticed that setting profile `cis-1.23` via API, it will not set the PSACT to `rancher-restricted`. Defaulting that seems to be a feature from the UI only.

So updating to 1.25 would require `kubernetes_version`, `rke_config.machine_selector_config.config.profile` and `default_pod_security_admission_configuration_template_name` once implemented.

Without a CIS profile it will work of course, but the RKE2 default under `cis-1.23` will block Rancher add-ons to be deployed.

nickvth:
Q3? This will be to late, can this issue be moved to Q2?

frankbou:
I configured a PSA template manually on a RKE1 test cluster by using the Rancher UI.
That RKE test cluster has been previously created by using Terraform.

After the PSA template change, the next time I did a "terraform apply -refresh-only", i got an error returned by Terraform.

```
[...]

PR #909, opened 2022-04-23
Add admission_configuration configuration

jocelynthode:
Addresses https://github.com/rancher/terraform-provider-rancher2/issues/908 

I'm rather new to write Golang and using the terraform SDK. I'm open to any help in improving this code.

Unfortunately as specified here: https://github.com/rancher/rancher/issues/37455 AdmissionConfiguration type coming from K8s is erased and is mapped to `map[string]interface{}` this forces me to do a lot of casting to make it work.

With this fix with the following terraform code:
```
admission_configuration {
  api_version = "apiserver.config.k8s.io/v1"
  kind        = "AdmissionConfiguration"
  plugins {
    name          = "PodSecurity"
    path          = ""
    configuration = file("files/security.yml")
  }
  plugins {
    name          = "EventRateLimit"
    path          = ""
    configuration = file("files/ratelimit.yml")
  }
}
```

We obtain the following code in the terraform state
```
"admission_configuration": [
  {
    "api_version": "apiserver.config.k8s.io/v1",
    "kind": "AdmissionConfiguration",
    "plugins": [
      {
        "configuration": "apiVersion: pod-security.admission.config.k8s.io/v1alpha1\ndefaults:\n  audit: restricted\n  audit-version: latest\n  enforce: restricted\n  enforce-version: latest\n  warn: restricted\n  warn-version: latest\nexemptions:\n  namespaces: []\n  runtimeClasses: []\n  usernames: []\nkind: PodSecurityConfiguration\n",
        "name": "PodSecurity",
        "path": ""
      },
      {
        "configuration": "apiVersion: eventratelimit.admission.k8s.io/v1alpha1\nkind: Configuration\nlimits:\n- burst: 20000\n  qps: 5000\n  type: Server\n",
        "name": "EventRateLimit",
        "path": ""
      }
    ]
  }
]
```

Issue #908, opened 2022-04-22
No support for `admission_configuration`


jocelynthode:
Hi,

The plugin currently expects a string for plugins but it receives a `[]interface {}` 

```
│ Error: rke_config.0.services.0.kube_api.0.admission_configuration.plugins: '' expected type 'string', got unconvertible type '[]interface {}', value: '[map[configuration:map[apiVersion:pod-security.admission.config.k8s.io/v1alpha1 defaults:map[audit:restricted audit-version:latest enforce:restricted enforce-version:latest warn:restricted warn-version:latest] exemptions:map[namespaces:[] runtimeClasses:[] usernames:[]] kind:PodSecurityConfiguration] name:PodSecurity path:] map[configuration:map[apiVersion:eventratelimit.admission.k8s.io/v1alpha1 kind:Configuration limits:[map[burst:20000 qps:5000 type:Server]]] name:EventRateLimit path:]]'
```

Here is the yaml that generates this error:
```
admissionConfiguration:
  apiVersion: apiserver.config.k8s.io/v1
  kind: AdmissionConfiguration
  plugins:
  - configuration:
      apiVersion: pod-security.admission.config.k8s.io/v1alpha1
      defaults:
        audit: restricted
        audit-version: latest
        enforce: restricted
        enforce-version: latest
        warn: restricted
        warn-version: latest
      exemptions:
        namespaces: []
        runtimeClasses: []
        usernames: []
      kind: PodSecurityConfiguration
    name: PodSecurity
    path: ""
  - configuration:
      apiVersion: eventratelimit.admission.k8s.io/v1alpha1
      kind: Configuration
      limits:
      - burst: 20000
        qps: 5000
        type: Server
    name: EventRateLimit
    path: ""
```
The issue seems to come from here: https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_cluster_rke_config_services_kube_api.go#L286 

as TypeMap only allows String as value. Is there a workaround for this? Otherwise I'd be glad to submit a PR for this fix if I could be pointed in the right direction to start fixing this.

a-blender:
## Test plan

The author of PR https://github.com/rancher/terraform-provider-rancher2/pull/909 has been using the tf rancher2 changes in production for a few months without any issue. He's been testing on an rke cluster with kubernetes 1.23.6 so I would test this on a k8s 1.24 cluster to make sure that works. If QA has low bandwidth this issue can be closed.

Steps

[...]

PR #1119, opened 2023-05-13
Enable setting default PSACT for rancher2_cluster add admission_configuration support with state migration logic

git-ival:
## Issue: 
  - https://github.com/rancher/terraform-provider-rancher2/issues/1112 (partial implementation for "V1" clusters)
  - https://github.com/rancher/terraform-provider-rancher2/issues/908

## Problem
Implement setting the default PSACT on V1 clusters using Terraform and add support for setting the `admission_configuration` (required since Rancher's response would cause terraform to error on th schema).
 
## Solution
  - Adds the `default_pod_security_admission_configuration_template_name` field to `rancher2_cluster` and `data_rancher2_cluster`
  - Re-instates PR rancher/terraform-provider-rancher2#909's changes (with some modifications)
  - Adds state upgrade migration logic to handle conversion of `admission_configuration` from a map of strings (in golang `map[string]interface{}`) to a list of maps of strings and complex types (in golang `[]map[string]interface{}{}`)
 
## Testing

Was able to create clusters on K8s 1.23, 1.24, 1.25 and 1.26 with valid admission configurations.

K8s 1.23 and 1.24 admission configuration:
```hcl
admission_configuration = {
    "api_version" : "apiserver.config.k8s.io/v1",
    "kind" : "AdmissionConfiguration",
    "plugins" : [
      {
        "configuration" : <<-EOT
          apiVersion: pod-security.admission.config.k8s.io/v1alpha1
          defaults:
            enforce: privileged
            enforce-version: latest
          exemptions: {}
          kind: PodSecurityConfiguration
        EOT
        "name" : "PodSecurity",
        "path" : ""
      }
    ]
```

K8s 1.25 and 1.26 admission configuration:
```hcl
admission_configuration = {
    "api_version" : "apiserver.config.k8s.io/v1",
    "kind" : "AdmissionConfiguration",
    "plugins" : [
      {
        "configuration" : <<-EOT
          apiVersion: pod-security.admission.config.k8s.io/v1
          defaults:
            enforce: privileged
            enforce-version: latest
          exemptions: {}
          kind: PodSecurityConfiguration
        EOT
        "name" : "PodSecurity",
        "path" : ""
      }
    ]
```

```hcl
resource "rancher2_cluster" "this" {
  name = "test-cluster"

  default_pod_security_admission_configuration_template_name = "rancher-privileged"
  rke_config {
    kubernetes_version = var.kubernetes_version
    services {
      kube_api {
        admission_configuration {
          api_version = var.admission_configuration.api_version
          kind        = var.admission_configuration.kind
          plugins {
            configuration = var.admission_configuration.plugins[0].configuration
            name          = var.admission_configuration.plugins[0].name
            path          = var.admission_configuration.plugins[0].path
          }
        }
      }
    }
  }
}
```

### Manual Testing

Tested state migration for clusters with no existing admission configuration.

Was not able to test on a cluster with an existing admission configuration imported by the previous tfp version.

### Automated Testing
Some unit tests were added by the contributor for PR rancher/terraform-provider-rancher2#909

## QA Testing Considerations

Could use testing on a cluster with an existing admission configuration imported by the previous tfp version
 
I am hopeful that there shouldn't be any regressions due to the added state migration logic.


nickvth:
This is needed to upgrade to k8s 1.25, we can't upgrade through terraform at this moment. 
@git-ival @a-blender @HarrisonWAffel can someone review this and make a release, so we can upgrade through terraform to 1.25.x, 1.24.x = EOL.

rancher2/resource_rancher2_cluster.go
```diff
@@ -306,17 +306,18 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
...
-		"dockerRootDir":                       d.Get("docker_root_dir").(string),
-		"fleetWorkspaceName":                  d.Get("fleet_workspace_name").(string),
-		"enableClusterAlerting":               d.Get("enable_cluster_alerting").(bool),
-		"enableClusterMonitoring":             d.Get("enable_cluster_monitoring").(bool),
-		"enableNetworkPolicy":                 &enableNetworkPolicy,
-		"istioEnabled":                        d.Get("enable_cluster_istio").(bool),
-		"localClusterAuthEndpoint":            expandClusterAuthEndpoint(d.Get("cluster_auth_endpoint").([]interface{})),
-		"annotations":                         toMapString(d.Get("annotations").(map[string]interface{})),
-		"labels":                              toMapString(d.Get("labels").(map[string]interface{})),
+		"defaultPodSecurityAdmissionConfigurationTemplateName": d.Get("default_pod_security_admission_configuration_template_name").(string),
```
a-blender, line 314: Fix formatting after adding new field

rancher2/schema_cluster_rke_config_services_kube_api.go
```diff
@@ -123,6 +125,115 @@ func clusterRKEConfigServicesKubeAPIAuditLogFields() map[string]*schema.Schema {
...
+	s := map[string]*schema.Schema{
+		"name": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+		},
+		"path": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Default:  "",
```
a-blender: If it's `Optional`, the default for admission config plugins path will never be used.

rancher2/schema_cluster_rke_config_services_kube_api.go
```diff
@@ -123,6 +125,115 @@ func clusterRKEConfigServicesKubeAPIAuditLogFields() map[string]*schema.Schema {
...
```
a-blender: Same as above about the path Default.

rancher2/schema_cluster_rke_config_services_kube_api.go
```diff
@@ -123,6 +125,115 @@ func clusterRKEConfigServicesKubeAPIAuditLogFields() map[string]*schema.Schema {
...
+					return
+				}
+				return
+			},
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+				if old == "" || new == "" {
+					return false
+				}
+				oldMap, _ := ghodssyamlToMapInterface(old)
+				newMap, _ := ghodssyamlToMapInterface(new)
```
a-blender, line 317: Do we expect both `old` and `new` admission config strings to be syntactically correct yaml here? Shouldn't we check for errors? Or actually, why not diff the strings without converting them both to maps? That's a less expensive operation.

rancher2/structure_cluster_rke_config_services_kube_api.go
```diff
@@ -210,6 +263,55 @@ func expandClusterRKEConfigServicesKubeAPIAuditLog(p []interface{}) (*management
...
+		if err != nil {
+			return obj, fmt.Errorf("Unmarshalling plugin configuration from yaml for %v: %v", pluginMap["name"], err)
+		}
+		newObj["configuration"] = configuration
+		obj = append(obj, newObj)
+	}
+	return obj, nil
+}
+
+func expandClusterRKEConfigServicesKubeAPIAdmissionConfiguration(p []interface{}) (map[string]interface{}, error) {
```
a-blender, line 289: @HarrisonWAffel I'm going to need your eyes on this.

The admission_configuration 'migration code' added here looks similar, if not exactly the same, to the update from PR https://github.com/rancher/terraform-provider-rancher2/pull/1026/commits/5788729cae9a63fb485d65248a562f571729b776 which caused a breaking change for users who created clusters with an earlier version than Terraform v1.24.2 because the schema is changed from map -> complex structure. Admission config templates are part of the PSA feature as I understand, but I don't think this can go in until we resolve this issue. Do you see the same potential regression here?

Why are we parsing out the yaml configuration for `admission_configuration` into API version / kind / plugins? Seems unnecessary. Is there a way to keep it as a map so the schema doesn't change?

@git-ival Added state migration logic may have handled this conversion, but to confirm you need to test managing a cluster with Terraform 3.0.0 that was originally created with v1.24.0 and confirm that there are no schema errors. I don't support merging until this is verified.

a-blender (changes requested):

rancher2/structure_cluster_rke_config_services_kube_api.go
HarrisonWAffel: The code you've highlighted with this comment is not the migration logic from what I can tell, it looks like the migration logic added in this PR is actually in the `rancher2/resource_rancher2_cluster.go` file on line 107. 

This code looks the same because this PR is attempting to introduce the same schema as the prior PR, but with migration logic included to move the old schema into the new schema. The core principal of this PR is a schema change so that admission configurations can be used more easily / at all. 

I agree, this shouldn't be merged until that migration path can be properly tested. Currently the PR states `Was not able to test on a cluster with an existing admission configuration imported by the previous tfp version.` which leaves room for regressions.

a-blender: After offline discussion, I agree. An original AC configuration will also be needed  in order to test the migration from an earlier TF version to one with these updates.

a-blender:
@git-ival Please test the migration state logic for the `admission_configuration` field before this PR is merged.

Ideal test steps
1. Provision an rke cluster via [Terraform v1.24.1](https://github.com/rancher/terraform-provider-rancher2/releases/tag/v1.24.1). Set `admission_configuration` field in your config, here is [an example](https://github.com/rancher/terraform-provider-rancher2/pull/909#issuecomment-1238238113).
2. Verify the field is used in the active cluster
3. Upgrade Terraform to a local version with your updates

I test locally on Mac by doing the following

Update `required_providers` block to
```
terraform {
  required_providers {
    rancher2 = {
      source  = "terraform.local/local/rancher2"
      version = "1.0.0"
    }
  }
}
```

Build your local repo and put the new binary in ~/.terraform.d/
```
make bin
cp terraform-provider-rancher2 ~/.terraform.d/plugins/terraform.local/local/rancher2/1.0.0/darwin_amd64/terraform-provider-rancher2
rm -r .terraform.lock.hcl
terraform init
terraform apply
```

This approach to local testing has an [open PR in review](https://github.com/rancher/terraform-provider-rancher2/pull/1034).

4. Observe that the migration was successful and no errors or data loss occurs on future runs of `terraform apply` or while managing the cluster.

a-blender:
@git-ival Do you have an eta on when this can be done? Thank you.

rancher2/resource_rancher2_cluster.go
git-ival: Is this done via `make fmt` ?

rancher2/schema_cluster_rke_config_services_kube_api.go
git-ival: Can you explain what you mean by diffing the strings without converting them? From my understanding the admission config strings are expected to  be properly formatted yaml, but if there is a way we could **not** require that, I guess that would be preferable.

git-ival:
> Update `required_providers` block to
> 
> ```
> terraform {
>   required_providers {
>     rancher2 = {
>       source  = "terraform.local/local/rancher2"
>       version = "1.0.0"
>     }
>   }
> }
> ```

@a-blender Shouldnt it be `version = "1.24.1"`?

git-ival:
> @git-ival Do you have an eta on when this can be done? Thank you.

@a-blender I'm still looking into testing after making some changes and I still have a few open questions, but I'm hoping to have this done before the end of the week.

a-blender:
@git-ival Well, you would start provisioning a cluster with TFP v1.24.1 with `admission_configuration` field yes. But then, you'll have to upgrade the TFP to a binary built from this branch (like my example) to test the upgrade. Sounds great, thank you for testing this so quickly!

rancher2/resource_rancher2_cluster.go
a-blender: If you are working in an IDE, it will likely auto format when you save the file. But yes running `make fmt` should also work.

git-ival: Hm, both the IDE and `make fmt` use this formatting, even if I format it after manually aligning the lines then it reverts to the current formatting. Should I override it manually?

rancher2/schema_cluster_rke_config_services_kube_api.go
a-blender: True, but I look a deeper look and this format for the diffsupress is being used across the provider, for configs like `RKEConfig` too. So removing this element update from the plan if the `DeepEquals` is true looks fine.

rancher2/resource_rancher2_cluster.go
a-blender: If that's the formatting `make fmt` leaves it in, then that's fine.

git-ival:
@a-blender I have tried various configurations and methods of setting the `admission_configuration` on the 1.24.1 version of the tfp, but I receive a spec error regardless. I couldn't follow the example given in the linked PR because PR https://github.com/rancher/terraform-provider-rancher2/pull/909 is not included in any official versions of the tfp. Is there some confirmed way to set the `admission_configuration` for clusters provisioned via the tfp without this PR or PR https://github.com/rancher/terraform-provider-rancher2/pull/909?

Here is the error:

```
│ Error: Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [message=Cluster spec conversion error, baseType=error, code=InvalidBodyContent] from [<rancher_url>/v3/clusters]
```

a-blender:
Unfortunately, we don't have a confirmed way. The `admission_configuration` field exists in v1.24.1 as a map https://github.com/rancher/terraform-provider-rancher2/blob/d6b9660b798969481c7cc5367cbaa4c83d2bd500/rancher2/schema_cluster_rke_config_services_kube_api.go#L285-L288. It still has to be tested before merge. I'll see if I can help figure it out

git-ival:
> It still has to be tested before merge. I'll see if I can help figure it out

@a-blender Agreed, in the meantime I imported 2 clusters with admission configuration and provisioned 2 clusters without it (I then added admission configuration to them manually via the UI). I will proceed with testing using this environment, and continue trying to provision a cluster with old tfp version with admission configuration.

git-ival:
@a-blender After many many attempts at trial and erroring a solution to get the previous tfp version to provision a cluster with admission configuration.  I ended up looking through the actual `rancher/rancher` code and found the admission configuration structure through that. The tfp is expecting the same structure defined in the `KubeAPIServiceFieldAdmissionConfiguration` struct (`map[string]interface{}`). 

However, passing in a map matching this structure does not result in a successful `terraform apply`. Additionally, trying to import a cluster that has a valid admission configuration also leads to errors parsing the struct. It expects the "plugins" value to be a `string`, but it receives a `map[string]interface{}` instead.

This leads me to believe that on tfp versions prior to this PR and PR #909, provisioning a cluster with admission configuration set is not possible. So I will proceed without testing the case where a cluster with admission configuration was provisioned on a previous tfp version. Instead I will test the case where a cluster provisioned by a previous tfp version has an admission configuration added to it manually and is then refreshed and/or imported into the state using the new tfp version.

a-blender:
@git-ival This is also what I found on Friday. My rke cluster with tf 1.24.1 provisions successfully with admission configuration `plugins` not defined, but all attempts I've made to configure plugins using [method 2](https://rke.docs.rancher.com/config-options/services/pod-security-admission) as a string (file reference, json encoded string EOF) is _syntactically correct_ but still fails on `terraform apply` with the error you keep seeing on cluster creation,

```
│ Error: Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [message=Cluster spec conversion error, baseType=error, code=InvalidBodyContent] from [<rancher_url>/v3/clusters]
```

Rancher does expect a `map[string]interface{}`. It's likely this field is broken.

Sounds like a plan! Let us know your results.

git-ival:
@a-blender 
Tested with two provisioned and two imported RKE1 clusters (k8s versions tested against are 1.24.10 and 1.26.3).

Test result: PASS

Test steps:

1. Provision a 1.24.10 and 1.26.3 RKE cluster via [Terraform v1.24.1](https://github.com/rancher/terraform-provider-rancher2/releases/tag/v1.24.1)
2. Import a 1.24.10 and 1.26.3 RKE cluster with `admission_configuration` set in the RKE cluster.yml (Note: this -must- include all relevant namespaces that Rancher may use in the exemptions see [here ](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions))
3. Modify each provisioned cluster and add a valid `admission_configuration` as appropriate for the specific k8s version (see examples in the original text of this PR for specific differences. See [this link](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions) for the PSA config used)
4. Verify the field is used in the active cluster
5. Create new namespace in each cluster
6. Create a new deployment in the created namespace
7. Observe that the pods cannot enter a running state due to the following `... is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false ...` (This confirms that the admission_configuration is running as expected)
8. Upgrade Terraform to a local version with this PR's updates
9. Update your terraform module to account for this PR's updates (admission_configuration is now a block)
10. Refresh the terraform state
11. Observe the state file to confirm the expected admission_configuration structure and values 

Notes: 
  - Reverting to a previous version is not possibe without a tfstate backup from when the desired provider version was used.
  - Imported clusters were also used, but Rancher does not appear to have any view into the PSA config of imported clusters, so the terraform state does not include any admission_configuration for imported clusters

provisioned 1.24.10 admission_configuration in terraform state file:
```
...
                    "kube_api": [
                      {
                        "admission_configuration": [
                          {
                            "api_version": "apiserver.config.k8s.io/v1",
                            "kind": "AdmissionConfiguration",
                            "plugins": [
                              {
                                "configuration": "apiVersion: pod-security.admission.config.k8s.io/v1alpha1\ndefaults:\n  audit: restricted\n  audit-version: latest\n  enforce: restricted\n  enforce-version: latest\n  warn: restricted\n  warn-version: latest\nexemptions:\n  namespaces:\n  - calico-apiserver\n  - calico-system\n  - cattle-alerting\n  - cattle-csp-adapter-system\n  - cattle-elemental-system\n  - cattle-epinio-system\n  - cattle-externalip-system\n  - cattle-fleet-local-system\n  - cattle-fleet-system\n  - cattle-gatekeeper-system\n  - cattle-global-data\n  - cattle-global-nt\n  - cattle-impersonation-system\n  - cattle-istio\n  - cattle-istio-system\n  - cattle-logging\n  - cattle-logging-system\n  - cattle-monitoring-system\n  - cattle-neuvector-system\n  - cattle-prometheus\n  - cattle-resources-system\n  - cattle-sriov-system\n  - cattle-system\n  - cattle-ui-plugin-system\n  - cattle-windows-gmsa-system\n  - cert-manager\n  - cis-operator-system\n  - fleet-default\n  - ingress-nginx\n  - istio-system\n  - kube-node-lease\n  - kube-public\n  - kube-system\n  - longhorn-system\n  - rancher-alerting-drivers\n  - security-scan\n  - tigera-operator\nkind: PodSecurityConfiguration\n",
                                "name": "PodSecurity",
                                "path": ""
                              }
                            ]
                          }
                        ],
...
```

provisioned 1.26.3 admission_configuration in terraform state file:
```
...
                    "kube_api": [
                      {
                        "admission_configuration": [
                          {
                            "api_version": "apiserver.config.k8s.io/v1",
                            "kind": "AdmissionConfiguration",
                            "plugins": [
                              {
                                "configuration": "apiVersion: pod-security.admission.config.k8s.io/v1\ndefaults:\n  audit: restricted\n  audit-version: latest\n  enforce: restricted\n  enforce-version: latest\n  warn: restricted\n  warn-version: latest\nexemptions:\n  namespaces:\n  - calico-apiserver\n  - calico-system\n  - cattle-alerting\n  - cattle-csp-adapter-system\n  - cattle-elemental-system\n  - cattle-epinio-system\n  - cattle-externalip-system\n  - cattle-fleet-local-system\n  - cattle-fleet-system\n  - cattle-gatekeeper-system\n  - cattle-global-data\n  - cattle-global-nt\n  - cattle-impersonation-system\n  - cattle-istio\n  - cattle-istio-system\n  - cattle-logging\n  - cattle-logging-system\n  - cattle-monitoring-system\n  - cattle-neuvector-system\n  - cattle-prometheus\n  - cattle-resources-system\n  - cattle-sriov-system\n  - cattle-system\n  - cattle-ui-plugin-system\n  - cattle-windows-gmsa-system\n  - cert-manager\n  - cis-operator-system\n  - fleet-default\n  - ingress-nginx\n  - istio-system\n  - kube-node-lease\n  - kube-public\n  - kube-system\n  - longhorn-system\n  - rancher-alerting-drivers\n  - security-scan\n  - tigera-operator\nkind: PodSecurityConfiguration\n",
                                "name": "PodSecurity",
                                "path": ""
                              }
                            ]
                          }
                        ],
...
```

rancher2/structure_cluster_rke_config_services_kube_api.go
git-ival: Tested without provisioning a cluster with admission_config enabled, as discussed (see latest comments for update).

a-blender:
@git-ival This is perfect! To confirm,

1. There were no errors on a `tf apply` or `tf refresh` for the 1.24 cluster post upgrade?
2. Are the shots that you posted of the tf state file pre or post-upgrade? I'm assuming post-upgrade? If not, what does `admission_configuration` look like in the state file after?

cc @HarrisonWAffel This is lookin' real good to me. Let me know if you have any additional feedback or concerns.

git-ival:
> @git-ival This is perfect! To confirm,
> 
> 1. There were no errors on a `tf apply` or `tf refresh` for the 1.24 cluster post upgrade?
> 2. Are the shots that you posted of the tf state file pre or post-upgrade? I'm assuming post-upgrade? If not, what does `admission_configuration` look like in the state file after?
> 
> cc @HarrisonWAffel This is lookin' real good to me. Let me know if you have any additional feedback or concerns.

1. That's correct, I did not encounter any errors post upgrade.
2. The tfstates are post-upgrade. Pre-upgrade the admission_configuration would be a null value since there doesn't seem to be a valid way to set this field (and you cant import a cluster with this field set) in older tfp versions.

a-blender:
@git-ival When you did this step

```
Modify each provisioned cluster and add a valid admission_configuration as appropriate for the specific k8s version 
```

And refreshed tf, did `admission_configuration` show up in the state file? Terraform should pick up the upstream field.

git-ival:
> @git-ival When you did this step
> 
> ```
> Modify each provisioned cluster and add a valid admission_configuration as appropriate for the specific k8s version 
> ```
> 
> And refreshed tf, did `admission_configuration` show up in the state file? Terraform should pick up the upstream field.

Yes, on both versions the tfp identified the change in `admission_configuration`, but on the previous tfp it could not complete the import/refresh because the spec did not match what the tfp was expecting.

a-blender:
We want to make sure what we've tested is usable. Could you clarify a bit more that `Yes, on both versions the tfp identified the change in admission_configuration, but on the previous tfp it could not complete the import/refresh because the spec did not match what the tfp was expecting.`? A refresh should be able to complete in this scenario. Tf should just update the local state file with the config in the upstream cluster. What did you see? Did you see an error on refresh?

git-ival:
> We want to make sure what we've tested is usable. Could you clarify a bit more that `Yes, on both versions the tfp identified the change in admission_configuration, but on the previous tfp it could not complete the import/refresh because the spec did not match what the tfp was expecting.`? A refresh should be able to complete in this scenario. Tf should just update the local state file with the config in the upstream cluster. What did you see? Did you see an error on refresh?

To clarify further: 
  - When using this PR's tfp changes, refreshing and/or importing a cluster with `admission_configuration` set **updates the state successfully**.
  - When using a previous tfp version, refreshing and/or importing a cluster with `admission_configuration` set **fails to update the state** due to a spec error on `admission_configuration`.

git-ival:
@a-blender @HarrisonWAffel 
I have tested the case where `admission_configuration` is unset throughout the above defined test. In this scenario the `admission_configuration` is never defined in the inputs. 

On the previous tfp version, post-refresh, the `admission_configuration`'s tfstate value is the empty/null object value of `{}`.
After upgrading to the latest local tfp version of this PR and refreshing the state, `admission_configuration`'s tfstate value is correctly updated to the empty/null list value of `[]` without errors.

Note: Before refresh and before updating to this PR's tfp changes, the `admission_configuration`'s tfstate value was the `null` value. **Post-refresh** tfstate values are shown below.

tfstate on previous tfp version:
```
...
"kube_api": [
                      {
                        "admission_configuration": {},
...
```

tfstate with updated tfp version:
```
...
"kube_api": [
                      {
                        "admission_configuration": [],
...
```

HarrisonWAffel:
@git-ival Thank you for also testing that scenario, as that is the case users ran into when we last changed the schema.

HarrisonWAffel (approved):

a-blender (approved):
Great! Thanks ya'll, lgtm.

PR #1121, opened 2023-05-13
Make DO userdata default empty for v2prov

jakefhyde:
## Problem

For some reason, the default userdata is populated as docker-user, when it should be empty. This prevent users from provisioning DO downstream clusters without specifying an empty userdata.

## Solution

Make default empty.

## Testing

I'll update this when I test it.


a-blender (approved):

jiaqiluo (approved):
LGMT.

Issue #1097, opened 2023-03-31
[RFE] New Fields for Cluster Agent Customization feature


snasovich:
Implement TFP-side changes for supporting functionality to be added by https://github.com/rancher/rancher/issues/41035

a-blender:
# QA Test Template

Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1097 

  
# Problem

 
Cluster/Fleet agent deployment customization is not supported by the TFP.

# Solution

  
Add `cluster_agent_deployment_customization` and `fleet_agent_deployment_customization` (and nested) fields tolerations, affinity, and resource requirements to the rancher2 provider so users will be able to define those values for RKE, RKE2/K3s, and EKS clusters.

# Testing


**Test RC: v3.1.0-rc2**
 
## Engineering Testing

This feature has been tested on rke and v2 prov clusters (EKS) by provisioning with both cluster and agent deployment customization, updating the customization, and then removing it and adding it back and making sure the cluster and agent redeploys function as expected.

### Manual Testing

 
Test plan
- [ ] rke single all-role node + agent customization
	- [ ] remove agent customization, verify removed
	- [ ] add it back, verify customization is added
	- [ ] update customization, verified it got updated on the v3 cluster spec and the agent deployment yaml
- [ ] rke2 (v2 prov) single all-role node + agent customization
	- [ ] same as above
- [ ] eks 2 node + agent customization
	- [ ] same as above

Note: _you may not see cluster go into `Updating` state in the UI for added/updated agent customization or minor updates because the redeploy is fast._

<details>

<summary>main.tf.rke_agent_customization</summary>

```yaml
terraform {
  required_providers {
    rancher2 = {
      source  = "terraform.local/local/rancher2"
      version = "1.0.0"
    }
  }
}

provider "rancher2" {
  api_url   = var.rancher_api_url
  token_key = var.rancher_admin_bearer_token
  insecure  = true
}

data "rancher2_cloud_credential" "rancher2_cloud_credential" {
  name = var.cloud_credential_name
}

resource "rancher2_cluster" "ablender-rke" {
  name = var.rke_cluster_name
  rke_config {
    kubernetes_version = "v1.26.4-rancher2-1"
    network {
      plugin = var.rke_network_plugin
    }
    services {
      etcd {
        creation = "6h"
        retention = "72h"
      }
    }
    ingress {
      default_backend = "false"
    }
  }
  cluster_auth_endpoint {}
[...]

PR #1137, opened 2023-06-01
Terraform rancher2 agent customization

a-blender:
## Issue: 

 
https://github.com/rancher/terraform-provider-rancher2/issues/1097

## Problem

 
Cluster/Fleet agent deployment customization is not supported by the TFP.

## Solution

 
Add `cluster_agent_deployment_customization` and `fleet_agent_deployment_customization` (and nested) fields tolerations, affinity, and resource requirements to the rancher2 provider so users will be able to define those values for RKE, RKE2/K3s, and EKS clusters.

**Error handling**

For agent deployment customization, we have a unique scenario with error handling. Most tf resources do not add error handling in the expander, but for rke clusters I have to marshal v1 Tolerations and Affinity strings into objects that exist on the rancher backend. They are being passed as strings because 1) it's too complex to struct out all the nested fields, 2) this is how terraform harvester encodes node affinity. If the Toleration or Affinity string passed via the tf config is invalid, expanding the field will fail. I argue that this should cause a cluster create/update to also fail.

My solution to this is to add error handling to the expander. If expanding an agent deployment customization returns nil with no error, customization is not defined and a create/update will pass. If it returns nil *with* an error, a create/update will fail and the user needs to fix their config.

## Testing

6/2 update: Not fully tested yet, but requesting preliminary review of backend logic for the rke case.

6/13 update: Finished testing rke with all new fields and fixing bugs. Working on adding EKS, testing v2 prov and EKS and updating tests.

6/16 update: Added `Toleration` v2 and `ResourceRequirement` resources, updated backend logic. Added/updated tests which turned out to be tricky. Need to re-test

### Manual Testing

Smoke testing
- [x] rke single all-role node + agent customization
	- [x] remove agent customization, verify removed
	- [x] add it back, verify customization is added
	- [x] update customization, verified it got updated on the v3 cluster spec and the agent deployment yaml
- [x] rke2 (v2 prov) single all-role node + agent customization
	- [x] same as above
- [x] eks 2 node + agent customization
	- [x] same as above

Note: _you may not see cluster go into `Updating` state in the UI for added/updated agent customization or minor updates because the redeploy is fast._


rancher2/data_source_rancher2_cluster.go
```diff
@@ -150,7 +150,7 @@ func dataSourceRancher2Cluster() *schema.Resource {
 				MaxItems: 1,
 				Computed: true,
 				Elem: &schema.Resource{
-					Schema: clusterRegistationTokenFields(),
+					Schema: clusterRegistrationTokenFields(),
```
jakefhyde, line 153: :+1:

rancher2/schema_agent_deployment_customization.go
```diff
@@ -0,0 +1,57 @@
...
+			Optional:    true,
+		},
+		"cpu_request": {
+			Type:        schema.TypeString,
+			Description: "The minimum CPU required for agent",
+			Optional:    true,
+		},
+		"memory_limit": {
+			Type:        schema.TypeString,
+			Description: "The maximum CPU limit for agent",
```
jakefhyde:
```suggestion
			Description: "The maximum memory limit for agent",
```

rancher2/structure_agent_deployment_customization.go
```diff
@@ -0,0 +1,96 @@
...
+
+		obj.Limits = make(map[string]string)
+		obj.Requests = make(map[string]string)
+
+		if v, ok := in["cpu_limit"].(string); ok && len(v) > 0 {
+			obj.Limits["cpu"] = v
+		}
+
+		if v, ok := in["cpu_request"].(string); ok && len(v) > 0 {
+			obj.Requests["cpu"] = v
```
jakefhyde: should `"cpu"` be `corev1.ResourceCPU`?

rancher2/structure_agent_deployment_customization.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		if v, ok := in["memory_limit"].(string); ok && len(v) > 0 {
+			obj.Limits["memory"] = v
+		}
+
+		if v, ok := in["memory_request"].(string); ok && len(v) > 0 {
+			obj.Requests["memory"] = v
```
jakefhyde: should `"memory"` be `corev1.ResourceMemory`?

rancher2/structure_agent_deployment_customization.go
```diff
@@ -0,0 +1,96 @@
...
+	return obj
+}
+
+func expandAgentDeploymentCustomization(p []interface{}) (*managementClient.AgentDeploymentCustomization, error) {
+	obj := &managementClient.AgentDeploymentCustomization{}
+	if len(p) == 0 || p[0] == nil {
+		return obj, fmt.Errorf("agent deployment customization was nil")
+	}
+
+	in := p[0].(map[string]interface{})
```
HarrisonWAffel, line 43: It might be a good idea to add an `ok` check here, just to be safe

rancher2/structure_agent_deployment_customization_v2.go
```diff
@@ -0,0 +1,119 @@
...
+func expandAgentDeploymentCustomizationOverrideResourceRequirementsV2(p []interface{}) (*corev1.ResourceRequirements, error) {
+	obj := &corev1.ResourceRequirements{}
+	if p == nil || len(p) == 0 || p[0] == nil {
+		return nil, nil
+	}
+
+	for i := range p {
+		in := p[i].(map[string]interface{})
+
+		obj.Limits = corev1.ResourceList{}
```
HarrisonWAffel: We need to instantiate the map inside any `ResourceList` before we can use it, since the actual type of a `ResourceList` is `map[ResourceName]resource.Quantity`. If we don't I'm pretty sure we will be working with a nil map and line 55 will panic

rancher2/structure_agent_deployment_customization.go
a-blender: None of the other tf expander funcs have that and I already check for p[0] == nil on line 69. So if we get there, we know that p[0] is non nil. Is checking the cast necessary?

HarrisonWAffel: Oh, good point it isn't necessary

jakefhyde (changes requested):

rancher2/structure_agent_deployment_customization.go
a-blender: It's interesting..
The management client that TF rke1 uses when creating/updating clusters has a resource requirement type of

```
type ResourceRequirements struct {
    Limits   map[string]string `json:"limits,omitempty" yaml:"limits,omitempty"`
    Requests map[string]string `json:"requests,omitempty" yaml:"requests,omitempty"`
}
```

Keeping it as a `map[string]string` and thus, "cpu" instead of `corev1.ResourceCPU` works fine during testing. I use the corev1 types for v2 prov.

rancher2/structure_cluster_v2.go
```diff
@@ -26,7 +28,7 @@ type ClusterV2 struct {
 
 func flattenClusterV2(d *schema.ResourceData, in *ClusterV2) error {
 	if in == nil {
-		return nil
+		return fmt.Errorf("[ERROR] flattening cluster: Input cluster is nil")
```
jakefhyde, line 31: We should make an EDI issue for this because `[ERROR]` as an error prefix is entirely non standard.

rancher2/structure_cluster_v2_test.go
```diff
@@ -60,14 +65,84 @@ func init() {
...
+			"append_tolerations": []interface{}{
+				map[string]interface{}{
+					"effect":   "NoSchedule",
+					"key":      "tolerate/test",
+					"operator": "Equal",
+					"seconds":  0,
+					"value":    "true",
+				},
+			},
+			"override_affinity": "{\n  \"nodeAffinity\": {\n    \"requiredDuringSchedulingIgnoredDuringExecution\": {\n      \"nodeSelectorTerms\": [{\n        \"matchExpressions\": [{\n          \"key\": \"not.this/nodepool\",\n          \"operator\": \"NotIn\",\n          \"values\": [\n            \"true\"\n          ]\n        }]\n      }]\n    }\n  }\n}",
```
jakefhyde: You should be able to use backtics to use an raw string for this.

rancher2/structure_agent_deployment_customization.go
jakefhyde: ~`corev1.ResourceCPU` is a constant, not a type~ Ah I see what you are saying. Do `corev1.ResourceCPU` and `corev1.ResourceMemory` work? If they do I think we should use them, if not no biggie.

rancher2/structure_resource_requirements.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+	if in.Requests != nil {
+		if len(in.Requests["cpu"]) > 0 {
+			obj["cpu_request"] = in.Requests["cpu"]
+		}
+		if len(in.Requests["memory"]) > 0 {
+			obj["memory_request"] = in.Requests["memory"]
+		}
+	}
```
jakefhyde, line 31: Same as https://github.com/rancher/terraform-provider-rancher2/pull/1137#discussion_r1219944777

rancher2/structure_resource_requirements.go
```diff
@@ -0,0 +1,69 @@
...
+			obj.Requests["cpu"] = v
+		}
+
+		if v, ok := in["memory_limit"].(string); ok && len(v) > 0 {
+			obj.Limits["memory"] = v
+		}
+
+		if v, ok := in["memory_request"].(string); ok && len(v) > 0 {
+			obj.Requests["memory"] = v
+		}
```
jakefhyde, line 65: Same as https://github.com/rancher/terraform-provider-rancher2/pull/1137#discussion_r1219944777

rancher2/structure_resource_requirements_test.go
```diff
@@ -0,0 +1,77 @@
+package rancher2
```
jakefhyde, line 1: Same as https://github.com/rancher/terraform-provider-rancher2/pull/1137#discussion_r1219944777

rancher2/structure_resource_requirements_v2.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if in.Requests != nil {
+		cpuRequestQuantity := in.Requests[corev1.ResourceCPU]
+		cpuRequest := cpuRequestQuantity.String()
+		if len(cpuRequest) > 0 {
+			obj["cpu_request"] = cpuRequest
+		}
+
+		memoryRequestQuantity := in.Requests[corev1.ResourceCPU]
```
jakefhyde:
```suggestion
		memoryRequestQuantity := in.Requests[corev1.ResourceMemory]
```

jakefhyde (changes requested):
Few nits

rancher2/structure_agent_deployment_customization.go
a-blender: @jakefhyde They are type `ResourceName` and not `string` so I'd have to do `obj.Limits[string(corev1.ResourceCPU)] = v` which I think is unnecessary. I think we should keep them as strings in the non-v2 files.

rancher2/structure_cluster_v2.go
a-blender: Why? Isn't the existence of [ERROR] to tag the verbosity of the log level if logs are enabled? https://developer.hashicorp.com/terraform/internals/debugging

jakefhyde: The error message isn't the place to do the formatting, it's too tightly coupling the error message to the output.

jakefhyde (approved):

HarrisonWAffel (approved):

Issue #1125, opened 2023-05-17
[RFE] New Fields for Rancher Machine Hostname Truncation feature


snasovich:
Implement TFP-side changes for supporting functionality to be added by https://github.com/rancher/rancher/issues/40901

PR #1147, opened 2023-06-14
1125 rancher machine hostname truncation

jakefhyde:
## Issue:  rancher/terraform-provider-rancher2#1125

 
Adding support for hostname truncation


rancher2/structure_cluster_v2_rke_config_machine_pool.go
```diff
@@ -97,6 +98,9 @@ func flattenClusterV2RKEConfigMachinePools(p []provisionv1.RKEMachinePool) []int
 		if in.UnhealthyRange != nil {
 			obj["unhealthy_range"] = *in.UnhealthyRange
 		}
+		if in.HostnameLengthLimit >= capr.MinimumHostnameLengthLimit && in.HostnameLengthLimit <= capr.MaximumHostnameLengthLimit {
```
jiaqiluo: The upper and lower limits are treated as valid values. However, they are not on the rancher side https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/capr/machineprovision/args.go#L358-L367 

 ```suggestion
 		if in.HostnameLengthLimit > capr.MinimumHostnameLengthLimit && in.HostnameLengthLimit < capr.MaximumHostnameLengthLimit {
```

**Review by jiaqiluo — CHANGES_REQUESTED** [review]

File: rancher2/structure_cluster_v2_rke_config_machine_pool.go (hunk 1 above)
Comment (jakefhyde) [reply rancher2/structure_cluster_v2_rke_config_machine_pool.go]: I think you may want to take another peek at the rancher code, as the logic is purposefully inverted there.

**Review by a-blender — APPROVED** [review]
@jakefhyde LGTM

A few observations which you can choose to change or not,
* `unhealthy` sounds strange to me in reference to a hostname, so `invalid_hostname_range` and `invalid_hostname_max` sound better but are these fields being used elsewhere?
* In the original issue, the default value was 10. Did 16 cause the least num of collisions in your testing?

**Comment by jakefhyde** [comment]
> @jakefhyde LGTM
> 
> A few observations which you can choose to change or not,
> 
> * `unhealthy` sounds strange to me in reference to a hostname, so `invalid_hostname_range` and `invalid_hostname_max` sound better but are these fields being used elsewhere?
> * In the original issue, the default value was 10. Did 16 cause the least num of collisions in your testing?

I didn't make any changes regarding `"unhealthy"`, I think you may be confusing that with the existing code that was already above? As far as choosing a default, I just chose 16 because its the nearest power of two greater than the minimum of 10, there is nothing special about it.

File: rancher2/structure_cluster_v2_rke_config_machine_pool.go (hunk 1 above)
Comment (jiaqiluo) [reply rancher2/structure_cluster_v2_rke_config_machine_pool.go]: @jakefhyde, I read the code again and have the following question: 
If the rancher side allows any value and only uses the value if `MinLimit < user-value < MaxLimit`, why do we need to check the value here? Why do we not just pass the value to Rancher and let Rancher determine?

**Review by jiaqiluo — CHANGES_REQUESTED** [review]

File: rancher2/structure_cluster_v2_rke_config_machine_pool.go (hunk 1 above)
Comment (jakefhyde) [reply rancher2/structure_cluster_v2_rke_config_machine_pool.go]: @jiaqiluo @a-blender Updated, added support for the cluster wide component (which I forgot) and minor tweaks based on feedback

**Review by jiaqiluo — APPROVED** [review]
LGTM 💯

File: rancher2/structure_cluster_v2_rke_config_machine_pool_defaults.go [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func flattenClusterV2RKEConfigMachinePoolDefaults(in provisionv1.RKEMachinePoolDefaults) []any {
+	if reflect.ValueOf(in).IsZero() {
+		return nil
+	}
+
+	obj := map[string]any{}
+
+	if in.HostnameLengthLimit != 0 {
```
a-blender: `in.HostnameLengthLimit > 0`

a-blender (changes requested):
@jakefhyde I think it's also a good idea to add structure tests for the RKE config machine pool default flattener and expander :) otherwise LGTM

a-blender (approved):

Issue #841, opened 2022-01-07
Data sources create new kubeconfig token when running `terraform plan` and `terraform apply`


sgapanovich:
### Setup: ###
Rancher version: 2.6.3
Rancher cluster type: HA
Terraform provider rancher 2 version: "1.22.2"

### Steps to reproduce: ###
1. Create a cluster using `rancher2_cluster` and/or  `rancherv2_cluster_v2` resources
2. Add data sources `rancher2_cluster` and/or  `rancher2_cluster_v2` with appropriate cluster names 

#### Issue: ####
New kubeconfig token is created every time you run `terraform plan` or `terraform apply` per each data source (`rancher2_cluster` and `rancherv2_cluster_v2`)

SURE-4979

eliyamlevy:
Just off a cursory look, is this not the behavior you'd want/expect from these commands?

flyte:
I've just noticed that it's doing this for me as well. I have my `RANCHER_ACCESS_KEY` and `RANCHER_SECRET_KEY` env vars set, then whenever I run a `terraform plan` (or apply) it creates a new kubeconfig token that never expires, even through I've only got the following data block:

```hcl
data "rancher2_cluster" "cluster" {
  name = var.rancher_cluster_name
}
```

> is this not the behavior you'd want/expect from these commands?

@eliyamlevy No. Why is it creating them at all, just to get information about a rancher cluster? Shouldn't the credentials in my env vars be sufficient to get this data? Also, why doesn't it set any expiry?

flyte:
I see why it's creating them now. It returns a generated `kube_config` from the `rancher2_cluster` data block.

Perhaps this should be configurable (generate credentials, expiry etc.) and by default not generate credentials? I'm only using this data block in order to get the cluster ID.

I just had to delete 65 generated credentials from my Rancher profile because it had created one each time I ran plan or apply.

raelix:
up

jmederer:
We've got the same Problem.
We do provisioning almost everything with terraform and so we use often rancher2_cluster module. 
At the end our etcd of Rancher stored over 32k kubeconfig and the etcd performance went down.

Workaround on our side right now - to delete those tokens daily by a cronjob 
Would be nice if this configurable like @flyte already suggest.

a-blender:
I have investigated this issue and this is expected in Rancher but a confirmed bug in Rancher TF provider.

[...]

PR #1158, opened 2023-06-22
Update replace kubeconfig token logic

a-blender:
## Problem

 
When using TF output or data resources that changes after the apply, all subsequent runs of `terraform plan` creates a new kubeconfig API token until `terraform apply` is run again to get the output and data changes. This was seen on both Rancher 2.6.9 and 2.6.10.

## Solution

After investigation, the root cause of this issue is likely that Terraform downloads / generates a kubeconfig on every run of a terraform plan or apply. TF replaces the kubeconfig token every time instead of using the token from the cached kubeconfig, which is causing the over generation of API tokens.

My solution is to update the `getClusterKubeconfig` logic [explained here](https://github.com/rancher/terraform-provider-rancher2/issues/841#issuecomment-1472784264) to use the API token from the cached kubeconfig (if it exists) instead of always replacing it.

### Manual Testing

@jakefhyde I was unable to reproduce this behavior again on 2.6.10 using the customer's lab setup, but was also not seeing the token over generation when testing on my fork.

Test plan
- [x] Verify new API tokens are *not* created on a terraform plan after the first terraform apply on latest rancher v2.6 with local fix
- [x] Verify same behavior on latest rancher v2.7 (rainbow)

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -697,10 +697,21 @@ func getClusterKubeconfig(c *Config, id, origconfig string) (*managementClient.G
 	if err != nil {
 		return nil, fmt.Errorf("Getting cluster Kubeconfig: %v", err)
 	}
+	// kubeconfig is not valid due to an invalid token. Use the cached kubeconfig
+	// and replace the token
+	if !kubeValid && len(token) == 0 {
+		newConfig, err := replaceKubeConfigToken(c, origconfig, token)
+		if err != nil {
+			return nil, err
+		}
+		origconfig = newConfig
+		kubeValid = true
+	}
 	if kubeValid {
 		return &managementClient.GenerateKubeConfigOutput{Config: origconfig}, nil
 	}
 
+	// kubeconfig is not valid for other reasons, download a new one
 	client, err := c.ManagementClient()
 	if err != nil {
 		return nil, fmt.Errorf("Getting cluster Kubeconfig: %v", err)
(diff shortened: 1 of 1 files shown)


rancher2/resource_rancher2_cluster.go
```diff
@@ -733,19 +744,12 @@ func getClusterKubeconfig(c *Config, id, origconfig string) (*managementClient.G
...
+			if err != nil {
+				if !IsNotFound(err) && !IsForbidden(err) && !IsServiceUnavailableError(err) {
+					return nil, fmt.Errorf("Getting cluster Kubeconfig: %v", err)
 				}
-				return kubeConfig, nil
-			}
-			if !IsNotFound(err) && !IsForbidden(err) && !IsServiceUnavailableError(err) {
-				return nil, fmt.Errorf("Getting cluster Kubeconfig: %v", err)
 			}
+			return kubeConfig, nil
```
jakefhyde: I think we still need to rety if the `err != nil` and its not one of the 3 error cases

rancher2/resource_rancher2_cluster.go
```diff
@@ -733,19 +744,12 @@ func getClusterKubeconfig(c *Config, id, origconfig string) (*managementClient.G
...
-			if err == nil {
-				if isRancher26 && len(token) > 0 {
-					newConfig, err := replaceKubeConfigToken(c, kubeConfig.Config, token)
-					if err != nil {
-						return nil, err
-					}
-					kubeConfig.Config = newConfig
+			if err != nil {
+				if !IsNotFound(err) && !IsForbidden(err) && !IsServiceUnavailableError(err) {
+					return nil, fmt.Errorf("Getting cluster Kubeconfig: %v", err)
```
bfbachmann: Minor: I'd just `%w` here just so the caller can check the error too if they want.
```suggestion
					return nil, fmt.Errorf("Getting cluster Kubeconfig: %w", err)
```

bfbachmann (approved):

jakefhyde (changes requested):

rancher2/resource_rancher2_cluster.go
a-blender: @jakefhyde That seems like more of an enhancement and not part of this bug fix, and we've never seen an issue with Terraform generating a new kubeconfig... unless you have? This block is also within a `context.Timeout` so it will wait up to the timeout for the HTTP response to resolve, I think. Should I still update this logic to retry instead? WDYT?

jakefhyde: You have to do error handling, in the previous code, `return kubeConfig, nil` was inside of the `if err == nil` block. You moved it outside of the error handling, but the if statement does not have the same semantic meaning, since you removed the `if err == nil` branch. Now it will always return the kubeconfig, with or without an error, also dropping the error, in which case it is likely nil and will probably cause the provider to panic. Beforehand it would drop into the select statement, now it's unreachable code from that branch.

jakefhyde (changes requested):

jakefhyde (approved):

PR #537, opened 2020-12-02
Retry enabling cluster monitoring if got apierr 500

rawmind0:
Closes https://github.com/rancher/rancher/issues/30188
Closes #546

PR #1159, opened 2023-06-27
[Q3] Add retry logic to rancher2_cluster update

a-blender:
## Problem

 
Error when upgrading rke cluster with Terraform 1.25

```
Error: Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [code=ServerError, message=Operation cannot be fulfilled on clusters.management.cattle.io "c-mhjrs": the object has been modified; please apply your changes to the latest version and try again, baseType=error]
```

which in this case, indicates an intermittent race condition on update of the `rancher2_cluster` resource.

## Solution

 
Add retry code to `rancher2_cluster` resource Update. I have chosen to use `resource.Retry` instead of `context.WithTimeout` like in [this code](https://github.com/rancher/terraform-provider-rancher2/pull/537/commits/657341e5972b5b3f2d4fa24bdca9be92d3e80601#) because we need a cluster update to retry until a non retryable error is returned or the timeout is reached. [WithTimeout](https://golang.cafe/blog/golang-context-with-timeout-example.html) controls the timeout of an HTTP response, say a 500, but we don't want the request cancelled if it fails. We want to retry several times, especially on upgrade since customers tend to get pretty frustrated if they have to run a `terraform apply` more than once.

I've also moved `updateClusterMonitoring` into a helper function to increase readability of the Update function and updated comments.

### Manual Testing

The race condition in the original issue is difficult to reproduce, but team discussion has concluded that adding retry code to `rancher2_cluster` Update will fix the failure on server error.

Test steps

- [x] Provision rke cluster on EC2 nodes with 1 cp/1 etcd/1 worker
- [x] Verify cluster provisions after one tf apply / no HTTP 500 error
- [x] Update rke cluster to increase worker count to 3 + add a fourth node pool
- [x] Verify cluster updates after one tf apply / no HTTP 500 error
- [x] Update rke cluster to `enable_cluster_monitoring = true` (preferably while the previous update is still running to simulate the race condition)
- [x] Verify cluster updates after one terraform apply (it took 27m for me!) / no HTTP 500 error

<details>

<summary>main.tf</summary>

```
terraform {
  required_providers {
    rancher2 = {
      source  = "terraform.local/local/rancher2"
      version = "1.0.0"
    }
  }
}
provider "rancher2" {
  api_url   = var.rancher_api_url
  token_key = var.rancher_admin_bearer_token
  insecure  = true
}
data "rancher2_cloud_credential" "rancher2_cloud_credential" {
  name = var.cloud_credential_name
}
resource "rancher2_cluster" "rancher2_cluster" {
  name = var.rke_cluster_name
  rke_config {
    kubernetes_version = "v1.26.4-rancher2-1"
    network {
      plugin = var.rke_network_plugin
    }
  }
}
resource "rancher2_node_template" "rancher2_node_template" {
  name = var.rke_node_template_name
  amazonec2_config {
    access_key           = var.aws_access_key
	  secret_key     = var.aws_secret_key
	  region         = var.aws_region
          ami            = var.aws_ami
	  security_group = [var.aws_security_group_name]
	  subnet_id      = var.aws_subnet_id
	  vpc_id         = var.aws_vpc_id
	  zone           = var.aws_zone_letter
	  root_size      = var.aws_root_size
	  instance_type  = var.aws_instance_type
  }
}
resource "rancher2_node_pool" "pool1" {
  cluster_id       = rancher2_cluster.rancher2_cluster.id
  name             = "pool1"
  hostname_prefix  = "tf-pool1-"
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = 1
  control_plane    = false
  etcd             = true 
  worker           = false 
}
resource "rancher2_node_pool" "pool2" {
  cluster_id       = rancher2_cluster.rancher2_cluster.id
  name             = "pool2"
  hostname_prefix  = "tf-pool2-"
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = 1
  control_plane    = true
  etcd             = false 
  worker           = false 
}
resource "rancher2_node_pool" "pool3" {
  cluster_id       = rancher2_cluster.rancher2_cluster.id
  name             = "pool3"
  hostname_prefix  = "tf-pool3-"
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = 1
  control_plane    = false
  etcd             = false 
  worker           = true 
}
```

</details>

<details>

<summary>main.tf (sections with cluster updates)</summary>

```
resource "rancher2_cluster" "rancher2_cluster" {
  name = var.rke_cluster_name
  rke_config {
    kubernetes_version = "v1.26.4-rancher2-1"
    network {
      plugin = var.rke_network_plugin
    }
  }
  enable_cluster_monitoring = true
}
.
.
.
resource "rancher2_node_pool" "pool3" {
  cluster_id       = rancher2_cluster.rancher2_cluster.id
  name             = "pool3"
  hostname_prefix  = "tf-pool3-"
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = 3
  control_plane    = false
  etcd             = false 
  worker           = true 
}
resource "rancher2_node_pool" "pool4" {
  cluster_id       = rancher2_cluster.rancher2_cluster.id
  name             = "pool4"
  hostname_prefix  = "tf-pool4-"
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = 1
  control_plane    = false
  etcd             = true
  worker           = true
}
```

</details>

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -192,7 +192,7 @@ func resourceRancher2ClusterCreate(d *schema.ResourceData, meta interface{}) err
 			Links:   newCluster.Links,
 			Actions: newCluster.Actions,
 		}
-		// Retry enabling monitoring until timeout if got apierr 500 https://github.com/rancher/rancher/issues/30188
+		// Retry enable monitoring until timeout if got api error 500
 		ctx, cancel := context.WithTimeout(context.Background(), meta.(*Config).Timeout)
 		defer cancel()
 		for {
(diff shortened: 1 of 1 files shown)


rancher2/resource_rancher2_cluster.go
```diff
@@ -755,6 +719,73 @@ func getClusterKubeconfig(c *Config, id, origconfig string) (*managementClient.G
 	}
 }
 
+func updateClusterMonitoring(d *schema.ResourceData, meta interface{}, newCluster Cluster) error {
+	client, err := meta.(*Config).ManagementClient()
```
HarrisonWAffel: The use of `meta.(*Config).ManagementClient()`  here creates an entirely new client which results in a few HTTP requests being made against the Rancher API. I think it would make more sense to reuse the client defined in the calling function, `resourceRancher2ClusterUpdate` on line 280. let me know what you think, or if I'm mistaken. I'm sure the additional http requests don't have too much impact, but every little bit helps

HarrisonWAffel (approved):
One small nit about not reusing a client, but otherwise lgtm

rancher2/resource_rancher2_cluster.go
```diff
@@ -385,93 +385,57 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
...
-						return err
-					}
-				}
-				err = client.APIBaseClient.Action(managementClient.ClusterType, monitoringActionEdit, clusterResource, monitoringInput, nil)
-				if err != nil {
-					return err
-				}
+		if err != nil {
+			if IsServerError(err) {
+				return nil
```
jiaqiluo: Q: Should we return the error or warp it as a RetryableError here so that a retry will be kicked off? 
Sorry if I misunderstand the code here.

jiaqiluo (changes requested):

thatmidwesterncoder (approved):

rancher2/resource_rancher2_cluster.go
a-blender: Nil _will_ kick off a retry. Anything except for a `resource.Retry` timeout or a `NonRetryableError` will retry the API call. My approach is mvp by only changing code if necessary. I suppose I could make it more explicit by returning a `RetryableError` - thoughts?

jiaqiluo: The `resource.Retry` method will return success if no error is returned from the argument function. If we want to retry on server errors, the argument function should return a `RetryableError` instead of `nil`.

a-blender: You, are right and probably should only return a nil error if it was nil. Good catch

rancher2/resource_rancher2_cluster.go
```diff
@@ -385,93 +385,57 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
...
+			return resource.NonRetryableError(fmt.Errorf("[ERROR] waiting for cluster (%s) to be updated: %s", newCluster.ID, waitErr))
+		}
+
+		// update cluster monitoring if it has changed
+		if d.HasChange("enable_cluster_monitoring") || d.HasChange("cluster_monitoring_input") {
+			err = updateClusterMonitoring(client, d, meta, *newCluster)
 			if err != nil {
-				return err
+				if IsServerError(err) {
+					return &resource.RetryError{Err: err}
```
jiaqiluo: eithter
```suggestion
					return &resource.RetryableError(err)
```
or 
```suggestion
					return &resource.RetryError{Err: err, Retryable: true}
```

jiaqiluo (changes requested):

jiaqiluo (approved):

Issue #1074, opened 2023-02-14
[BUG] kubelet-args are not supported via TF


boris-stojnev:
**Rancher Server Setup**
- Rancher version: v2.7.0
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version: 3-node cluster, rancher-stable, v1.24.9+rke2r1
- Proxy/Cert Details: no proxy, self-signed

**Information about the Cluster**
- In process of creating downstream custom cluster

**User Information**
- What is the role of the user logged in?: Admin

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? 
   - 1.25.0
- What is the version of Terraform in use? 
   - v1.2.9

### Describe the bug
Using kubelet-arg via Terraform rancher2 provider resulting in unexpected behavior of cluster not starting. 
This bug is described here https://github.com/rancher/rancher/issues/38112
I think this is specific to the Terraform provider, not the Rancher itself.

Kubelet-arg which is part of `config` under `machine_selector_config` should be list type not string.  In other words, `config` should be able to accept list types.

a-blender:
@boris-stojnev I see the associated issue has been closed https://github.com/rancher/rancher/issues/38112 with the workaround of setting `kubelet_arg` under `machine_global_config` in your Terraform config which then causes the args to successfully get pushed down to the nodes. For v2 prov clusters, [machine_global_config](https://github.com/rancher/terraform-provider-rancher2/blob/c677294f40087998e0194142eac56b5277d4ccaf/rancher2/schema_cluster_v2_rke_config.go#L63C9-L63C9) is a string that must be provided in YAML format - here's [an example](https://github.com/rancher/terraform-provider-rancher2/blob/c677294f40087998e0194142eac56b5277d4ccaf/docs/resources/cluster_v2.md?plain=1#L146-L150).

```
machine_global_config = <<EOF
  config:
    cloud-provider-name: "external"
    kubelet-arg: "key1=value1,key2=value2"
EOF
```

Does this work for you?

a-blender:
@boris-stojnev Any update on this issue?

boris-stojnev:
@a-blender sorry for the late reply.

My workaround was placing `kubelet-arg` in config.yaml leveraging Ansible with other additional configuration I have.

[...]

PR #1168, opened 2023-07-11
Update `rancher2_cluster_v2.rke_config.machine_selector_config.config` to optional/computed

slicen:
Related Issue: https://github.com/rancher/terraform-provider-rancher2/issues/1074

## Problem
It is currently not possible to supply a compute value for the `rancher2_cluster_v2.rke_config.machine_selector_config.config` argument. It is common to require this, for instance when supplying a kubeconfig retrieved from a previous Terraform resource (e.g. when creating clusters using the Harvester provider).

Making this field computed will enable argument map entries (such as `cloud-provider-config`) to be computed at apply time.
 
## Solution
This change simply makes the `rancher2_cluster_v2.rke_config.machine_selector_config.config` field optional/computed.
 
## Testing
1. Take unmodified code base.
2. Create a test `rancher2_cluster_v2` resource with `rke_config.machine_selector_config.config` field set to the value of another resource.
3. Observe that the argument's value is null at plan and apply-time.
4. Apply this PR and re-run the plan and apply.
5. Observe that the value is listed as '(known after apply)' at plan-time, and is set correctly during apply.

### Manual Testing
Conducted test steps above.

### Automated Testing
No automated test changes required.

## QA Testing Considerations
No special QA test cases required.
 
Unlikely to have high chance of regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ FEATURES:
 
 ENHANCEMENTS:
 
-
+* Updated `rancher2_cluster_v2.rke_config.machine_selector_config.config` argument as optional/computed
 
 BUG FIXES:
 
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -584,7 +584,7 @@ The following attributes are exported:
 ##### Arguments
 
 * `machine_label_selector` - (Optional) Machine selector label (list maxitems:1)
-* `config` - (Optional) Machine selector config (map)
+* `config` - (Optional/Computed) Machine selector config (map)
 
 ##### `machine_label_selector`
 
--- a/rancher2/schema_cluster_v2_rke_config_system_config.go
+++ b/rancher2/schema_cluster_v2_rke_config_system_config.go
@@ -65,6 +65,7 @@ func clusterV2RKEConfigSystemConfigFields() map[string]*schema.Schema {
 		"config": {
 			Type:        schema.TypeMap,
 			Optional:    true,
+			Computed:    true,
 			Description: "Machine selector config",
 		},
 	}


a-blender (approved):

jiaqiluo:
~hi @a-blender ,  do we need to add the same fix to  [this function](https://github.com/rancher/terraform-provider-rancher2/blob/9a5fa08d1ed6a24f0482bd8e21bf8f5dc7c6d9d2/rancher2/schema_cluster_v2_rke_config_system_config.go#L68-L73) ?~

a-blender:
@jiaqiluo I think it'd be a good idea yes, so users on the old schema will be able to apply Computed values if they want to.

jiaqiluo:
@a-blender, I think about this twice and realize that even if it will work it is not a good practice to modify a released version of the schema. Please ignore my previous suggestion.

a-blender:
@jiaqiluo I see your point.

I do want to say though, that I don't think adding `Computed` to the old schema would cause a behavior regression since it's just allowing a user to add values not known at the time of apply [docs - schema behaviors](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors), not changing the type. I'm fine to leave this as is! Could you approve?

jiaqiluo:
@a-blender, yeah, it makes sense. Thank you for sharing the docs.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ FEATURES:
 
 ENHANCEMENTS:
 
-
+* Updated `rancher2_cluster_v2.rke_config.machine_selector_config.config` argument as optional/computed
```
jiaqiluo, line 9: This can be dropped for two reasons: 1) this file is outdated, 3.0.1 was already released 2) we will add this info when releasing the next version.

jiaqiluo (changes requested):

a-blender:
No problem! @slicen can you please make the requested changes?

a-blender:
@slicen What's the status on this? Can you make the requested changes/rebase please?

kkaempf:
closing stale PRs.

PR #1181, opened 2023-07-28
Fix machine selector config to allow kubelet arg list

a-blender:
## Problem

Machine Selector Config `kubelet-arg` values could not be passed via Terraform to downstream machines and would not show up in the rancher UI. 

## Solution

 
My solution is to convert `machine_selector_config.config` from Type.Map (string map) to Type.String that supports yaml like Machine Global Config + state migration logic to handle the schema update. This allows for users to input both strings and lists, which allows subfield `kubelet-arg` to be passed as a list which is what Rancher is expecting. It also prevents a potential regression caused by 1) only allowing string args or lists which will reduce the flexibility of Machine Selector Config or 2) defining kubelet-arg as a subfield which will break other config fields that are already expected in the Rancher backend such as `private-default-registry.`

Before, passing `kubelet-arg` under Machine Selector Config would not work. This feature now works with the following configuration

```
machine_selector_config {
  config = <<EOF
    protect-kernel-default: true
    kubelet-arg:
      - key1=value1
      - key2=value2
  EOF
}
```

### Manual Testing

Test plan
- [x] Prov rke2 cluster with Machine Selector Config and 2 kubelet args set. Verify cluster provisions successfully and args are displayed correctly in the rancher ui
- [x]  Add/remove a kubelet-arg via tf and verify list is updated in the ui
- [x] Prov rke2 cluster with tf 3.1.0. Add machine selector config with 2 kubelet args via the rancher ui
- [x] Upgrade tf to rc version with fix
- [x] Verify no errors appear on `terraform plan/apply` or updates to Machine Selector Config.

### Automated Testing

Tests updated. Ran `go test -v ./rancher2` to make sure all automated tests pass.

### DIFF
--- a/rancher2/resource_rancher2_cluster_v2.go
+++ b/rancher2/resource_rancher2_cluster_v2.go
@@ -22,11 +22,18 @@ func resourceRancher2ClusterV2() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceRancher2ClusterV2Import,
 		},
-		Schema: clusterV2Fields(),
+		Schema:        clusterV2Fields(),
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceRancher2ClusterV2Resource().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceRancher2ClusterV2StateV0,
+				Version: 0,
+			},
+		},
 		CustomizeDiff: func(d *schema.ResourceDiff, i interface{}) error {
 			if d.HasChange("rke_config") {
 				oldObj, newObj := d.GetChange("rke_config")
-				//return fmt.Errorf("\n%#v\n%#v\n", oldObj, newObj)
 				oldInterface, oldOk := oldObj.([]interface{})
 				newInterface, newOk := newObj.([]interface{})
 				if oldOk && newOk && len(newInterface) > 0 {
@@ -41,7 +48,6 @@ func resourceRancher2ClusterV2() *schema.Resource {
 			}
 			if d.HasChange("local_auth_endpoint") {
 				oldObj, newObj := d.GetChange("local_auth_endpoint")
-				//return fmt.Errorf("\n%#v\n%#v\n", oldObj, newObj)
 				oldInterface, oldOk := oldObj.([]interface{})
 				newInterface, newOk := newObj.([]interface{})
 				if oldOk && newOk && len(newInterface) > 0 {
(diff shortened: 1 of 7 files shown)


rancher2/schema_cluster_v2_rke_config_system_config.go
```diff
@@ -51,6 +54,47 @@ func clusterV2RKEConfigSystemConfigLabelSelectorFields() map[string]*schema.Sche
 	return s
 }
 
+func clusterV2RKEConfigSystemConfigFieldsV0() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
```
felipe-colussi, line 58: I do not know how TF works, so I can be wrong here.

Ain't this function exactly the same as  clusterV2RKEConfigSystemConfigFields() ?

If names are important,  shouldn't u have a base function and wrap it ?

This is the same for clusterV2RKEConfigFieldsV0() and clusterV2FieldsV0()

HarrisonWAffel (approved):

rancher2/resource_rancher2_cluster_v2.go
```diff
@@ -64,6 +70,37 @@ func resourceRancher2ClusterV2() *schema.Resource {
...
+					// upgrade all machine selector configs
+					for m := range machineSelectorConfigs {
+						if machineSelectorConfig, ok := machineSelectorConfigs[m].(map[string]any); ok && len(machineSelectorConfig) > 0 {
+
+							// machine selector config data found. Migrate state from map -> list
+							if config, ok := machineSelectorConfig["config"].([]any); ok {
+								var newValue []any
+								newValue = append(newValue, config)
+								rawState["rke_config"].([]interface{})[i].(map[string]any)["machine_selector_config"].([]any)[m].(map[string]any)["config"] = newValue
+							}
```
felipe-colussi: I may be wrong, but this dosen't cast  from a Map to a list. (at leas not a go map to a go list). 

If "config" is a map it will be ignored. 

If "config" already is a list/slice this function will take place and replace something like:`...config := [1, 2, 3]`  to `config = [[1,2,3]]`

Is this the idea of the function ?

rancher2/resource_rancher2_cluster_v2.go
```diff
@@ -64,6 +70,37 @@ func resourceRancher2ClusterV2() *schema.Resource {
...
 
+func resourceRancher2ClusterV2Resource() *schema.Resource {
+	return &schema.Resource{
+		Schema: clusterV2FieldsV0(),
+	}
+}
+
+func resourceRancher2ClusterV2StateV0(rawState map[string]any, meta interface{}) (map[string]any, error) {
+	if rkeConfigs, ok := rawState["rke_config"].([]any); ok && len(rkeConfigs) > 0 {
+		for i := range rkeConfigs {
```
felipe-colussi, line 81: Is there another place that checks if the passed struct is ok?  This if it isn't this function will just ignore it.

rancher2/schema_cluster_v2_rke_config_system_config.go
a-blender: Good question :) yes, they are the same but there need to be two schema definitions for the state migration of migration `machine_selector_config.config` from a `Type.Map` to a `Type.String` that supports yaml. That way the top level RKE Config schema will reference the correct types both on cluster creation and on an upgrade of the provider. Check out the docs on [Terraform v0.12 State Migrations](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/state-migration) if you are curious.

rancher2/resource_rancher2_cluster_v2.go
a-blender: Oh actually, I need to update the state migration logic since I adjusted the fix. This should be migration from a map -> string, I'll update that.

rancher2/resource_rancher2_cluster_v2.go
```diff
@@ -64,6 +70,37 @@ func resourceRancher2ClusterV2() *schema.Resource {
 	}
 }
 
+func resourceRancher2ClusterV2Resource() *schema.Resource {
+	return &schema.Resource{
+		Schema: clusterV2FieldsV0(),
+	}
+}
+
+func resourceRancher2ClusterV2StateV0(rawState map[string]any, meta interface{}) (map[string]any, error) {
```
jakefhyde: Good use of `any`.

jakefhyde (dismissed):
Consider this a lgtm once @felipe-colussi's comments are addressed.

felipe-colussi (approved):

PR #1213, opened 2023-08-31
change default ubuntu image from DO

felipe-colussi:
## Problem

 Wile testing the migration of SDK to SDK2 I noticed that the default image for DO is no longer valid. 

`
[tf-do-image.zip](https://github.com/rancher/terraform-provider-rancher2/files/12486407/tf-do-image.zip)
Error creating machine: Error in driver during machine creation: POST https://api.digitalocean.com/v2/droplets: 422 You specified an invalid image for Droplet creation.`
 
 After checking DO only supports ubuntu 20 and 22. 
 
## Solution

 Update the DO default image to: ubuntu-22-04-x64
 
 
### Manual Testing

Using the terraform as it is: 
Tried to generate a UBUNTU RKE cluster using the default image.  It fails.
Change the image for `ubuntu-22-04-x64` , it works. 

After the change: 
I ran the  tf with an empty image. It worked. 

TF files:
[tf-do-image.zip](https://github.com/rancher/terraform-provider-rancher2/files/12486420/tf-do-image.zip)

### Automated Testing

I ran the automated tests. 
That said I didn't create one to test the "Default" value provided by TF SDK.

### DIFF
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -191,7 +191,7 @@ The following attributes are exported:
 
 * `access_token` - (Optional/Sensitive) Digital Ocean access token. Mandatory on Rancher v2.0.x and v2.1.x. Use `rancher2_cloud_credential` from Rancher v2.2.x (string)
 * `backups` - (Optional) Enable backups for droplet. Default `false` (bool)
-* `image` - (Optional) Digital Ocean Image. Default `ubuntu-16-04-x64` (string)
+* `image` - (Optional) Digital Ocean Image. Default `ubuntu-22-04-x64` (string)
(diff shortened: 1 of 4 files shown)


jakefhyde (approved):

a-blender (approved):
@felipe-colussi Prov an RKE cluster on DO didn't give you any ssh/networking issues like you saw before? Not sure if the fix for this is a fix for what you were seeing a few days ago. Otherwise great catch!

felipe-colussi:
@a-blender 
> @felipe-colussi Prov an RKE cluster on DO didn't give you any ssh/networking issues like you saw before? Not sure if the fix for this is a fix for what you were seeing a few days ago. Otherwise great catch!

Nop, that was happening on AWS, so I swapped to DO and got this image error :/

PR #1221, opened 2023-09-06
modify Jenkinsfile + Dockerfile for `terraform-provider-rancher2/tfp-automation` tests to run in our Jenkins env

Josh-Diamond:
integrate `rancher/terraform-provider-rancher2/tfp-automation` tests into Jenkins env, as part of https://github.com/rancher/qa-tasks/issues/386

### DIFF
--- a/tfp-automation/Jenkinsfile
+++ b/tfp-automation/Jenkinsfile
@@ -1,26 +1,30 @@
-pipeline {
-  agent any
-
-  stages {
-    stage('Build Docker image') {
-      steps {
-        script {
-            writeFile file: 'config.yml', text: params.CONFIG
-            env.CATTLE_TEST_CONFIG='/home/jenkins/workspace/rancher_qa/tfp-automation/config.yml'
-            sh 'docker build --build-arg CONFIG_FILE=config.yml -t tfp-automation .'
+#!groovy
+node {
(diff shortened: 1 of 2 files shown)


markusewalker (approved):
LGTM

tfp-automation/Jenkinsfile
```diff
@@ -1,26 +1,29 @@
-pipeline {
-  agent any
-
-  stages {
+node {
```
slickwarren, line 2: I don't think things break if you omit this, but all our other jenkinsfiles have the shebang `#!groovy` as the first line of each file.

slickwarren (commented):
small nit, everything else lgtm

tfp-automation/Jenkinsfile
Josh-Diamond: I had that initially, and it threw an error so i took it out, i'll add it back rn and re-test this and will DM you results 👍

Josh-Diamond: not hitting the error today, so likely encountered that error yesterday due to some syntax issue, but not reproducing it now, so  I'll update PR to include `#!groovy` 👍

slickwarren (approved):

PR #1223, opened 2023-09-08
Fix msc old schema to have correct type

a-blender:
## Solution

- [x] Update the `machine_selector_config.config` old schema to original type Map since it will only be used in the schema migration on tf upgrade. Won't affect newly created clusters.
- [x] Update docs
 
### Manual Testing

Tested provisioning an RKE2 cluster on the latest version of tf rancher2 with MachineConfigSelector with config defined. Successful, new schema is used, no errors.

Tested upgrade case from 3.1.0 -> latest. State is upgraded to the correct type.

Example

Old state
```
"machine_selector_config": [
                  {
                    "config": {
                      "protect-kernel-default": "true"
                    },
                    "machine_label_selector": []
                  }
                ],
```

New state
```
"machine_selector_config": [
                  {
                    "config": "protect-kernel-default: \"true\"\n",
                    "machine_label_selector": []
                  }
                ],
```

### DIFF
--- a/docs/resources/cluster_v2.md
+++ b/docs/resources/cluster_v2.md
@@ -15,7 +15,7 @@ Provides a Rancher v2 Cluster v2 resource. This can be used to create RKE2 and K
 # Create a new rancher v2 RKE2 custom Cluster v2
 resource "rancher2_cluster_v2" "foo" {
   name = "foo"
-  kubernetes_version = "<RANCHER_KUBERNETES_VERSION>"
+  kubernetes_version = "rancher-kubernetes-version"
   fleet_namespace = "fleet-ns"
   enable_network_policy = false
   default_cluster_role_for_project_members = "user"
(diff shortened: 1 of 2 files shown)


HarrisonWAffel:
@a-blender does this mean that prior versions of the provider which used TypeString for this field are not properly migrating?  or have we not released a version where the field uses TypeString?

a-blender:
@a-blender Possible, it worked fine before but I had tested the old TF version by adding MSC kubelet args _via the rancher UI_ because this field originally was broken in the provider and wasn't passing values correctly so it didn't make logical sense to test a use case that users couldn't even use themselves. I re-tested that a migration works correctly via TF input only for this.

HarrisonWAffel (approved):

thatmidwesterncoder (approved):

PR #1224, opened 2023-09-08
[DNM until after 2.8 release] Migrate from sdk-v1 to sdk2-v2

felipe-colussi:
# DNM:  waiting for release of 4.0.0

## Problem

Migrate SDK1 to SDK2
 
## Solution

Migrate SDK1 to SDK2:
Main changes: 
* Change Create, Read, Update, Delete fields to CreateContext, ReadContext . . . 

* Change function  signatures from  `dataSourceRancher2....( d *schema.ResourceData, meta interface{})` error  to  `dataSourceRancher2...(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics`

* Change some error handling to behave with `diag.Diagnostics`.

* Remove `MaxItems` from  TF arguments with `Compiled: true`. Required by migrating sdk1 to 2. 

* Add some argument that were being set without being on the schema. Required by migrating sdk1 to 2. 

* Changed some types and added some so they correspond to the schemas. Required by migrating sdk1 to 2.  

Observation: 
I had a loot of errors on the Tests, mainly 401 during the delete of TestAccRancher2Upgrade.  I Did 2 commits to add some debug logic to that test and on that moment It just worked.  I'm letting the debug (prints) on this PR so if I go back having that error I can confirm if the envs were being overwritten without need (I do believe that this was causing the error). 

 
### Manual Testing

I Tested on TF trying to create, destroy scale and change objects. 

[DO-RKE2-Basic.txt](https://github.com/rancher/terraform-provider-rancher2/files/12728372/DO-RKE2-Basic.txt)
[AWS-RKE2-CUSTOM.txt](https://github.com/rancher/terraform-provider-rancher2/files/12728373/AWS-RKE2-CUSTOM.txt)
[aws-rke1-custom.txt](https://github.com/rancher/terraform-provider-rancher2/files/12728377/aws-rke1-custom.txt)

### Automated Testing

I ran all automated tests, also had to change some to match the SDK upgrade.

## QA Testing Considerations

This is a huge change, I wasn't able to get a bug that happened due to the migration, that said  please test this intensively.


rancher2/data_source_rancher2_notifier.go
```diff
@@ -45,46 +46,41 @@ func dataSourceRancher2Notifier() *schema.Resource {
 				},
 			},
 			"pagerduty_config": {
-				Type:          schema.TypeList,
-				MaxItems:      1,
-				Computed:      true,
-				ConflictsWith: []string{"dingtalk_config", "msteams_config", "smtp_config", "slack_config", "webhook_config", "wechat_config"},
+				Type:     schema.TypeList,
+				Computed: true,
+				//ConflictsWith: []string{"dingtalk_config", "msteams_config", "smtp_config", "slack_config", "webhook_config", "wechat_config"},
```
a-blender: Were the `//ConflictsWith` auto added to the schema as part of the upgrade to v2? I don't think we have those other conflicting fields defined. Can you remove these?

rancher2/import_rancher2_catalog_v2.go
```diff
@@ -1,17 +1,20 @@
...
+func resourceRancher2CatalogV2Import(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
 	clusterID, name := splitID(d.Id())
 	d.Set("cluster_id", clusterID)
 	d.Set("name", name)
 
-	err := resourceRancher2CatalogV2Read(d, meta)
-	if err != nil || d.Id() == "" {
-		return []*schema.ResourceData{}, err
+	diag := resourceRancher2CatalogV2Read(ctx, d, meta)
+	if diag.HasError() || d.Id() == "" { // TODO - Check this if, it can break, didn't change the logic,.
```
a-blender: We're making an assumption here that is `d.Id` (the cluster id) is empty on Read then there must be an error but since `diag[0]` could crash the provider I'd play it safe and only include `diag[0]` in the error message if that element has been checked and exists. Apply to all instances, remove TODO

rancher2/data_source_rancher2_cluster.go
```diff
@@ -205,6 +205,42 @@ func dataSourceRancher2Cluster() *schema.Resource {
 				Type:     schema.TypeMap,
 				Computed: true,
 			},
+			/*
+				TODO - ANDY I added the following fields here cause v2 was panicking. AS they didn't exists I used computed.
```
a-blender: Lgtm. I'd adjust this comment to not include the actual panic message/source code files but just say

```
// Andy and I added the following fields because the upgrade to sdk v2 required they be present in the data source for rancher2_cluster and v2 cluster prov to work, etc.
```

rancher2/data_source_rancher2_cluster.go
```diff
@@ -216,10 +216,16 @@ func dataSourceRancher2Cluster() *schema.Resource {
 			"cluster_agent_deployment_customization": {
 				Type:     schema.TypeList,
 				Computed: true,
+				Elem: &schema.Schema{
```
a-blender: This is incorrect. The type of `cluster_agent_deployment_customization` is a Type.List with sub fields https://github.com/rancher/terraform-provider-rancher2/blob/9a5fa08d1ed6a24f0482bd8e21bf8f5dc7c6d9d2/rancher2/schema_cluster.go#L332-L334

rancher2/data_source_rancher2_cluster_v2.go
```diff
@@ -109,14 +109,23 @@ func dataSourceRancher2ClusterV2() *schema.Resource {
...
 				Type:     schema.TypeList,
 				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
 			},
 			"fleet_agent_deployment_customization": {
 				Type:     schema.TypeList,
 				Computed: true,
+				Elem: &schema.Schema{
```
a-blender: Same with `fleet_agent_deployment_customization.` Fix the type https://github.com/rancher/terraform-provider-rancher2/blob/9a5fa08d1ed6a24f0482bd8e21bf8f5dc7c6d9d2/rancher2/schema_cluster.go#L340-L342

main.go
```diff
@@ -9,14 +9,12 @@ import (
 
 func main() {
 	var debug bool
-
```
a-blender: Why are you removing lines to force a build? You can build the repo with `make bin` so I don't understand why there are multiple commits doing this

rancher2/structure_agent_deployment_customization_v2.go
```diff
@@ -21,11 +21,13 @@ func flattenAgentDeploymentCustomizationV2(in *v1.AgentDeploymentCustomization)
...
 	if in.OverrideAffinity != nil {
-		obj["override_affinity"] = in.OverrideAffinity
+		b, _ := interfaceToJSON(in.OverrideAffinity)
+		_ = b
+		obj["override_affinity"] = b // TODO - ANDY, changed it to keep it simpler instead of creating a new schema.  The expand reads a json
 	}
 
 	if in.OverrideResourceRequirements != nil {
-		obj["override_resource_requirements"] = in.OverrideResourceRequirements
+		obj["override_resource_requirements"] = flattenResourceRequirementsV2(in.OverrideResourceRequirements) // TODO - ANDY added a flatten
```
a-blender: Good catch!

rancher2/structure_agent_deployment_customization_v2.go
```diff
@@ -21,11 +21,13 @@ func flattenAgentDeploymentCustomizationV2(in *v1.AgentDeploymentCustomization)
 	}
 
 	if in.OverrideAffinity != nil {
-		obj["override_affinity"] = in.OverrideAffinity
+		b, _ := interfaceToJSON(in.OverrideAffinity)
+		_ = b
+		obj["override_affinity"] = b // TODO - ANDY, changed it to keep it simpler instead of creating a new schema.  The expand reads a json
```
a-blender: 1. That's fine but please be descriptive with variables. `_` is the error returned from `interfaceToJSON` and we shouldn't be ignoring it so return an error

```
overrideAffinityStr, err := interfaceToJSON(in.OverrideAffinity)
if err != nil {
  return nil, err
}
obj["override_affinity"] = overrideAffinityStr
```

2. What were your results of testing a v2 cluster with cluster agent customization defined?

rancher2/structure_cluster_v2.go
```diff
@@ -60,10 +60,18 @@ func flattenClusterV2(d *schema.ResourceData, in *ClusterV2) error {
 		d.Set("cloud_credential_secret_name", in.Spec.CloudCredentialSecretName)
 	}
 	if in.Spec.ClusterAgentDeploymentCustomization != nil {
-		d.Set("cluster_agent_deployment_customization", flattenAgentDeploymentCustomizationV2(in.Spec.ClusterAgentDeploymentCustomization))
+		flat, err := flattenAgentDeploymentCustomizationV2(in.Spec.ClusterAgentDeploymentCustomization)
```
a-blender: Nit: I know it seems verbose but be descriptive so I suggest

```
agentDeploymentCustomizationV2, err := flattenAgentDeploymentCustomizationV2(in.Spec.ClusterAgentDeploymentCustomization)
```

rancher2/0_provider_upgrade_test.go
```diff
@@ -427,15 +427,21 @@ func testAccRancher2UpgradeRancher() resource.TestCheckFunc {
...
 			if rs.Type != "rancher2_bootstrap" {
 				continue
 			}
-
 			token := rs.Primary.Attributes["token"]
-			os.Setenv("RANCHER_TOKEN_KEY", token)
+			rancherTokenKey := os.Getenv("RANCHER_TOKEN_KEY")
+			rancherAdminPass := os.Getenv("RANCHER_ADMIN_PASS")
+			if err := os.Setenv("RANCHER_TOKEN_KEY", token); err != nil {
+				fmt.Printf("Failed to update RANCHER_TOKEN_KEY on ressource %s with err %s", k, err.Error())
```
a-blender: Nit: `resource` instead of `ressource`. Fix all instances and clean up error messages

a-blender (changes requested):

a-blender:
@felipe-colussi In general, nice job! Had some comments following up our discussion on the source code and schema updates last week. Please do the following

* Remove all comments/TODO comments left in the source code
* Clean up the error messages I referenced
* Post more details in `Manual Testing` (types of cluster tested, test upgrade tf from old version to version with SDK v2 updates, tf config files)
* Post test results for provisioning an RKE and v2 cluster with `cluster_agent_deployment_customization` and `fleet_agent_deployment_customization` defined. Verify it works and is showing the correct values in the tf state and cluster mgmt yaml

rancher2/0_provider_upgrade_test.go
```diff
@@ -427,15 +427,21 @@ func testAccRancher2UpgradeRancher() resource.TestCheckFunc {
...
+			}
 			currentPassword := rs.Primary.Attributes["current_password"]
-			os.Setenv("RANCHER_ADMIN_PASS", currentPassword)
+			if err := os.Setenv("RANCHER_ADMIN_PASS", currentPassword); err != nil {
+				fmt.Printf("Failed to update RANCHER_ADMIN_PASS on ressource %s with err %s", k, err.Error())
+			}
+			fmt.Printf("Changing env on ressource %s: RANCHER_TOKEN_KEY: %s -> %s; RANCHER_ADMIN_PASS: %s -> %s", k, rancherTokenKey, token, rancherAdminPass, currentPassword)
```
jakefhyde: I feel like this is going to log the token and password to CI, which is probably not what we want. We can still log an error/print the fact that it was updated, just not a statement that exposes the values.

rancher2/data_source_rancher2_cluster.go
```diff
@@ -38,79 +39,69 @@ func dataSourceRancher2Cluster() *schema.Resource {
 			},
 			"rke_config": {
 				Type:     schema.TypeList,
-				MaxItems: 1,
```
jakefhyde, line 41: Why were max items removed here?

rancher2/data_source_rancher2_project.go
```diff
@@ -1,17 +1,19 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/rancher/norman/types"
 )
 
 func dataSourceRancher2Project() *schema.Resource {
 	return &schema.Resource{
-		Exists: dataSourceRancher2ProjectExists,
-		Read:   dataSourceRancher2ProjectRead,
+		Exists:      dataSourceRancher2ProjectExists, // TODO Check if I can delete.
```
jakefhyde: Just marking TODO so this is not forgotten. I am reviewing per commit.

rancher2/data_source_rancher2_notifier.go
felipe-colussi: The conflict With always existed, I commented it Cause the SDK-V2 complains about a non configurable field having the ConfigWith passed by. 

* data source rancher2_notifier: webhook_config: ConflictsWith is for configurable attributes,there's nothing to configure on computed-only field

I'll delete them

rancher2/import_rancher2_catalog_v2.go
felipe-colussi: After talking with you I noticed that It could panic. 

I changed that Logic to : 

```
diag := resourceRancher2CatalogV2Read(ctx, d, meta)
	if diag.HasError() {
		return []*schema.ResourceData{}, errors.New(diag[0].Summary)
	}
	if d.Id() == "" {
		return []*schema.ResourceData{}, nil
	}

```

That way it matches the old behavior.

rancher2/data_source_rancher2_cluster.go
felipe-colussi: Yep, Those fields weren't set And they were pointing errors, I created all of them as strings then I fixed it. 
Already was fixed. 

```
		"fleet_agent_deployment_customization": {
				Type:        schema.TypeList,
				Optional:    true,
				Description: "Optional customization for fleet agent",
				Elem: &schema.Resource{
					Schema: agentDeploymentCustomizationFields(),
				},
			},

```

rancher2/import_rancher2_catalog_v2.go
a-blender: Oh yeah, I see that in the new code/later commits. Lgtm thank you

rancher2/data_source_rancher2_cluster_v2.go
felipe-colussi: Same as the previous one, They were pointing all missing fields at the 1st error, so I found easier to 1st set them all to strings then fix them. 

Current code is: 

```

			"cluster_agent_deployment_customization": {
				Type:        schema.TypeList,
				Optional:    true,
				Description: "Optional customization for cluster agent",
				Elem: &schema.Resource{
					Schema: agentDeploymentCustomizationFields(),
				},
			},
			"fleet_agent_deployment_customization": {
				Type:        schema.TypeList,
				Optional:    true,
				Description: "Optional customization for fleet agent",
				Elem: &schema.Resource{
					Schema: agentDeploymentCustomizationFields(),
				},
			},

```

main.go
felipe-colussi: Dumb me. I didn't know if Drone was keeping something in memory (from a docker build or something like that), I had a problem with that on an old company that we used GitlabCi,  After some flaky tests I decided to do that for testing.  

I was wrong, the flaky was being caused by a race condition that was solved after adding a print to debug it :/.

rancher2/structure_agent_deployment_customization_v2.go
felipe-colussi: Thx, but it wasn't me. The SDK-V2 really don't like missing fields or fields with wrong values. 
The v1 would just ignore it but V2 throws an error on our face : D

rancher2/structure_agent_deployment_customization_v2.go
felipe-colussi: 1.  That was testing for testing, i wasn't supposed to commit it.   
Current code is: 
```
		b, err := interfaceToJSON(in.OverrideAffinity)
		if err != nil {
			return []interface{}{}, err
		}
		obj["override_affinity"] = b

```

a-blender: Gotcha, go ahead and remove it. Could you also post some brief results of testing a v2 cluster with cluster agent customization? I sent you a test tf config for that :) that was a large feature so I wanna make sure everything's still working before we merge.

main.go
a-blender: No worries, yep! For visibility, we have an open tf issue https://github.com/rancher/rancher/issues/42212 to deal with flaky tests so hopefully those race conditions can get fixed. Glad it's solved :)

rancher2/structure_agent_deployment_customization_v2.go
felipe-colussi: I'm trying to . . . 
I'm having some issues runing rancher locally from v2.7 and from 2.8 :/  

As soon as I'm able to I'll.

rancher2/structure_cluster_v2.go
felipe-colussi: Done.

a-blender: I don't see a new commit but I'll an eye out ;)

a-blender (changes requested):

a-blender (approved):

rancher2/structure_cluster_v2.go
felipe-colussi: I hadn't Pushed it yet. 
https://github.com/rancher/terraform-provider-rancher2/pull/1224/commits/fe71d96fffff23a6af9905367821df16de38f4d1

rancher2/0_provider_upgrade_test.go
```diff
@@ -435,15 +435,14 @@ func testAccRancher2UpgradeVars() resource.TestCheckFunc {
 			rancherTokenKey := os.Getenv("RANCHER_TOKEN_KEY")
 			rancherAdminPass := os.Getenv("RANCHER_ADMIN_PASS")
 			if err := os.Setenv("RANCHER_TOKEN_KEY", token); err != nil {
-				fmt.Printf("Failed to update RANCHER_TOKEN_KEY on ressource %s with err %s", k, err.Error())
+				fmt.Printf("Failed to update RANCHER_TOKEN_KEY on resource %s with err: %s", k, err.Error())
```
a-blender, line 434: Based on our discussion I'd leave these os.Env and print statement in but comment in the code that they fix a race condition in Drone. Otherwise, you'd have to run Drone 8-10 times to get a passing build. Add what you fixed into the PR description as well and reference it here https://github.com/rancher/rancher/issues/42212. Thank you!

a-blender (changes requested):

a-blender:
@felipe-colussi Can you also please squash the top 8-10 commits https://github.com/rancher/terraform-provider-rancher2/pull/1224/commits? They are all pretty small updates.

jakefhyde (changes requested):

rancher2/data_source_rancher2_cluster.go
felipe-colussi: SDK-V2 dosen't allow "Configurable attributes" for computed-only fields. 

On that case "rke_config" have Computed: true without having optional so the SDK-V2 Complains about the MaxItems config.  Is the same for the "ConflictsWith". 

    data source rancher2_notifier: webhook_config: ConflictsWith is for configurable attributes,there's nothing to configure on computed-only field

rancher2/data_source_rancher2_project.go
felipe-colussi: I didn't Removed the Exists, just the Todo.

rancher2/0_provider_upgrade_test.go
felipe-colussi: Yes, that is exactly what it is doing. 

I added some comments to that function, mainly I was having a flaky test (more than 90% of the cases) where I was getting an unauthorized and rancher was letting some dangling resources. 

I added those prints to understand what was happening.   After adding that print I didn't got that error anymore. So I left them. 

Do you think that it is Ok to keep the token get / print ?

jakefhyde: As it stands I don't feel confident merging this if the issue is no longer there because of some print statements. If there is a race condition, we need to fix it.

rancher2/structure_agent_deployment_customization_v2.go
felipe-colussi: 2.  I Created an RKE2 cluster with the following cluster_agent_deployment_customization : 

```
  cluster_agent_deployment_customization {
  override_affinity = <<EOF
{
  "nodeAffinity": {
    "requiredDuringSchedulingIgnoredDuringExecution": {
      "nodeSelectorTerms": [{
        "matchExpressions": [{
          "key": "beta.kubernetes.io/os",
          "operator": "NotIn",
          "values": [
            "windows"
          ]
        }]
      }]
    }
  }
}
EOF
    override_resource_requirements {
      cpu_limit      = "800m"
      cpu_request    = "250m"
      memory_limit   = "800Mi"
      memory_request = "250Mi"
    }
  }

```

I wasn't able to make it to be ready (nor I was able to do that using UI). 

But when I checked the management.cattle.io/v3  - Cluster it had: 
 
 
 ```
 spec:
  agentImageOverride: ''
  answers: {}
  clusterAgentDeploymentCustomization:
    overrideAffinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: beta.kubernetes.io/os
                  operator: NotIn
                  values:
                    - windows
    overrideResourceRequirements:
      limits:
        cpu: 800m
        memory: 800Mi
      requests:
        cpu: 250m
        memory: 250Mi

 
 ```

rancher2/0_provider_upgrade_test.go
felipe-colussi: I'll take a deeper look.

I imagined that could be a race cause the print statements fixed it. 

Those errors just happens on tests and that function is a test one.

rancher2/structure_agent_deployment_customization_v2.go
a-blender: @felipe-colussi Could you post those results on the PR description? Thanks!

rancher2/0_provider_upgrade_test.go
felipe-colussi: Thanks for asking this. 

At the end everything that I tried  wasn't 100% Then I found out reading the Struct that is passed to tests that they don't support "Aliases" on the test framework and that was causing the flaky test.,

rancher2/data_source_rancher2_cluster.go
a-blender: Yeah, Computed fields usually refer to non-user configurable fields and are determined by the provider on plan/apply execution https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors. So it wouldn't make sense for a Computed field to have a MaxItems defined because that's not configurable anyway. Sounds like v2 is just enforcing it now.

Gotta love trying to get users to be less stupid ;)

a-blender:
Looks like the build is failing due to go fmt. Run gofmt -w on these files

```
./rancher2/structure_cluster_v2_test.go
./rancher2/resource_rancher2_global_role.go
```

a-blender:
The changes got massively confusing on this PR so we did the following
- `git rebase master` and resolved the merge conflicts
- verified changes by the upgrade to sdk v2 were still the same, checked build and tests are working and squashed 30 commits into 1 with a readable message.

@felipe-colussi If there are additional review comments, please include them in only 1-2 additional commits that are for addressed comments.

Issues with the auto-generated `/vendor` dir turned out to be unique to my local environment. Unused dependency fix will be put into a separate PR if needed.

felipe-colussi:
Smoke tests after migration: 
Rke2 - AWS = Create Auth ->  Cluster creation -> scale down -> scale Up -> Delete.  OK.

RKE2 - DO = Create Auth ->  Create Cluster -> Upgrade K8S version -> Delete.  Ok.

RKE - DO =  "Data" Auth  -> Create Cluster with cluster_agent_deployment_customization -> Remove override_affinity+override_resource_requirements+fleet_agent_deployment_customization -> Increase node count -> Delete.  Ok
[rke-do-customization.txt](https://github.com/rancher/terraform-provider-rancher2/files/13057252/rke-do-customization.txt)

RKE - AWS -> Create Auth -> Create Cluster (k8s version and network plugin only) ->  increase node count -> dete. Ok.

a-blender (approved):

rancher2/0_provider_upgrade_test.go
```diff
@@ -53,15 +53,14 @@ resource "rancher2_namespace" "testacc" {
 	testAccCheckRancher2UpgradeCatalogV24 = testAccRancher2CatalogGlobal + testAccRancher2CatalogCluster + testAccRancher2CatalogProject
 	testAccCheckRancher2UpgradeCertificateV24 = testAccRancher2Certificate + testAccRancher2CertificateNs
 	testAccCheckRancher2BootstrapV23 = `
-provider "rancher2" {
```
jakefhyde, line 56: For my own understanding can you explain why this was necessary

jakefhyde (approved):

rancher2/0_provider_upgrade_test.go
felipe-colussi: This is the fix for the flaky test that was happening. 

I didn't found anything related to it on the migration guides, but once I started reading the code docs the new test framework explicitly says that we can't use alias on providers and recommend to use multiple providers instead. 

 https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource@v2.28.0#TestCase.ProviderFactories

matttrach:
Closing this in favor of a different approach: https://github.com/rancher/terraform-provider-rancher2/issues/1696

Issue #1222, opened 2023-09-06
[RFE] Graceful Shutdown of vSphere VM


jiaqiluo:
Terraform side of https://github.com/rancher/rancher/issues/42308

jiaqiluo:
can be validated in https://github.com/rancher/terraform-provider-rancher2/releases/tag/v4.0.0-rc3

slickwarren:
tested the following on rancher v2.8-head (6376016):
* rke2 fresh cluster, default settings -> delete a node in the cluster -> non-graceful shutdown
* rke2 fresh cluster, default settings -> delete the cluster -> non-graceful shutdown
* rke1 fresh cluster, default settings -> delete a node in the cluster -> non-graceful shutdown
* rke1 fresh cluster, default settings -> delete the cluster -> non-graceful shutdown
* rke2 fresh cluster, 0 for graceful shutdown in settings -> delete a node in the cluster -> non-graceful shutdown
* rke2 fresh cluster, 0 for graceful shutdown in settings -> delete the cluster -> non-graceful shutdown
* rke1 fresh cluster,  0 for graceful shutdown in settings -> delete a node in the cluster -> non-graceful shutdown
* rke1 fresh cluster, 0 for graceful shutdown in settings -> delete the cluster -> non-graceful shutdown
* rke2 fresh cluster, 10 for graceful shutdown in settings -> delete a node in the cluster -> attempt at graceful shutdown, then non-graceful shutdown after 10 seconds
* rke2 fresh cluster, 10 for graceful shutdown in settings -> delete the cluster -> attempt at graceful shutdown, then non-graceful shutdown after 10 seconds
* rke1 fresh cluster, 10 for graceful shutdown in settings -> delete a node in the cluster -> attempt at graceful shutdown, then non-graceful shutdown after 10 seconds
* rke1 fresh cluster, 10 for graceful shutdown in settings -> delete the cluster -> attempt at graceful shutdown, then non-graceful shutdown after 10 seconds

* rke2 fresh cluster, 100 for graceful shutdown in settings -> delete a node in the cluster -> successful graceful shutdown
* rke2 fresh cluster, 100 for graceful shutdown in settings -> delete the cluster -> successful graceful shutdown
* rke1 fresh cluster, 100 for graceful shutdown in settings -> delete a node in the cluster -> successful graceful shutdown
* rke1 fresh cluster, 100 for graceful shutdown in settings -> delete the cluster -> successful graceful shutdown


[...]

PR #1228, opened 2023-09-19
add the support for graceful_shutdown_timeout for vSphere nodes

jiaqiluo:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1222

 
## Problem

We would like to have support for graceful_shutdown_timeout for vSphere nodes in the TF provider. 

## Solution

 
Add the support for graceful_shutdown_timeout for vSphere nodes

### Manual Testing

This PR has been validated working in the internal vSphere envenomation, including RKE1/RKE2/K3s node-driver vsphere cluster 

If the field `graceful_shutdown_timeout` is set in the vsphere_config, when we delete the cluster (`terraform destory`), the task "initial guest os shutdown" is triggered in vSphere Center Client, which does not happen if we do not set the flag.

### Automated Testing

Exiting tests are expanded to cover the new field, and that is all we can do for now. 

## QA Testing Considerations

Provisioning and upgrading of the RKE1/RKE2/K3s node-driver vsphere cluster  

 Provisioning and upgrading of the RKE1/RKE2/K3s node-driver vsphere cluster

### DIFF
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -312,6 +312,7 @@ The following attributes are exported:
 * `datastore_cluster` - (Optional) vSphere datastore cluster for virtual machine (string)
 * `disk_size` - (Optional) vSphere size of disk for docker VM (in MB). Default `20480` (string)
 * `folder` - (Optional) vSphere folder for the docker VM. This folder must already exist in the datacenter (string)
+* `graceful_shutdown_timeout` (Optional) Duration in seconds for wait for graceful shutdown of the VM. A force destroy will be performed when the value is zero (string)
 * `hostsystem` - (Optional) vSphere compute resource where the docker VM will be instantiated. This can be omitted if using a cluster with DRS (string)
 * `memory_size` - (Optional) vSphere size of memory for docker VM (in MB). Default `2048` (string)
 * `network` - (Optional) vSphere network where the docker VM will be attached (list)
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -312,7 +312,7 @@ The following attributes are exported:
 * `datastore_cluster` - (Optional) vSphere datastore cluster for virtual machine (string)
 * `disk_size` - (Optional) vSphere size of disk for docker VM (in MB). Default `20480` (string)
 * `folder` - (Optional) vSphere folder for the docker VM. This folder must already exist in the datacenter (string)
-* `graceful_shutdown_timeout` (Optional) Duration in seconds for wait for graceful shutdown of the VM. A force destroy will be performed when the value is zero (string)
+* `graceful_shutdown_timeout` (Optional) Duration in seconds before the graceful shutdown of the VM times out and the VM is destroyed. A force destroy will be performed when the value is zero (string)
 * `hostsystem` - (Optional) vSphere compute resource where the docker VM will be instantiated. This can be omitted if using a cluster with DRS (string)
 * `memory_size` - (Optional) vSphere size of memory for docker VM (in MB). Default `2048` (string)
 * `network` - (Optional) vSphere network where the docker VM will be attached (list)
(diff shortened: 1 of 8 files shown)


felipe-colussi:
Shouldn't u add a line under `vsphere_config` docs?

jiaqiluo:
> Shouldn't u add a line under `vsphere_config` docs?

@felipe-colussi, Good catch. PR is updated.

docs/resources/machine_config_v2.md
```diff
@@ -312,6 +312,7 @@ The following attributes are exported:
 * `datastore_cluster` - (Optional) vSphere datastore cluster for virtual machine (string)
 * `disk_size` - (Optional) vSphere size of disk for docker VM (in MB). Default `20480` (string)
 * `folder` - (Optional) vSphere folder for the docker VM. This folder must already exist in the datacenter (string)
+* `graceful_shutdown_timeout` (Optional) Duration in seconds for wait for graceful shutdown of the VM. A force destroy will be performed when the value is zero (string)
```
HarrisonWAffel: small nit, but maybe the wording `(Optional) Duration in seconds before the graceful shutdown of the VM times out and the VM is destroyed. A force destroy will be performed when the value is zero (string)` would sound better? This would also apply to other files which use this wording.

rancher2/resource_rancher2_node_template_test.go
```diff
@@ -267,6 +267,7 @@ resource "` + testAccRancher2NodeTemplateType + `" "foo-vsphere" {
 	cpu_count = "8"
 	disk_size = "20480"
 	pool =  "pool-YYYYYYYY"
+	graceful_shutdown_timeout = "30"
```
a-blender, line 270: Nit: set the shutdown timeout as a `const` and use that variable here and on line 746

rancher2/schema_cluster_rke_config_cloud_provider_vsphere.go
```diff
@@ -51,6 +51,11 @@ func clusterRKEConfigCloudProviderVsphereGlobalFields() map[string]*schema.Schem
 			Optional: true,
 			Computed: true,
 		},
+		"graceful_shutdown_timeout": {
```
a-blender, line 54: Add a `Description` to this field and reference it in the `rancher2_cluster` docs

rancher2/schema_machine_config_v2_vsphere.go
```diff
@@ -206,6 +206,11 @@ func machineConfigV2VmwarevsphereFields() map[string]*schema.Schema {
 			Default:     "443",
 			Description: "vSphere Port for vCenter",
 		},
+		"graceful_shutdown_timeout": {
```
a-blender, line 209: Is this field not also Computed? Does only the Global one need to be?

a-blender (dismissed):

rancher2/schema_machine_config_v2_vsphere.go
jiaqiluo: No, it is not. It must be set explicitly. 
Good catch! I am going to drop the 'computed: true' from the global one.

rancher2/resource_rancher2_node_template_test.go
jiaqiluo: I think I will not move it to const to match other values, like the disk_size, cpu_count.

rancher2/schema_cluster_rke_config_cloud_provider_vsphere.go
jiaqiluo: No entry in this list has a description, so I guess it is the same as this one. 
The description is added in another place; docs are updated too.

docs/resources/machine_config_v2.md
jiaqiluo: fixed

HarrisonWAffel (approved):

thatmidwesterncoder (approved):

Issue #107, opened 2019-08-27
When a node pool is created and quality set to 0 still it create a node and following run deletes it


bensonarokiadoss:
When a node pool is created and quantity set to 0 still it create a node

Step 1: During first terraform execution a node pool is created with 1 node
Step 2: During second terraform execution the node created before will be deleted

rawmind0:
The minimum and the default quantity for node-pool is 1, same on rancher UI. What would be the sense to create a node pool with quantity set to 0??

bensonarokiadoss:
My use case we dont use state files and in certain scenarios we might need to create the node template and leave it for the AZ. So we are fine with how it works as we would create a node pool using the template if we need to set up nodes on the AZ. This is more of an observation when i set to  0 it makes sense it create the node with 1 as that how it works in the UI. But on the second run it goes and delete those since set to 0. 

Reason for the raising as expecting either one of the below

1. Fails/Error out for 0
2. Create node pool but leave the node count to 0 (Preferred - Since it give my customer the ability to add nodes if needed on those pools)
3. Create a min of 1 node if set to 0 but dont delete on the second run

Feel free to close out if you feel not appropriate

rawmind0:
It's completely approppiate, thanks for reporting it.

We'll address it following one of your suggestions.

rawmind0:
Submitted PR #121 to address it.

rawmind0:
Merged PR #121 to address issue. Please, reopen it if needed.

PR #1232, opened 2023-09-24
Allow machine pools to be created and scaled down to 0

klippo:
## Problem

It's not possible to set a machine_pool quantity to 0, this blocks temporarily scaling down a pool, and instead it will be deleted. It's possible to both create and scale down clusters from the UI with a value of 0, so the provider should allow the same.
 
## Solution
Remove the [ValidateFunc: validation.IntAtLeast(1),](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_cluster_v2_rke_config_machine_pool.go#L99)
 
## Testing
N/A

### Manual Testing
Verified functionality by deploying to a RKE2 clusters in OpenStack with machine_pool quantity set to 0.

### Automated Testing
N/A

## QA Testing Considerations
N/A
 
I've manually tested this change on OpenStack and it seems to work as expected. But I'm not sure if it will work with other providers.

This was first discussed in https://github.com/rancher/terraform-provider-rancher2/issues/107, and while it's not possible to create a cluster with only empty pools, it's still possible create empty pools as long as there's at least one pool with quantity >= 1.

### DIFF
--- a/rancher2/schema_cluster_v2_rke_config_machine_pool.go
+++ b/rancher2/schema_cluster_v2_rke_config_machine_pool.go
@@ -93,11 +93,10 @@ func clusterV2RKEConfigMachinePoolFields() map[string]*schema.Schema {
 			Description: "Machine pool paused",
 		},
 		"quantity": {
-			Type:         schema.TypeInt,
-			Optional:     true,
-			Default:      1,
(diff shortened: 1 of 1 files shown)


a-blender (approved):
Lgtm, thank you!

mjura (approved):
LGTM

PR #1234, opened 2023-09-25
Testing - Changes to terraform

felipe-colussi:
I want to do some test on the build fails after v2. 

Creating a new draft to avoid doing more commits on the main PR.


felipe-colussi:
Got what I neee

PR #1235, opened 2023-09-25
Revert "Remove OpenStack keypair_name references"

thatmidwesterncoder:
This reverts commit 4096070b7a7618e2f05cbf844757fab0ce8e02f9.

## Problem

the keypair_name field for openstack was necessary even if there weren't any explicit code references to it in rancher.
 
## Solution

reverting commit 4096070b7a7618e2f05cbf844757fab0ce8e02f9 will fix the issue.

PR #1236, opened 2023-09-25
[Backport release/v3] Revert "Remove OpenStack keypair_name references"

thatmidwesterncoder:
This reverts commit 4096070b7a7618e2f05cbf844757fab0ce8e02f9.

## Problem

the keypair_name field for openstack was necessary even if there weren't any explicit code references to it in rancher.
 
## Solution

reverting commit 4096070b7a7618e2f05cbf844757fab0ce8e02f9 will fix the issue. 

---

This is the same as rancher/terraform-provider-rancher2#1235 - just backporting to release/v3


a-blender (approved):
Ah, darn it. We couldn't have tested it on all hosted platforms - glad an OpenStack customer caught this. Can we add a comment to the provider on why we are leaving this in?

PR #1242, opened 2023-10-04
Add inherited_cluster_roles attribute to global_role

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/42967
 
## Problem

`inheritedClusterRoles` is a new field in global roles. We should be able to set it using rancher terraform provider when creating/updating a global role resource.
 
## Solution

A new attribute `inherited_cluster_roles` was added to the global role.
 
## Testing

Check that `inherited_cluster_roles` attribute is correctly populated when a global role is created/updated.

### Manual Testing

Used a local build of the rancher terraform provider against a local k8s cluster v1.27.4+k3s1 running rancher v2.8 HEAD (c7af81ec9c55b) to create and update a global role with inherited cluster roles.

```hcl
resource "rancher2_global_role" "test-global-role" {
    name = "test-global-role"
    description = "test-global-role"
    new_user_default = false
    rules {
      api_groups = ["*"]
      resources = ["secrets"]
      verbs = ["create"]
    }
    inherited_cluster_roles = ["cluster-owner"]
}
```

```sh
terraform plan -var-file rancher.tfvars -auto-approve
...

  # rancher2_global_role.test-global-role will be created
  + resource "rancher2_global_role" "test-global-role" {
      + annotations             = (known after apply)
      + builtin                 = (known after apply)
      + description             = "test-global-role"
      + id                      = (known after apply)
      + inherited_cluster_roles = [
          + "cluster-owner",
        ]
      + labels                  = (known after apply)
      + name                    = "test-global-role"
      + new_user_default        = false

      + rules {
          + api_groups = [
              + "*",
            ]
          + resources  = [
              + "secrets",
            ]
          + verbs      = [
              + "create",
            ]
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
rancher2_global_role.test-global-role: Creating...
rancher2_global_role.test-global-role: Creation complete after 4s [id=gr-xt2jb]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```
```sh
k get globalrole gr-xt2jb -o yaml
apiVersion: management.cattle.io/v3
description: test-global-role
displayName: test-global-role
inheritedClusterRoles:
- cluster-owner
kind: GlobalRole
...
```

### Automated Testing

Unit tests updated to cover the new attribute.


go.mod
```diff
@@ -178,8 +180,8 @@ replace (
 	github.com/crewjam/saml => github.com/crewjam/saml v0.4.5
 	github.com/docker/distribution => github.com/docker/distribution v2.8.1+incompatible
 	github.com/docker/docker => github.com/docker/docker v20.10.17+incompatible
-	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20230512212658-eaa7b90049a4
-	github.com/rancher/rancher/pkg/client => github.com/rancher/rancher/pkg/client v0.0.0-20230512212658-eaa7b90049a4
+	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20230901132600-5e1ee2611616
+	github.com/rancher/rancher/pkg/client => github.com/rancher/rancher/pkg/client v0.0.0-20230901132600-5e1ee2611616
```
pmatseykanets, line 184: It's not clear what version of rancher/rancher to reference yet.

rancher2/resource_rancher2_global_role.go
```diff
@@ -94,13 +94,16 @@ func resourceRancher2GlobalRoleUpdate(d *schema.ResourceData, meta interface{})
 			return resource.NonRetryableError(err)
 		}
 
+		log.Println("updated inherited_cluster_roles", toArrayString(d.Get("inherited_cluster_roles").([]interface{})))
```
MbolotSuse: I think that we can remove this now. None of the other fields have a log statement, so I don't think we should have a special case for this one.

rancher2/schema_global_role.go
```diff
@@ -39,6 +39,15 @@ func globalRoleFields() map[string]*schema.Schema {
 				Schema: policyRuleFields(),
 			},
 		},
+		"inherited_cluster_roles": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Computed:    true,
```
MbolotSuse: Based on the [definition of computed](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed) I don't think this should be true, since users can set this as part of the input and the definition states:
> Computed is often used to represent values that are not user configurable or can not be known at time of terraform plan or apply, such as date of creation or a service specific UUID.

MbolotSuse (changes requested):
Couple of small issues, and I think we also need to update the `docs/resources/role_template.md` file as well.

rancher2/resource_rancher2_global_role.go
pmatseykanets: Thanks for catching this. Absolutely, it was used to troubleshoot the update issue and wasn't meant to be committed.

rancher2/schema_global_role.go
pmatseykanets: That is a great point! Shouldn't the same logic apply to majority of other fields though? 
```
* `description` - (Optional/Computed) Role template description (string)
* `role_template_ids` - (Optional/Computed) Inherit role template IDs (list)
* `rules` - (Optional/Computed) Role template policy rules (list)
* `annotations` - (Optional/Computed) Annotations for role template object (map)
* `labels` - (Optional/Computed) Labels for role template object (map)
```

MbolotSuse: I'd agree that this should apply to most of the other fields. However, we can't copy a bad practice just because another area uses it. I would recommend looking into the history of those other fields - if you can find a justification for using the value as true I'll consider it, but absent that I don't think it makes sense to continue the pattern here.

pmatseykanets: I agree there is no point in spreading it further. FYI I couldn't find any justifications in commits so far.

go.mod
pmatseykanets: Right now it's pointing to a merge commit [5e1ee2611616](https://github.com/rancher/rancher/commit/5e1ee2611616) where `inheritedClusterRoles` was introduced.

pmatseykanets: @a-blender Is this acceptable?

crobby (commented):
Aside from the outstanding question on versioning for the rancher bits, I think this looks good.

eliyamlevy (approved):
LGTM other than rancher version bits as well. Not sure how we handle versioning in this repo.

go.mod
a-blender: If TF rancher2 requires `inheritedClusterRoles` from the rancher backend for this feature to work correctly, than yes but in general the provider should reference the latest commit on rancher release/v2.8. Why are we updating those other deps? Are they indirect?

thatmidwesterncoder (approved):
echoing LGTM but not sure about the version bumps.

go.mod
pmatseykanets: Yes, this requires `inheritedClusterRoles` from the rancher backend, so the [5e1ee2611616](https://github.com/rancher/rancher/commit/5e1ee2611616) is the minimum required version for this feature to work.

> in general the provider should reference the latest commit on rancher release/v2.8

Ideally IMHO, once the release is tagged the provider dependency should be updated to point to the tagged version. 

> Why are we updating those other deps? Are they indirect?

We don't, the Go toolchain does when we point `replace`s to the above-mentioned commit and run `go mod tidy`.

crobby (approved):

crobby (approved):

MbolotSuse (approved):

go.mod
a-blender: On second thought, I think that is the safest option. Lgtm

a-blender (approved):
@pmatseykanets Rancher version bump lgtm, on second thought I think it's the safest option to have TF reference the rancher/rancher commit where the backend for `InheritedGlobalRoles` was merged in. Can't be 100% sure the provider supports the later commits yet. Good to merge

PR #1244, opened 2023-10-05
Set protect-kernel-defaults on v2 clusters

a-blender:
## Problem

 
When going to Cluster Management and clicking Edit Config on a cluster, the page crashes and displays a "Loading ..." text but ends up with the following message: Cannot read properties of undefined (reading 'length'). This happens if a MachineSelectorConfig config is set as null.

## Solution

Rancher sets [protect-kernel-defaults](https://ranchermanager.docs.rancher.com/pages-for-subheaders/rke2-hardening-guide#:~:text=Ensure%20protect%2Dkernel%2Ddefaults%20is%20set%E2%80%8B&text=This%20is%20a%20kubelet%20flag,the%20cluster%20configuration%20in%20Rancher.) false by default under the hood for rke2/k3s clusters but it is not exposed in the UI. My solution is for the TF rancher2 provider to set this field for any empty MachineSelectorConfig config so tf doesn't remove it every time to reconcile the state.
 
## Testing

Confirmed removing `protect-kernel-defaults: false` and setting `config: null` causes the page to crash. Adding the field back in fixes it. Prov an rke2 cluster on EC2 with the fix resolves the bug. Adding the same MachineSelectorConfig via the UI that @Josh-Diamond did during our debug session and then running a `terraform refresh` to refresh the state causes the following to be set,

```
machineSelectorConfig:
      - config:
          protect-kernel-defaults: true
      - config:
          kubelet-arg:
            - cloud-provider=external
        machineLabelSelector:
          matchExpressions: []
          matchLabels:
            key: value
```

which is correct.

### DIFF
--- a/rancher2/schema_cluster_v2_rke_config_system_config.go
+++ b/rancher2/schema_cluster_v2_rke_config_system_config.go
@@ -68,6 +68,7 @@ func clusterV2RKEConfigSystemConfigFieldsV0() map[string]*schema.Schema {
 		"config": {
 			Type:        schema.TypeMap,
 			Optional:    true,
+			Default:     "protect-kernel-defaults: false",
 			Description: "Machine selector config",
 		},
 	}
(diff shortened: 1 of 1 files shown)


thatmidwesterncoder (approved):

a-blender:
DNM waiting on QA testing before merging this forward port to 2.8

a-blender:
@HarrisonWAffel Tested that on upgrade from a previous tf version to this one, a hardened RKE2 cluster with existing `protect-kernel-defaults: true` is not overwritten.

HarrisonWAffel (approved):

PR #1245, opened 2023-10-05
[Backport release/v3] Set protect-kernel-defaults on v2 clusters

a-blender:

### DIFF
--- a/rancher2/schema_cluster_v2_rke_config_system_config.go
+++ b/rancher2/schema_cluster_v2_rke_config_system_config.go
@@ -68,6 +68,7 @@ func clusterV2RKEConfigSystemConfigFieldsV0() map[string]*schema.Schema {
 		"config": {
 			Type:        schema.TypeMap,
 			Optional:    true,
+			Default:     "protect-kernel-defaults: false",
 			Description: "Machine selector config",
 		},
 	}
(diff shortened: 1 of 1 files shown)


HarrisonWAffel (commented):
This makes sense as to why it would fix the UI, the only thing I am worried about is that `protect-kernel-defaults` is deprecated, https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/, if it's completely removed / no longer supported by rancher we would need a new fix for this down the line.

This may be a dumb question, but do we know if this will effect existing clusters which have this set to true? I would think the sync'ing would take care of that, but it would be good to double check since overwriting that value would 'break' hardened RKE2 cluster (CIS Scans would fail)

a-blender:
@HarrisonWAffel Mmm true - but such is the journey of having TF parity with Rancher. I'm not overwriting the value, I'm setting the default if config is empty. So if a user has `protect-kernel-defaults` set to true for an RKE2 hardened cluster, that value will be maintained.

HarrisonWAffel (approved):
talked on slack, we will want to make sure that we return to this issue when/if this field is removed from the v2 machine selector config.

a-blender:
@jakefhyde Can you review? To get your take, since I'm adapting to a v2 prov CLI arg.

a-blender:
Sounds good! cc @jakefhyde to glance at this when you get back next week since this fix is adapting to a v2 prov argument. Ping me with any concerns.

thatmidwesterncoder (approved):

HarrisonWAffel (approved):

PR #1246, opened 2023-10-05
[DNM see note below] Add PSACT custom resource and tests

a-blender:
## Problem

 
Add PSACT template resource https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates to the provider.

## Solution

What was added

- [x] `pod_security_admission_configuration_template` resource and data source
- [x] `name`, `description`, `defaults`, and `exemption` subfields within the template designed to fit the Rancher backend + ValidateFunc for default modes.
- [x] Update tests
- [x] Update docs and config examples

 
### Manual Testing

Tested creating, modifying, and removing a custom PSACT template on a 1.27 rke2 cluster. All tests successful. Also tested creating a template with bad data and this fails.

_Note: Modifying / removing PSACT exemptions (usernames, runtime classes, or namespaces) requires the template to be reapplied to the kube API server for this change to take effect. If a cluster is already using a template, you must remove the reference, update the template, then re-add the template to the cluster._

_Note: Rancher webhook manages the machine selector file sources created for a reference to a PSACT template. A user has to add a `ignore_changes` to their tf config or tf will try to add a file source back to the provisioning spec on template deletion._


a-blender:
All tests are working here except for cluster deletion. If you try to remove all infra (cluster + custom template) at the same time, template deletion will error out because the cluster (and reference to the template) is not removed yet. A retry needs to be added to the PodSecurityAdmissionConfigurationTemplate 'Delete' func to retry deleting the template a few times instead of returning a static err.

kkaempf:
closing stale PRs.

Issue #927, opened 2022-05-16
rancher2_cluster eks_config_v2 is unable to launch nodes with the launch teamplate.


cem-altuner:
`rancher/rancher2 version : 1.23.0`

Hi, I am trying to add a Windows node group using the launch template. The link below shows how to use the `rancher2_cluster` resource using launch templates.

https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster#creating-eks-cluster-from-rancher-v2-using-eks_config_v2-and-launch-template-for-rancher-v256-or-above

However, I am receiving the following error.

`InvalidParameterException: You cannot specify an AMI Type other than CUSTOM, when specifying an image id in your launch template. { RespMetadata: { StatusCode: 400, RequestID: "362d0ec0-d42d-48a7-8d67-8835847b0f51" }, ClusterName: "cluster", Message_: "You cannot specify an AMI Type other than CUSTOM, when specifying an image id in your launch template.", NodegroupName: "windows" }`

It doesn't work even if you give template name or template id. You have to manually select the launch template name via console.

matttrach:
This issue is a couple of years old now and can only reference unsupported versions, so I am going to close it, feel free to open a new issue if this is still an problem.

PR #1247, opened 2023-10-09
Add note linking to AWS docs

mallardduck:
## Problem
Some customers come into Infrastructure as Code because they are Rancher users, rather than because they are embracing IaC out right. As such, some users don't realize they should user other Terraform providers than just Rancher. This gives the impression that if the Rancher TFP doesn't do a thing then it's a missing feature or bug.
 
## Solution
We can provide users with context clues of when to user other terraform providers as necessary. In this case we are linking to the relevant AWS terraform data-sources. By telling users to fetch image IDs this way we are implicitly telling them to use other Terraform providers.
 
## Testing
N/A; docs change.

### Manual Testing
N/A; docs change.

### Automated Testing
N/A; docs change.

## QA Testing Considerations
N/A; docs change.
 
N/A; docs change.

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -424,7 +424,7 @@ resource "rancher2_cluster" "foo" {
 
 ### Creating EKS cluster from Rancher v2, using `eks_config_v2` and launch template. For Rancher v2.5.6 and above.
 
-Note: To use `launch_template` you must provide the ID (seen as `<EC2_LAUNCH_TEMPLATE_ID>`) to the template either as a static value. Or fetched via AWS data-source using one of: [aws_ami](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami.html), [aws_ami_ids](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami_ids), or similar data-sources.
+Note: To use `launch_template` you must provide the ID (seen as `<EC2_LAUNCH_TEMPLATE_ID>`) to the template either as a static value. Or fetched via AWS data-source using one of: [aws_ami](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami.html), [aws_ami_ids](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami_ids), or similar data-sources. You can also create a custom [`launch_template`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template) first and provide the ID to that.
 
 ```hcl
 resource "rancher2_cloud_credential" "foo" {
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -424,6 +424,8 @@ resource "rancher2_cluster" "foo" {
 
 ### Creating EKS cluster from Rancher v2, using `eks_config_v2` and launch template. For Rancher v2.5.6 and above.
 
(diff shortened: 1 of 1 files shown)


bennysp:
Hi - I have helped work through this with SUSE vendor. I don't think this is all there is to this one unfortunately.

https://github.com/rancher/terraform-provider-rancher2/issues/927

The above issue/error is what we see when simply going from no launch template to a launch template in Rancher provider; It breaks because we are not providing an AMI, but that is not even enough to get it to work. AMI ID is not enough on it's own. You also need to then bootstrap it in order to get it to work too.

You have to do something like this in order to get it to work with the Rancher provided setup and a launch_template....

```yaml
# get data for aws_ssm_parameters_by_path for var.kubernetes_version
# https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html
data "aws_ssm_parameter" "ami_imageid" {
  name = "/aws/service/eks/optimized-ami/${var.kubernetes_version}/amazon-linux-2/recommended/image_id"
}

# bootstrap the EKS nodes since we have to provide an image id as a workaround
# https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html
data "template_file" "ng_user_data" {
  template = <<-EOF
              #!/bin/bash
              set -o xtrace
              /etc/eks/bootstrap.sh ${var.eks_ds_name}
              EOF
}

resource "aws_launch_template" "rancher_ds_worker" {
  for_each = zipmap(var.node_groups[*].name, var.node_groups)
 
  name_prefix = "${var.eks_ds_name}-${each.value.name}-launchtemp"

  # 
  # When specifying launch template, do NOT specify these in here and keep in node group; limitation of AWS
  # https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html
  # DO NOT USE HERE: iam_instance_profile
  # 

  instance_type = each.value.node_instance_type
  image_id = data.aws_ssm_parameter.ami_imageid.value  # must provide a workaround to avoid a Rancher bug

  block_device_mappings {
    device_name = "/dev/xvda"
    ebs {
      volume_size = each.value.disk_size
      volume_type = "gp3"
      iops                  = 3000
      throughput            = 125
      encrypted             = true
      delete_on_termination = true
    }
  }
  update_default_version = true
  key_name = var.ec2_ssh_key
  vpc_security_group_ids = [var.eks_node_group_sg_id]
  
  user_data = base64encode(data.template_file.ng_user_data.rendered)
}

resource "rancher2_cluster" "rancher-cluster" {
  name = var.eks_ds_name
  description = "Terraform EKS downstream cluster"

  eks_config_v2 {
    cloud_credential_id = var.cloud_credential
    region = var.aws_region
    kubernetes_version = var.kubernetes_version

    subnets = var.subnet_ids
    logging_types = var.eks_logging_types

    security_groups = [var.eks_cluster_sg_id]

    dynamic "node_groups" {
      for_each = var.node_groups
      content {
        name            = "${var.eks_ds_name}-${node_groups.value["name"]}"
        subnets         = node_groups.value["subnet_ids"]
        node_role       = var.ng_iam_role

        # 
        # When specifying launch template, do NOT specify these in node groups; limitation of AWS
        # https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html
        # DO NOT USE HERE: ec2_ssh_key, instance_type, disk_size, image_id, user_data
        # 
        
        launch_template {
          id = resource.aws_launch_template.rancher_ds_worker[node_groups.value["name"]].id
          version = resource.aws_launch_template.rancher_ds_worker[node_groups.value["name"]].latest_version
          name = resource.aws_launch_template.rancher_ds_worker[node_groups.value["name"]].name
        }

        desired_size  = node_groups.value["des_dataplane_nodes"]
        max_size      = node_groups.value["max_dataplane_nodes"]
        min_size      = node_groups.value["min_dataplane_nodes"]
    }
  }
`

richardcase (dismissed):

matttrach (approved):
lgtm

mjura (approved):

mjura:
looks good, I have restarted CI

PR #1256, opened 2023-10-25
Use dl.k8s.io for getting kubectl

superseb:
## Problem
Preparing for GCS bucket deprecation at some point (https://github.com/kubernetes/k8s.io/issues/2396)
 
## Solution
Change to the currently documented source
 
## Testing
Using [start_testacc.sh](https://github.com/rancher/terraform-provider-rancher2/blob/master/scripts/start_testacc.sh)

### DIFF
--- a/scripts/start_testacc.sh
+++ b/scripts/start_testacc.sh
@@ -42,7 +42,7 @@ if [ "${JQ_BIN}" == "none" ] ; then
 fi
 ## kubectl
 KUBECTL_NAME=kubectl
-KUBECTL_URL="https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+KUBECTL_URL="https://dl.k8s.io/release/$(curl -sL https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
 KUBECTL_BIN=$(which ${KUBECTL_NAME} || echo none)
 if [ "${KUBECTL_BIN}" == "none" ] ; then
   echo Downloading ${KUBECTL_NAME}


a-blender:
@superseb Hey is this ready for review? Is there a linked issue? Any thoughts on the scope of testing this issue... I'm not entirely sure when/how the TF provider uses kubectl. It uses a token to access the Rancher API client and accesses downstream resources that way.

superseb:
According to the script it is being used here https://github.com/rancher/terraform-provider-rancher2/blob/master/scripts/start_testacc.sh#L185.

I had two Drone builds being killed so I left it but will create the corresponding issue but this is the same as the one for `terraform-provider-rke`.

a-blender (approved):
Can you verify `make testacc` still passes? Otherwise lgtm

superseb:
@a-blender There seems to be no reference to this file anywhere, can we just remove it instead?

a-blender:
@superseb Not entirely sure, running `make testacc` runs the infrastructure tests but right now it continuously tries to create a k3s cluster without being fully successful . 

```
+++ /Users/annablendermann/.rd/bin/docker run -i --rm curlimages/curl -sk https://172.17.0.2:6443/ping
+++ echo starting rancher k3s cluster
++ k3s_ready='starting rancher k3s cluster'
++ '[' 'starting rancher k3s cluster' '!=' pong ']'
++ sleep 5
```

I think tf will need `KUBECTL_URL` to download the kubectl binary to access the cluster if it doesn't already exist. I'd probably just update it for now.

Is that what you were referring to?

superseb:
My point is that `scripts/start_testacc.sh` is not referenced anymore in the code. The reference to it was removed a long time ago [here](https://github.com/rancher/terraform-provider-rancher2/commit/7cb2e0a0d14a6c4538703ddb541408ca06fc993b#diff-0e561388a544c440ff3be1d24f72a8d9ec80b8f8d0d1eb4ecb9d0bfb1c1fd8cdL13).

Unless there is an external dependency that uses is, or if there is a case to keep it, I would remove it to clean the codebase from unused scripts.

jiaqiluo (approved):

Issue #1187, opened 2023-08-03
[Q4] rancher2_app_v2 Ignoring Cluster Private Registry configuration and use only system default registry


matteogazzadi:
### Rancher Server Setup
- Rancher version: 2.7.5
- Global default registry configured on corporate private registry Azure ACR.

### Information about the Cluster
- Kubernetes version: v1.25.9-rancher2-1
- Cluster Type Downstream:
* Custom = Running a docker command on a node. Using RKE1

### User Information
- What is the role of the user logged in?  Admin

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? 3.1.0
- What is the version of Terraform in use? 1.0.9

### Describe the bug

Our downstream infrastructure is designed to have a local container registry (Harbor) acting as proxy-cache towards the corporate container registry Azure Container Registry. Both are private registries. 
The idea is to reduce the overall required network bandwidth during commissioning and avoid to contact corporate registry when not needed.
 
The RKE1 downstream cluster is created using `rancher2_cluster`, and `private_registries` section is set as:
```
    private_registries {
      is_default = true
      url = "${var.harbor_dns_name}/${var.harbor_project_name}"
      user = "${harbor_robot_account.harbor_k8s_account.full_name}"
      password = "${random_password.harbor_k8s_account_password.result}"
    }
```

Harbor is managed still via Terraform and we set up a proxy-cache project towards our corporate Container Registry.

Cluster creation does work as expected, all containers are pulled from corporate private registry through local harbor.

Once Cluster is created, we want to install an app from rancher catalog using `rancher2_app_v2`.  The installation starts but the containers are pulled using the system default registry configured on rancher and not from the configured cluster private registry.
Of course this lead to ImagePullBackOff on any pod.

We tried to override the `global.cattle.system-default-registry` of the app but this does not work, See:  rancher/terraform-provider-rancher2#813 

We cannot set the system-default-registry in rancher as the one on downstream (completely no sense in general)

### To Reproduce

1. In Rancher configure a global default registry
2. Create a downstream cluster RKE1 using `rancher2_cluster` and configure a `private_registry` different from the Rancher global default registry
3. Install an app using `rancher2_app_v2`. 

### Actual Result

- App PODs Images are pulled from Rancher Global system default registry and not from the cluster private registry (the local Harbor).

### Expected Result
- App PODs Images are pulled from  the configured cluster private registry (the local Harbor).

### Screenshots
[...]

PR #1265, opened 2023-11-17
Updating default_system_registry behavior at rancher2_app_v2 resource

nicholasSUSE:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1187
 
## Problem

 
The terraform resource `rancher2_app_v2` had no **Argument** for setting up a **Custom** System Registry, limiting clients to install an app only from a previously configured **SystemDefaultRegistry**.

There was in place a **computed attribute** named `default_system_registry`.

## Solution

 
Transform the **computed attribute** `default_system_registry` into an **optional argument** for the resource too. 

- **If the argument is used**: override the installation with its value. 
- **If it is not used**: use the **Global Default System Registry** value which might be an empty string or not.

## Testing

The reproducing steps in the Github Issue are very hard and take a long time to execute. 

In order to test without needing to set all the infrastructure (specially the private registry with all the images):

Testing With Terraform:
1. Apply 1 AWS Instance + Rancher 
2. Apply 1 AWS Instance + Import RKE downstream cluster
3. Use resource `rancher2_app_v2` to create a new app installation passing the recently created `system_default_registry` **argument** pointing to a random string pretending to be the private registry DNS.
4. Check if the Charts begins the installation under Apps.
5. Check if the related system image for running the chart fails because it points out to the random string we passed. 

Results when trying to install CIS Benchmark passing a random string as the value:
image

---

### Manual Testing

Part of the terraform code to execute the tests:
```(hcl)

resource "rancher2_cluster" "bug_cluster" {
    provider = rancher2.BUG

    depends_on = [
        aws_instance.bug_node
    ]

    name = "${var.prefix}-bug-cluster"
    enable_cluster_alerting = false
    enable_cluster_monitoring = false

    rke_config {
      network {
        plugin = "canal"
      }
      enable_cri_dockerd    = true
      kubernetes_version = "v1.25.9-rancher2-1"
      ignore_docker_version = true
    }
}

# Create Rancher-CIS-Benchark App from Helm chart
resource "rancher2_app_v2" "cis_benchmark_bug_cluster" {
  provider = rancher2.BUG

  depends_on = [
    rancher2_cluster_sync.bug_sync
  ]

  cluster_id = rancher2_cluster_sync.bug_sync.cluster_id
  name = "rancher-cis-benchmark"
  namespace = "cis-operator-system"
  repo_name = "rancher-charts"
  chart_name = "rancher-cis-benchmark"
  cleanup_on_fail = true
  wait = false

  # comment/uncomment to reproduce/disable the bug
  # system_default_registry = "YYYYZ"
  timeouts {
    create = "5m"
    update = "5m"
    delete = "5m"
  }
}


```

### DIFF
--- a/docs/resources/app_v2.md
+++ b/docs/resources/app_v2.md
@@ -21,6 +21,23 @@ resource "rancher2_app_v2" "foo" {
 }
 ```
 
+### Create an App from a Helm Chart using a different registry
+
+The `system_default_registry` argument can override the global value at App installation. If argument is not provided, the global value for System Default Registry will be used instead.
+
+```hcl
+resource "rancher2_app_v2" "cis_benchmark" {
+  cluster_id = "<CLUSTER_ID>"
+  name = "rancher-cis-benchmark"
+  namespace = "cis-operator-system"
+  repo_name = "rancher-charts"
+  chart_name = "rancher-cis-benchmark"
+
+  # Valid DNS for Private Registry
+  system_default_registry = "<some.dns.here>:<PORT>"
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
@@ -40,6 +57,7 @@ The following arguments are supported:
 * `wait` - (Optional) Wait until app is deployed. Default: `true` (bool)
 * `annotations` - (Optional/Computed) Annotations for the app v2 (map)
 * `labels` - (Optional/Computed) Labels for the app v2 (map)
+* `system_default_registry` - (Optional/Computed) System default registry providing images for app deployment (string)
 
 ## Attributes Reference
 
--- a/rancher2/resource_rancher2_app_v2.go
+++ b/rancher2/resource_rancher2_app_v2.go
@@ -98,7 +98,7 @@ func resourceRancher2AppV2Read(d *schema.ResourceData, meta interface{}) error {
 			d.Set("cluster_name", cluster.Name)
 		}
 		if systemDefaultRegistry, ok := d.Get("system_default_registry").(string); !ok || len(systemDefaultRegistry) == 0 {
-			systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(appV2DefaultRegistryID)
+			systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(clusterID, appV2DefaultRegistryID)
 			if err != nil {
 				return resource.NonRetryableError(err)
 			}
--- a/rancher2/resource_rancher2_app_v2.go
+++ b/rancher2/resource_rancher2_app_v2.go
@@ -37,6 +37,7 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
 	repoName := d.Get("repo_name").(string)
 	chartName := d.Get("chart_name").(string)
 	chartVersion := d.Get("chart_version").(string)
+	systemDefaultRegistry := d.Get("system_default_registry").(string)
 
 	log.Printf("[INFO] Creating App V2 %s at cluster ID %s", name, clusterID)
 
@@ -49,11 +50,16 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
 	}
 	d.Set("cluster_name", cluster.Name)
 
-	systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(appV2DefaultRegistryID)
+	globalSystemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID("local", appV2DefaultRegistryID)
 	if err != nil {
 		return err
 	}
-	d.Set("system_default_registry", systemDefaultRegistry.Value)
+
+	if systemDefaultRegistry != globalSystemDefaultRegistry.Value {
+		d.Set("system_default_registry", systemDefaultRegistry)
+	} else {
+		d.Set("system_default_registry", globalSystemDefaultRegistry.Value)
+	}
 
 	repo, chartInfo, err := infoAppV2(meta.(*Config), clusterID, repoName, chartName, chartVersion)
 	if err != nil {
--- a/rancher2/resource_rancher2_app_v2.go
+++ b/rancher2/resource_rancher2_app_v2.go
@@ -50,14 +50,12 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
 	}
 	d.Set("cluster_name", cluster.Name)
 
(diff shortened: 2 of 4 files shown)


nicholasSUSE:
@Jono-SUSE-Rancher @snasovich 

Hi, I need reviews on this PR however, I don't have permission to add reviewers. 
Please, can you help with that? 

Also, there is a failing automated test but it has nothing to do with PR code changes and I can't restart the drone build, how should I proceed?

Jono-SUSE-Rancher:
Hey @nicholasSUSE - I've added Andy, Sergey, and Bruno - whichever two get to take a look first.

a-blender (changes requested):
I'd update the tf docs `cluster_v2.md` to reflect the new functionality of `system_default_registry` now and make sure to test it having an empty value/not being set which would cause tf to use the global system registry within Rancher and make sure that still works. Lgtm

nicholasSUSE:
> I'd update the tf docs `cluster_v2.md` to reflect the new functionality of `system_default_registry` now and make sure to test it having an empty value/not being set which would cause tf to use the global system registry within Rancher and make sure that still works. Lgtm

Thank you for your review.
I have updated the documentation in a new commit. 
I have tested the empty value for `system_default_registry`

docs/resources/app_v2.md
```diff
@@ -40,6 +57,7 @@ The following arguments are supported:
 * `wait` - (Optional) Wait until app is deployed. Default: `true` (bool)
 * `annotations` - (Optional/Computed) Annotations for the app v2 (map)
 * `labels` - (Optional/Computed) Labels for the app v2 (map)
+* `system_default_registry` - (Optional/Computed) Private registry for system images
```
a-blender: The description should match the schema and have the field type, so

`system_default_registry` - (Optional/Computed) System default registry providing images for app deployment (string)`

docs/resources/app_v2.md
```diff
@@ -21,6 +21,23 @@ resource "rancher2_app_v2" "foo" {
 }
 ```
 
+### Create an App from a Helm Chart using a different registry
+
+The `system_default_registry` argument can override the global value at App installation time. If  argument is not provided, the global value for System Default Registry will be used instead.
```
a-blender: Suggestion: `at App installation.` and remove extra space between If/argument

a-blender (changes requested):

docs/resources/app_v2.md
nicholasSUSE: Keen eyes

docs/resources/app_v2.md
nicholasSUSE: Done

a-blender (approved):

rancher2/resource_rancher2_app_v2.go
```diff
@@ -92,7 +98,7 @@ func resourceRancher2AppV2Read(d *schema.ResourceData, meta interface{}) error {
 			d.Set("cluster_name", cluster.Name)
 		}
 		if systemDefaultRegistry, ok := d.Get("system_default_registry").(string); !ok || len(systemDefaultRegistry) == 0 {
-			systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(appV2DefaultRegistryID)
+			systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(clusterID, appV2DefaultRegistryID)
```
rohitsakala: Settings are not defined for downstream clusters I guess. Correct me if I am wrong. 

Settings are always defined in the local cluster *only*. So, I think we need to either send "local" here or get back to the previous way where "local" is hardcoded.

rancher2/resource_rancher2_app_v2.go
```diff
@@ -49,11 +50,16 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
...
 	d.Set("cluster_name", cluster.Name)
 
-	systemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(appV2DefaultRegistryID)
+	globalSystemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID("local", appV2DefaultRegistryID)
 	if err != nil {
 		return err
 	}
-	d.Set("system_default_registry", systemDefaultRegistry.Value)
+
+	if systemDefaultRegistry != globalSystemDefaultRegistry.Value {
```
rohitsakala: nit: 

if systemDefaultRegistry == "" {
   d.Set("system_default_registry", globalSystemDefaultRegistry.Value)
}

rohitsakala (changes requested):
Commented

rancher2/resource_rancher2_app_v2.go
nicholasSUSE: Done

rancher2/resource_rancher2_app_v2.go
nicholasSUSE: Done

nicholasSUSE:
@rohitsakala 

I have tested both scenarios:

Setting the system_default_registry variable
```
  system_default_registry = "${aws_instance.default_registry.public_dns}:5000"
```

And not setting.

The `resourceRancher2AppV2Read` function, returns the right value for `system_default_registry`.
The `d.Get("system_default_registry")` method, checks the state from terraform and not the actual configuration from Rancher or Downstream cluster. 

This way, whatever override we do with the new variable behavior will output the right value.

rancher2/resource_rancher2_app_v2.go
a-blender: @rohitsakala In Rancher yeah, settings are only defined in the local cluster and a change to a settings field applies to all downstream clusters. Thanks for spotting this.

rancher2/resource_rancher2_app_v2.go
```diff
@@ -49,11 +50,16 @@ func resourceRancher2AppV2Create(d *schema.ResourceData, meta interface{}) error
...
+	globalSystemDefaultRegistry, err := meta.(*Config).GetSettingV2ByID(appV2DefaultRegistryID)
 	if err != nil {
 		return err
 	}
-	d.Set("system_default_registry", systemDefaultRegistry.Value)
+
+	if systemDefaultRegistry == "" {
+		d.Set("system_default_registry", globalSystemDefaultRegistry.Value)
+	} else {
+		d.Set("system_default_registry", systemDefaultRegistry)
```
rohitsakala: @nicholasSUSE I think the else condition is not necessary since line 40 already has `system_default_registry` set by the user.

rohitsakala (changes requested):

rohitsakala (approved):

Issue #923, opened 2022-05-12
for_each in rancher2_registry keeps changing the order of the list that has not changed


sarg3nt:
Each time we `terraform apply` registries keep getting updated because the list order is getting changed.  Our local data is not changing, you can  `terraform apply` over and over and watch it flip back and forth.  Very annoying because we deploy this with CD and it's saying there are changes every time when there are not.

Here's our code:

```
variable "containerRepositories" {
  type = list(object({
    url      = string
    username = string
    password = string
  }))
  default     = []
  description = "A list of Docker container repositories."
}

 containerRepositories = [
    {
      url      = "dcc-docker-release.<redacted>"
      username = "bld_dcc"
      password = "<redacted>"
    },
    {
      url      = "dcc-docker-dev.<redacted>"
      username = "bld_dcc"
      password = "<redacted>"
    },
    {
      url      = "dcsws-docker-release.<redacted>"
      username = "bld_dcc"
      password = "<redacted>"
    },
    {
      url      = "dcsws-docker-dev.<redacted>"
      username = "bld_dcc"
      password = "<redacted>"
    }
  ]

resource "rancher2_registry" "regcred-build" {
  count        = length(var.containerRepositories) > 0 ? 1 : 0
  name         = "regcred"
  description  = "Registry entry for k8s to talk to customers Artifactory registries"
  project_id   = rancher2_project.jenkins-project-build.id
  namespace_id = rancher2_namespace.jenkins-namespace-build.id
  dynamic "registries" {
    for_each = var.containerRepositories
    content {
      address  = registries.value["url"]
      username = registries.value["username"]
      password = registries.value["password"]
    }
  }
}

```

The result of running terraform apply when there have been no updates. If we apply this, then run it again, they will “switch back”

```
~ resource "rancher2_registry" "regcred" {
        id           = "jenkins-dcs:regcred"
        name         = "regcred"
        # (5 unchanged attributes hidden)

      ~ registries {
          ~ address  = "dcsws-docker-release.<redacted>" -> "dcc-docker-release.<redacted>"
            # (2 unchanged attributes hidden)
        }
      ~ registries {
          ~ address  = "dcc-docker-release.<redacted>" -> "dcsws-docker-release.<redacted>"
            # (2 unchanged attributes hidden)
        }
        # (2 unchanged blocks hidden)
    }
```

We are running Terraform v1.1.9 and rancher/rancher2 v1.23.0

I posted this on the Terraform forum here
[...]

PR #1268, opened 2023-11-22
Fix the issue where the for_each in rancher2_registry keeps changing the order of the list that has not changed

jiaqiluo:
## Issue: 

 
https://github.com/rancher/terraform-provider-rancher2/issues/923

## Problem

 
When we keep invoking `terraform apply` multiple times, Terraform may report changes in the `rancher2_registry` resource when we do not change anything.  
Please check the linked GH issue for detailed examples. 

It happens because when terraform-provider-rancher2 flattens the registry credential, it needs to convert a Map into a List,  Golang does not guarantee the order of returned objects when traversing a Map, and terraform-provider-rancher2 saves items in the order of receiving,  so the order of items in the final List may change even though it is the same set of items. Therefore, Terraform detects and reports changes. 

## Solution

 
Preserve the order of existing items in the final List can prevent Terraform from reporting such changes

### Manual Testing

This fix is tested locally by adding extra logging lines which show that the order of existing items is preserved. 

### Automated Testing

The existing test case is updated. 

## QA Testing Considerations

 
the same case as the GH issue provides. 

none

### DIFF
--- a/rancher2/structure_registry.go
+++ b/rancher2/structure_registry.go
@@ -13,14 +13,20 @@ func flattenRegistryCredential(in map[string]projectClient.RegistryCredential, p
 	}
 
 	out := make([]interface{}, len(in))
-	lenP := len(p)
 	i := 0
+	// Golang does not guarantee the order of returned objects when traversing a Map;
+	// Terraform reports changes when the order of items is changed even though it is the same set of items in a List.
+	// Preserve the order of existing items in the final List can prevent Terraform from reporting such changes.
 	for key := range in {
-		var obj map[string]interface{}
-		if lenP <= i {
-			obj = make(map[string]interface{})
-		} else {
-			obj = p[i].(map[string]interface{})
+		obj := make(map[string]interface{})
+		existingIndex := -1
+		for idx, existing := range p {
+			tmp := existing.(map[string]interface{})
+			if key == tmp["address"] {
+				existingIndex = idx
+				obj = tmp
+				break
+			}
 		}
 
 		obj["address"] = key
@@ -33,7 +39,11 @@ func flattenRegistryCredential(in map[string]projectClient.RegistryCredential, p
 			obj["username"] = in[key].Username
 		}
 
-		out[i] = obj
+		if existingIndex != -1 {
+			out[existingIndex] = obj
+		} else {
+			out[i] = obj
+		}
 		i++
 	}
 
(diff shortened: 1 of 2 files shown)


rancher2/structure_registry.go
```diff
@@ -13,14 +13,20 @@ func flattenRegistryCredential(in map[string]projectClient.RegistryCredential, p
...
+	// Preserve the order of existing items in the final List can prevent Terraform from reporting such changes.
 	for key := range in {
-		var obj map[string]interface{}
-		if lenP <= i {
-			obj = make(map[string]interface{})
-		} else {
-			obj = p[i].(map[string]interface{})
+		obj := make(map[string]interface{})
+		existingIndex := -1
+		for idx, existing := range p {
```
a-blender, line 23: Might be worth it to extract this code into a util func

rancher2/structure_registry_test.go
```diff
@@ -24,17 +24,41 @@ func init() {
...
+			"password": "password5",
 		},
 	}
 	testDockerCredentialConf = &projectClient.DockerCredential{
 		ProjectID:   "project:test",
 		Name:        "name",
 		Description: "description",
 		Registries: map[string]projectClient.RegistryCredential{
-			"address": *testRegistryCredentialConf,
+			"rds-dev.tea1.inf.rancher.com":   {Username: "username5", Password: "password5"},
```
a-blender, line 57: You can realistically test this with only 3 registry entries instead of 5 :)

a-blender (changes requested):

rancher2/structure_registry_test.go
jiaqiluo: Based on my test, the more items on the list, the higher the chance of triggering the bug.

rancher2/structure_registry.go
jiaqiluo: I prefer to keep the logic here, instead of a standalone function, for two reasons: 1) It is not needed anywhere except here; 2) it introduces unnecessary complexity to handle the type conversations. 

We can put it into an until func next time when needed.

a-blender: That's fine! If there are more instances of it being used again we should pull it out into `util.go`.

a-blender (approved):

HarrisonWAffel (approved):

PR #1287, opened 2024-01-03
Update rancher2 config map import command

ip025:
fix config_map_v2 import example

## Problem
Most likely a copy/paste error in the documentation
 
## Solution
fix the resource name in the example
 
## Testing
none


matttrach (approved):
lgtm

matttrach:
since this is just documentation and there are no conflicts I am going to push it through.

PR #1296, opened 2024-01-24
Fix double base64-encoding of `ca_bundle` field

adamkpickering:
## Issue: 

For rancher/terraform-provider-rancher2#1297
 
## Problem

The user reports that configuring a `rancher_catalog_v2` resource with the `ca_bundle` field set fails. This is because the `caBundle` field in the configured `ClusterRepo` resource is base64-encoded twice. For more details, the JIRA issue gives a pretty good description.

Also note that there is a discrepancy between the documentation for configuring a repo/catalog via the Rancher UI, and the documentation for the `ca_bundle` field in the `rancher_catalog_v2` resource:

> Git-based chart repositories: You must add a base64 encoded copy of the CA certificate in DER format to the spec.caBundle field of the chart repo [link](https://ranchermanager.docs.rancher.com/v2.7/pages-for-subheaders/helm-charts-in-rancher#repositories)

> [ca_bundle](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/catalog_v2#ca_bundle) - (Optional) PEM encoded CA bundle which will be used to validate the repo's certificate (string) [link](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/catalog_v2#argument-reference)

## Solution

The cleanest solution, as far as @nicholasSUSE and I can tell, is to slightly change the behavior of the `ca_bundle` field so that the user sets it to a base64-encoded DER value. When expanding the terraform object into a `ClusterRepo`, the `rancher2` provider base64-decodes the passed `ca_bundle` value and sets the `spec.caBundle` field of the `ClusterRepo` to the result, which is of type `[]byte`. When the `encoding/json` package marshals the `ClusterRepo` struct created by the `rancher2` provider, it automatically base64-encodes the `caBundle` field (see the [docs for `json.Marshal`](https://pkg.go.dev/encoding/json#Marshal)), which appears to be what the Rancher API expects.
 
## Testing

This issue can be reproduced by running the terraform provider against a v2.7.6 Rancher cluster. Note that running it against a version of Rancher run via `go run main.go` or `dlv` doesn't work - it seems that the terraform provider chokes on the fact that something (I never figured out what) has the version `dev`.

### Manual Testing

On a Rancher v2.7.6 deployment running in k3d, I did the following tests using the terraform provider:
- Creating a `ClusterRepo` resource with the `caBundle` field set.
- Creating a `ClusterRepo` resource without the `caBundle` field set.
- Deleting a `ClusterRepo` resource.
- Updating a `ClusterRepo` resource by changing the `name` field in the terraform resource. This failed, but I determined that this failure is present on the `master` branch and thus not related to this PR. This seems wrong to me, but also seems like it may just be due to how the Rancher API works. I'm new to the Rancher project, so can someone who has more experience with Rancher check this?

### Automated Testing

I updated the `ca_bundle` field in the `expandCatalogV2` and `flattenCatalogV2` unit tests to reflect the HCL value being base64-encoded DER, and the ClusterRepo value being a `[]byte` containing unencoded DER.

## QA Testing Considerations

 
I don't *think* there is any issue with changing the behavior of the `ca_bundle` field, since it is broken as is. I could easily be wrong though...

### DIFF
--- a/rancher2/structure_catalog_v2.go
+++ b/rancher2/structure_catalog_v2.go
@@ -29,7 +29,8 @@ func flattenCatalogV2(d *schema.ResourceData, in *ClusterRepo) error {
 	}
 	d.Set("resource_version", in.ObjectMeta.ResourceVersion)
 
-	d.Set("ca_bundle", string(in.Spec.CABundle))
+	encodedCABundle := base64.StdEncoding.EncodeToString(in.Spec.CABundle)
+	d.Set("ca_bundle", encodedCABundle)
 	if in.Spec.Enabled != nil {
 		d.Set("enabled", *in.Spec.Enabled)
 	}
--- a/rancher2/structure_catalog_v2.go
+++ b/rancher2/structure_catalog_v2.go
@@ -1,6 +1,9 @@
 package rancher2
 
 import (
+	"encoding/base64"
+	"fmt"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
 )
@@ -46,9 +49,9 @@ func flattenCatalogV2(d *schema.ResourceData, in *ClusterRepo) error {
 
 // Expanders
 
-func expandCatalogV2(in *schema.ResourceData) *ClusterRepo {
+func expandCatalogV2(in *schema.ResourceData) (*ClusterRepo, error) {
 	if in == nil {
-		return nil
+		return nil, nil
 	}
 	obj := &ClusterRepo{}
 
@@ -69,7 +72,15 @@ func expandCatalogV2(in *schema.ResourceData) *ClusterRepo {
 		obj.ObjectMeta.ResourceVersion = v
 	}
 	if v, ok := in.Get("ca_bundle").(string); ok {
-		obj.Spec.CABundle = []byte(v)
+		// The rancher API expects the CA bundle to be in base64-encoded DER
+		// format. The caBundle field of ClusterRepo is of type []byte, so
+		// json.Marshal base64-encodes this field for us. So internally, we
+		// set the caBundle field to non-base64-encoded DER.
+		decodedCABundle, err := base64.StdEncoding.DecodeString(v)
+		if err != nil {
+			return nil, fmt.Errorf("failed to base64-decode ca_bundle: %w", err)
+		}
+		obj.Spec.CABundle = decodedCABundle
 	}
 	if v, ok := in.Get("enabled").(bool); ok {
 		obj.Spec.Enabled = &v
@@ -102,5 +113,5 @@ func expandCatalogV2(in *schema.ResourceData) *ClusterRepo {
 		obj.Spec.URL = v
 	}
 
-	return obj
+	return obj, nil
 }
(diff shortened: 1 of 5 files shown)


nicholasSUSE (approved):
I approved, but please wait for @rohitsakala review before merging

rancher2/structure_catalog_v2.go
```diff
@@ -26,7 +29,8 @@ func flattenCatalogV2(d *schema.ResourceData, in *ClusterRepo) error {
 	}
 	d.Set("resource_version", in.ObjectMeta.ResourceVersion)
 
-	d.Set("ca_bundle", string(in.Spec.CABundle))
+	encodedCABundle := base64.StdEncoding.EncodeToString(in.Spec.CABundle)
```
rohitsakala, line 32: Why are we encoding this? 

This is the clusterRepo 

```
apiVersion: catalog.cattle.io/v1
kind: ClusterRepo
metadata:
  creationTimestamp: '2024-01-29T17:06:01Z'
  generation: 3
  managedFields:
    - apiVersion: catalog.cattle.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        f:spec:
          .: {}
          f:caBundle: {}
          f:clientSecret: {}
          f:url: {}
      manager: rancher
      operation: Update
      time: '2024-01-29T17:06:42Z'
    - apiVersion: catalog.cattle.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        f:status:
          .: {}
          f:conditions: {}
          f:downloadTime: {}
          f:observedGeneration: {}
      manager: rancher
      operation: Update
      subresource: status
      time: '2024-01-29T17:06:42Z'
  name: testing
  resourceVersion: '186876'
  uid: 3dee071c-ec22-413a-a506-d9c1d10035b2
spec:
  caBundle: >-
    MIIB6TCCAY+gAwIBAgIIdt3GyRspkn4wCgYIKoZIzj0EAwIwRjEcMBoGA1UEChMTZHluYW1pY2xpc3RlbmVyLW9yZzEmMCQGA1UEAwwdZHluYW1pY2xpc3RlbmVyLWNhQDE3MDY1MjU5ODAwHhcNMjQwMTI5MTA1OTQwWhcNMjUwMTI4MTEwMjMzWjAkMRAwDgYDVQQKEwdkeW5hbWljMRAwDgYDVQQDEwdkeW5hbWljMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEkaT5FLdTzgs6yNaa8PltS785HCW2oNGFsm/4ZBgoMgQC63QpGA5+asRePP9tvSMmNf97zGc4T13cwkfu4RDYtKOBiDCBhTAOBgNVHQ8BAf8EBAMCBaAwEwYDVR0lBAwwCgYIKwYBBQUHAwEwHwYDVR0jBBgwFoAUZ2qHTZJTBsvTvg7FJDVSss8YB38wPQYDVR0RBDYwNIIJbG9jYWxob3N0ghVyYW5jaGVyLmNhdHRsZS1zeXN0ZW2HBH8AAAGHBKwRAAKHBKzUUUQwCgYIKoZIzj0EAwIDSAAwRQIhAKnE7zEoSuwX6XMUcffPC/LYpIqEGieQLTjMF8QQggYgAiB5eHGz7JY8biWQA4m5imWIRbTqZ9RPnflrKufZnCCBxg==
  clientSecret: null
  url: https://172.212.81.68/dashboard/
 ````
 
 The in.spec.Cabundle will be `MIIB6TCCAY+gAwIBAgIIdt3GyRspk...` Why can't we show this ? Why should we encode this?

Comment (nicholasSUSE) [inline rancher2/structure_catalog_v2.go:32]: Let's test the following:
```
output "some_NAME" {
  value = <the_catalog_resource>.ca_bundle
}
```

to validate this

rohitsakala (approved):

Issue #1189, opened 2023-08-03
[RFE] Add new resources for PodSecurityAdmissionConfigurationTemplate


a-blender:
### Is your feature request related to a problem? Please describe.

Rancher v2.7.2 now comes w/ a new CRD for [Pod Security Admission (PSA) Configuration Templates](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates) that can only be provisioned manually.

### Describe the solution you'd like

Add a new resource/datasource for [`rancher2_pod_security_admission_configuration_template`](https://github.com/rancher/rancher/blob/release/v2.7/pkg/client/generated/management/v3/zz_generated_pod_security_admission_configuration_template.go) so users can manage _custom_ admission configuration templates via Terraform for RKE and RKE2/K3s clusters. This is a continuation of the solution for PSACT support https://github.com/rancher/terraform-provider-rancher2/issues/1112.

### Additional context

SURE-6756

snasovich:
This is not going to make it into TFP 4.0.0 release (TFP release corresponding to Rancher v2.8.0 release). For engineer to be assigned to it in the future, please consider work done in https://github.com/rancher/terraform-provider-rancher2/pull/1246 as a starting point.

jiaqiluo:
PR was merged, now waiting for an RC.

jiaqiluo:
Can be validated in https://github.com/rancher/terraform-provider-rancher2/tree/v4.1.0-rc1

Josh-Diamond:
Ticket #1189 - Test Results - ✅ 
---

Verified with tfp-rancher2 `v4.1.0-rc1` on Rancher `v2.8-c4743333c5d37c32e91b6de5e2c6b1c23ac90fa8-head`:

| Scenario | Test Case | Result |
|---|---|---|
| 1. | create, update, and delete `rancher2_pod_security_admission_configuration_template` | ✅ |
| 2. | delete a `rancher2_pod_security_admission_configuration_template` resource when it is used by existing clusters | ✅ |
| 3. | create and delete a `rancher2_pod_security_admission_configuration_template` and an RKE2 or RKE1 cluster that uses it in the same TF configuration | ✅ |
| 4. | fetch an existing `rancher2_pod_security_admission_configuration_template` data source | ✅ |

---

Scenario 1 - ✅ 
1. Fresh install of Rancher `v2.8-head`
[...]

PR #1302, opened 2024-02-02
add PSACT custom resource

jiaqiluo:
## Issue: 

 
https://github.com/rancher/terraform-provider-rancher2/issues/1189
## Problem

 
Rancher v2.7.2 supports a new CRD for [Pod Security Admission (PSA) Configuration Templates](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates) that can only be created outside of TF manually.

## Solution

 
Add a new resource and data source named  `rancher2_pod_security_admission_configuration_template`  for [rancher2_pod_security_admission_configuration_template](https://github.com/rancher/rancher/blob/release/v2.7/pkg/client/generated/management/v3/zz_generated_pod_security_admission_configuration_template.go?rgh-link-date=2023-08-03T18%3A34%3A57Z) so users can manage custom admission configuration templates via Terraform for RKE and RKE2/K3s clusters.

### Manual Testing

- Tested creating, modifying, and removing a custom PSACT resource. All tests are successful. 
- Tested creating a template with bad data and this failed as expected.
- Tested creating and destroying a PSACT along with an RKE2 cluster and RKE1 cluster. All tests are successful. 
- Tested fetching an existing `rancher2_pod_security_admission_configuration_template` data source

### Automated Testing

**unit test**

The standard sets of tests are added for the new `Resource` & `DataSource`. 

**acceptance test**

The acceptance tests are added but "disabled" by making the function names lowercase. This is because the current framework does not support running acceptance tests in Rancher v2.7.9 and v2.8.1. 

Here is how the acceptance test works:
- step 1: start a k3s cluster, currently using the version`v1.19.10-k3s1` 
- step 2: deploy Rancher v2.3.6 using Helm 
- step 3: create a set of Resources via TF into the Rancher setup, and validate the expected values present
- step 4: upgrade  Rancher to 2.5.9 using Helm

Steps 3 and 4 are repeated to upgrade and test on Rancher v2.4.13,  v2.5.9, and v2.6.2. 

To add the tests for the PSACT resource, I tried to add Rancher v2.7.9 and v2.8.1 to the version list. However, it did not work due to the following reasons:
 - the local cluster version, v1.19.10-k3s1, is too old for Rancher v2.7.9 and v2.8.1, as well as a newer version of Cert-manager, to run
 - Some legacy resources, such as rancher2_app, can be not deployed in Rancher v2.7.2 and above. More specifically the app used in the test is deprecated and no new chart is available.  

To make the acceptance test work, we need to:
- add support for upgrading the local cluster
- add support for deploying new versions of Rancher and Cert-manger
- filter out the broken tests on the newer version of Rancher 

That work is large enough to be a separate task. (Update: the issue was made https://github.com/rancher/terraform-provider-rancher2/issues/1308) 

The good news is that the manual and automated tests that QA performs can cover more than the acceptance I planned to add. 

## QA Testing Considerations

 
- create, update, and delete `rancher2_pod_security_admission_configuration_template` 
- delete a `rancher2_pod_security_admission_configuration_template` resource when it is used by existing clusters
- create and delete a `rancher2_pod_security_admission_configuration_template`  and an RKE2 or RKE1 cluster that uses it in the same TF configuration
- fetch an existing `rancher2_pod_security_admission_configuration_template` data source

The new resource and data source themselves should not affect anything existing. 

QA should validate there is no regression for deleting a `rancher2_cluster_v2` resource because one change is made in this PR.


rancher2/resource_rancher2_cluster_v2.go
```diff
@@ -202,6 +202,28 @@ func resourceRancher2ClusterV2Delete(d *schema.ResourceData, meta interface{}) e
...
+			Pending:    []string{"removing"},
+			Target:     []string{"removed"},
+			Refresh:    clusterStateRefreshFunc(client, v1ClusterName),
+			Timeout:    d.Timeout(schema.TimeoutDelete),
+			Delay:      1 * time.Second,
+			MinTimeout: 3 * time.Second,
+		}
+		_, waitErr = stateConf.WaitForState()
+		if waitErr != nil {
+			return fmt.Errorf("[ERROR] waiting for cluster (%s) to be removed: %s", cluster.ID, waitErr)
```
bfbachmann: Minor: could wrap this error with %w. Same goes for the other errors created with `fmt.Errorf` below in this PR.
```suggestion
			return fmt.Errorf("[ERROR] waiting for cluster (%s) to be removed: %w", cluster.ID, waitErr)
```

rancher2/resource_rancher2_pod_security_admission_configuration_template.go
```diff
@@ -0,0 +1,162 @@
...
+	log.Printf("[INFO] Deleting PodSecurityAdmissionConfigurationTemplate with ID %s", id)
+	client, err := meta.(*Config).ManagementClient()
+	if err != nil {
+		return err
+	}
+
+	psact, err := client.PodSecurityAdmissionConfigurationTemplate.ByID(id)
+	if err != nil {
+		if IsNotFound(err) || IsForbidden(err) {
+			log.Printf("[INFO] PodSecurityAdmissionConfigurationTemplate with ID %s not found.", id)
```
bfbachmann, line 117: Will we get the "forbidden" error if the PSACT is not found? This log suggests that's the case, but that seems strange.

rancher2/resource_rancher2_pod_security_admission_configuration_template_test.go
```diff
@@ -0,0 +1,133 @@
...
+  exemptions {
+    usernames = ["testuser"]
+    runtime_classes = ["testclass"]
+    namespaces = ["ingress-nginx","kube-system"]
+  }
+}
+`
+)
+
+func init() {}
```
bfbachmann: Can we remove this since it's empty?

bfbachmann (commented):
LGTM for the most part. Just small comments.

rancher2/resource_rancher2_pod_security_admission_configuration_template.go
jiaqiluo: TBH, I did not hit the forbidden error myself while testing, but it looks like a custom to check the notFound error along with the forbidden error in this repo.  To make it reasonable, my best guess is that there is a chance to hit the forbidden error if TF requests the resource while it is being removed on the server side.

rancher2/resource_rancher2_pod_security_admission_configuration_template_test.go
jiaqiluo: fixed.

rancher2/resource_rancher2_cluster_v2.go
jiaqiluo: fixed.

jakefhyde (approved):

bfbachmann (approved):

PR #1331, opened 2024-03-28
Add search option exact_match for data source rancher2_principal

manhtukhang:
## Problem
When using the data source rancher2_principal to search for an LDAP user, sometimes it returns the wrong result if the inputted name has multiple matched results.
 
## Solution
Because the provider picks first element in the list returned from Rancher API. But that list
is unsorted or just in random order, therefore, picking the first element is not the best way. We should loop through it to find the exact match, but changing the default behavior can break compatibility.
So I decided to add an option `exact_match` to control this behavior.
 
## Testing
Set option `exact_match=true` must only return an exactly matched user as the inputted name


wgjak47:
Do we really need a exact_match option? The data source rancher2_principal will return a random result when we have multiple principals like k8s-admin, k8s-admin-dev, k8s-admin-staging. It is meaningless when you set name "k8s-admin" but return the "k8s-admin-dev".

manhtukhang:
> Do we really need a exact_match option? The data source rancher2_principal will return a random result when we have multiple principals like k8s-admin, k8s-admin-dev, k8s-admin-staging. It is meaningless when you set name "k8s-admin" but return the "k8s-admin-dev".

@wgjak47 I want to keep compatibility with current behavior, that won't break any use case, in the future release (2 or more versions) we can make this `true` by default.

@jakefhyde @jiaqiluo Could you take a look at this PR? Thanks!

jiaqiluo (approved):
Hi @manhtukhang, thank you for reporting the issue and making the fix!  The PR LGTM. 
I am going to add the issue to the team's working queue, but would like to let you know that when the fix is merged and a new version is released will depend on the team's priority and capacity. Thank you for understanding.

matttrach (approved):

PR #1335, opened 2024-04-04
fix: wrong result of data rancher2_principal

wgjak47:
## Problem
When using the data source rancher2_principal to search for a principal, it may return wrong result when one principal name is the prefix of another principal, for example: k8s-admin, k8s-admin-test
 
## Solution
In this code , check the return result is really match the request name.
 
## Testing
  As the issue describe.

### Manual Testing

Now the data source rancher2_principal will always return k8s-admin when you set the name as k8s-admin, instead of return k8s-admin-test.


matttrach:
This appears to still be relevant and I was able to update it to the latest from main without conflicts so I am moving forward with this.
@wgjak47, is this still valuable/relevant?

matttrach (approved):

matttrach:
This looks good from my perspective, but it needs someone from the team responsible for authentication to approve before I can merge.

matttrach:
nvm, looks like the team has chosen a different solution for this #1331, closing for now, thank you for your contribution!

PR #1351, opened 2024-05-17
Github Action Migration

HarrisonWAffel:
A successful release run can be seen here: https://github.com/HarrisonWAffel/terraform-provider-rancher2/actions/runs/9135094356
The github release and artifacts can be seen here: https://github.com/HarrisonWAffel/terraform-provider-rancher2/releases/tag/v0.0.0

A successful pre-release run can be seen here: https://github.com/HarrisonWAffel/terraform-provider-rancher2/actions/runs/9134965367
The github release and artifacts can be seen here: https://github.com/HarrisonWAffel/terraform-provider-rancher2/releases/tag/v0.0.0-rc1

I've checked the GPG signing process using my personal GPG key 

```
gpg: Signature made Fri May 17 23:01:41 2024 UTC
gpg:                using RSA key B1BB4CAE50A884D616C4DCF69A6B757E291A63DF
gpg: Good signature from "Harrison Affel <harrison.affel@suse.com>" [unknown]
Primary key fingerprint: B1BB 4CAE 50A8 84D6 16C4  DCF6 9A6B 757E 291A 63DF
```

In a similar fashion as the rke terraform provider, I've removed the `golint` step in the make file. `golint` is no longer maintained and does not work properly with go modules. We should introduce a better linting solution in the near future, however this PR currently only focuses on GHA migration.

SURE-8006

PR #1359, opened 2024-06-18
Add pull request workflow

HarrisonWAffel:
## Problem
#1351 missed a workflow for pull requests, as such we are not building and packaging the provider before allowing PRs to be merged 
 
## Solution
Add a pull request workflow which builds and packages the provider for each PR. This PR does not include the acceptance tests, as additional effort needs to be put into how we can modernize those tests such that they can work against a more modern version of rancher. 
 
## Testing
I've tested this workflow on my personal fork 

Actions run: https://github.com/HarrisonWAffel/terraform-provider-rancher2/actions/runs/9571691853/job/26389359180

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -1,0 +1,16 @@
+name: Pull Request
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: build binaries
+        env:
+          CROSS: 1
+          VERSION: ${{ github.ref_name }}
+        run: |
+          make build-rancher


snasovich (commented):
LGTM at a glance but some observations/questions:
1. On Drone we ran `make docker-build` and `make build-rancher` is basically equivalent of it minus Docker. Why do we need to run `make package-rancher`?
2. Any specific reason we're running on `ubuntu-20.04` vs. `ubuntu-latest`?

HarrisonWAffel:
I've updated to `ubuntu-latest`, and removed `make package-rancher`. Initial thought process was to more closely mirror what release CI does, but if this step was not being done while running in drone I agree we should drop it.

jiaqiluo (approved):
LGTM

.github/workflows/pull_request.yaml
```diff
@@ -0,0 +1,16 @@
+name: Pull Request
+
+on:
+  pull_request:
+
+jobs:
+  buildAndPackage:
```
snasovich: Hate to nit but we should rename it to just `build` if we removed packaging step.

HarrisonWAffel: Good catch

snasovich (approved):
LGTM as soon as https://github.com/rancher/terraform-provider-rancher2/actions/runs/9572629031/job/26392362041?pr=1359 succeeds

PR #1361, opened 2024-06-28
fix: try removing the decode

matttrach:
Pre-release is failing at decoding the gpg private key. My hunch is that the secret isn't base64 encoded. This will remove the decode call in the workflow so that we can try again.


snasovich (approved):
LGTM.
For reference, here is the failing run: https://github.com/rancher/terraform-provider-rancher2/actions/runs/9717306724/job/26822716464
🤞 it works after the change

PR #1364, opened 2024-07-01
Use unzip instead of gunzip in `setup-provider.sh`

HarrisonWAffel:
## Problem

When testing `setup-provider.sh` I was unable to properly download both `4.0.0-rc5` or `4.2.0-rc3`, as `gunzip` reported an `unknown compression format`. This resulted in an empty file in the directory which should include the provider binary. This script references `gzip`, but does not use it when uncompressing the provider zip file. After switching the command, I also encountered issues with `gzip`. While the command executed without error, the downloaded file was not properly decompressed and thus was not useable. 

In both scenarios running `terraform plan` on a file using the downloaded provider resulted in 

```
provider "terraform.local/local/rancher2" to obtain schema: fork/exec .terraform/providers/terraform.local/local/rancher2/4.2.0-rc3/darwin_arm64/terraform-provider-rancher2: exec
│ format error..
```
 
## Solution
Remove `gunzip` and `gzip` in favor of only using `unzip`. `unzip` is specifically built for zip files, and is installed by default on mac. For linux, it can be installed via `apt`. This change also removes a `brew` command invocation which would fail on linux systems.
 
## Testing

Run this script against the latest rancher2 version (`4.2.0-rc3`) as well as the latest rke version (`v1.5.0`) and ensure that a simple terraform file can be planned / applied as expected. Run the `file` command against the downloaded provider binary, and ensure it is a valid executable for your system and is not compressed. 

### Manual Testing
I've done the above

n/a

### DIFF
--- a/setup-provider.sh
+++ b/setup-provider.sh
@@ -28,11 +28,10 @@ PROVIDER=$1
 VERSION=$2
 VERSION_TAG=$(echo $2 | cut -c 2-)
 
-# Install gzip
-
-if ! command -v "gzip" &> /dev/null; then
-  echo "Missing gzip. Installing..."
-  brew install gzip
+# Check for unzip
+if ! command -v "unzip" &> /dev/null; then
+  echo "Missing required tool unzip. unzip can be installed via apt on Linux"
+  exit 1
 fi
 
 # Download binary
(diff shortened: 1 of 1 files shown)


matttrach (approved):
lgtm

PR #1371, opened 2024-07-16
Update clusterrepo on rancher2 terraform provider with oci fields

diogoasouza:
Adding new fields to the ClusterRepo CRD which were added to `rancher/rancher` in the PR https://github.com/rancher/rancher/pull/45062

### DIFF
--- a/rancher2/structure_catalog_v2.go
+++ b/rancher2/structure_catalog_v2.go
@@ -41,6 +41,15 @@ func flattenCatalogV2(d *schema.ResourceData, in *ClusterRepo) error {
 		d.Set("secret_name", in.Spec.ClientSecret.Name)
 		d.Set("secret_namespace", in.Spec.ClientSecret.Namespace)
 	}
+
+	if in.Spec.ExponentialBackOffValues != nil {
+		d.Set("exponential_backoff_min_wait", in.Spec.ExponentialBackOffValues.MinWait)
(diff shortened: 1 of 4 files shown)


rancher2/structure_catalog_v2.go
```diff
@@ -96,6 +105,39 @@ func expandCatalogV2(in *schema.ResourceData) (*ClusterRepo, error) {
...
+				MinWait: v,
+			}
+		}
+	}
+
+	if v, ok := in.Get("exponential_backoff_max_wait").(int); ok {
+		if obj.Spec.ExponentialBackOffValues != nil {
+			obj.Spec.ExponentialBackOffValues.MaxWait = v
+		} else {
+			obj.Spec.ExponentialBackOffValues = &v1.ExponentialBackOffValues{
```
rohitsakala, line 125: Correct me here. 

Seems like line no 125 is is removing minWait completely.

rohitsakala (changes requested):

rancher2/structure_catalog_v2.go
diogoasouza: It only goes into the `else` condition if `ExponentialBackOffValues` is nil, so in this case there wouldn't be a minWait

matttrach (approved):

rohitsakala (approved):

Issue #1354, opened 2024-05-29
[BUG] Provider crash on changing rancher2_custom_user_token's password


mouellet:
### Rancher Server Setup
- Rancher version: v2.8.2

### Information about the Cluster
- Kubernetes version: v1.27.11
- Cluster Type: Downstream/Custom/Azure

### User Information
N/D

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? v4.1.0
- What is the version of Terraform in use? v1.5.7

### Describe the bug
Changing the 'password' attribute of the `rancher2_custom_user_token` resource doesn't recreate it as documented.

https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/custom_user_token.md?plain=1#L15
> Tokens can't be updated once created. Any diff in token data will recreate the token. If any token expire, Rancher2 provider will generate a diff to regenerate it. 

https://github.com/rancher/terraform-provider-rancher2/blob/master/docs/resources/custom_user_token.md?plain=1#L51
> [password](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/custom_user_token#password) - (Required/ForceNew) The user password (string)

### To Reproduce
1. create a `rancher2_custom_user_token` resource 
2. change the password attribute

### Actual Result
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ rancher2_custom_user_token.this,
│ provider "provider[\"registry.terraform.io/rancher/rancher2\"]" produced an
│ unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
```

### Expected Result
Resource should be recreated

### Additional context
Schema is missing a `ForceNew: true,` here: https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_custom_user_token.go#L11-L16

matttrach:
Thank you so much! I agree with your assessment, I will work on making this change for our next release.

markusewalker:
Validated that I do not see a crash in `rancher2_custom_resource` with 5.0.0-rc1. See details below:

**ENVIRONMENT DETAILS**
- Rancher install: Docker
- Rancher version: `main` https://github.com/rancher/rancher/commit/78a3ca992970297f8fc3df0ade8cea189ae5fc68
- Terraform version: `5.0.0-rc1`

**TEST RESULTS**
|#|Scenario|Test Result|
[...]

PR #1378, opened 2024-08-01
fix: custom user token changes replace resource

matttrach:
Addresses https://github.com/rancher/terraform-provider-rancher2/issues/1354

As part of this I am also including a change to increase the verbosity of the tests so accentuate the fact that unit tests are being run and pass in the CI.


jiaqiluo (approved):
lgtm

PR #1380, opened 2024-08-02
Add `must_change_password` attribute to `rancher2_user` resource

alegrey91:
## Problem
When declaring users with `rancher2_user`, there is no option to ask the user to change their password on the next login like there is in the UI.

 
## Solution
Add the missing attribute in order to allow the user to force the reset of rancher users managed with the provider.

 
### Manual Testing

Setup a rancher instance from docker:

```bash
sudo docker run --privileged -it --rm -p 8080:80 -p 443:443 rancher/rancher
```

In order to test this new attribute I've used the following terraform script to ensure the new user had this attribute set:
```hcl
resource "rancher2_user" "testuser" {
  name     = "Test User"
  username = "testuser"
  password = "password12345"
  must_change_password = true
  enabled     = true
}

# Assign global role bindings to the user (optional)
resource "rancher2_global_role_binding" "new_user_role" {
  user_id       = rancher2_user.testuser.id
  global_role_id = "user-base"
}
```

Once terraform completed its run, you can login with the new created user `testuser`.
Once logged in the UI will show the reset password page.

### DIFF
--- a/rancher2/resource_rancher2_user_test.go
+++ b/rancher2/resource_rancher2_user_test.go
@@ -15,8 +15,9 @@ const (
 )
 
 var (
-	testAccRancher2User       string
-	testAccRancher2UserUpdate string
+	testAccRancher2User                    string
+	testAccRancher2UserUpdate              string
+	testAccRancher2UserUpdatePasswordReset string
 )
 
 func init() {
@@ -26,6 +27,7 @@ resource "` + testAccRancher2UserType + `" "foo" {
   username = "foo"
   password = "TestACC123456"
   enabled = true
+  must_change_password = true
 }
 `
 	testAccRancher2UserUpdate = `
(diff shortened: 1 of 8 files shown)


rancher2/resource_rancher2_user_test.go
```diff
@@ -26,6 +26,7 @@ resource "` + testAccRancher2UserType + `" "foo" {
   username = "foo"
   password = "TestACC123456"
   enabled = true
+  must_change_password = true
 }
```
bigkevmcd, line 31: It looks like there should be a test where `must_change_password` is `false`?

bigkevmcd (commented):
Looks ok, I think there's a missing test tho'.

rancher2/resource_rancher2_user_test.go
alegrey91: test added :)

bigkevmcd (approved):

HarrisonWAffel (approved):

matttrach:
per private discussion this PR needs to be reverted and prepared for resubmission after the next release

Issue #1398, opened 2024-09-05
[BUG] Certain new users cannot create clusters with Terraform


matttrach:
### Rancher Server Setup
- Rancher version: v2.7.9
- Installation option (Docker install/Helm Chart): unknown
- Proxy/Cert Details: unknown

### Information about the Cluster
- Kubernetes version: unknown
- Cluster Type (Local/Downstream): downstream, harvester

### User Information
- What is the role of the user logged in? unknown

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? v4.2.0
- What is the version of Terraform in use? unknown

### Describe the bug

A specific group of users is not able to start the cluster using Terraform.
From UI the cluster creation works for the same users.

See SURE-8955 for more details

matttrach:
This involves ldap integration. It appears that the user(s) is being denied when they have many different groups attached.
The user suggested updating the machine config code so that forbidden responses would be subject to timeouts.
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_machine_config_v2.go#L51-L78

matttrach:
I think the user's suggestion may not work because if we are getting a timeout from ldap due to having to parse a user with a lot of groups then a retry won't have a different result. Instead, the ldap integration may need to be configured with longer timeouts, which is not configured in this resource.

matttrach:
I have requested more information from the user regarding the Terraform config. If there is another resource generating the user in the same apply then potentially this could be fixed in the Terraform dependency chain, otherwise I may need to get some advice from the team that maintains the ldap integration.

matttrach:
Got a second option on this one and it turns out I was wrong. It looks like the user's suggestion is a good idea, so I will proceed in making the suggested change.

joesims22:
Validated on v2.10-head id `691f7ae` and tfp `v6.0.0-rc1` successfully.

PR #1400, opened 2024-09-05
fix: remove extra return in machine template

matttrach:
addresses https://github.com/rancher/terraform-provider-rancher2/issues/1398

### DIFF
--- a/rancher2/resource_rancher2_machine_config_v2.go
+++ b/rancher2/resource_rancher2_machine_config_v2.go
@@ -60,10 +60,6 @@ func waitForMachineConfigV2(d *schema.ResourceData, config *Config, interval tim
 			return nil
 		}
 		log.Printf("[INFO] Retrying on error Refreshing Machine Config V2 %s: %v", d.Id(), err)
-		if IsNotFound(err) || IsForbidden(err) {
-			d.SetId("")
-			return fmt.Errorf("Machine Config V2 %s not found: %s", d.Id(), err)
-		}
 		if IsNotAccessibleByID(err) {
 			// Restarting clients to update RBAC
 			config.RestartClients()
--- a/rancher2/resource_rancher2_machine_config_v2.go
+++ b/rancher2/resource_rancher2_machine_config_v2.go
@@ -68,6 +68,7 @@ func waitForMachineConfigV2(d *schema.ResourceData, config *Config, interval tim
 		select {
 		case <-time.After(rancher2RetriesWait * time.Second):
 		case <-ctx.Done():
+			d.SetId("")
 			return fmt.Errorf("Timeout waiting for machine config V2 ID %s", d.Id())
 		}
 	}


matttrach:
When accepted, this will need to be backported to v5 and v4

rancher2/resource_rancher2_machine_config_v2.go
```diff
@@ -60,10 +60,6 @@ func waitForMachineConfigV2(d *schema.ResourceData, config *Config, interval tim
 			return nil
 		}
 		log.Printf("[INFO] Retrying on error Refreshing Machine Config V2 %s: %v", d.Id(), err)
-		if IsNotFound(err) || IsForbidden(err) {
```
jiaqiluo, line 63: I don't believe the for loop is being used correctly in this function. If getMachineConfigV2ByID returns an error, the function will remain stuck in the loop indefinitely and never reach the select block. I’d recommend refactoring the logic to use a backoff function, similar to [OnError](https://pkg.go.dev/k8s.io/client-go/util/retry@v0.30.1#OnError), if you prefer not to introduce a new module.

jiaqiluo (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
matttrach: but the select block is inside the loop

matttrach: with this change there are only two return scenarios:
1. getMachineConfigV2ByID returns nil for the error (there wasn't an error)
2. the time is after the timeout and the context is done

am I missing something?

matttrach: I am not a fan of loops that don't end like this one, but without a good test suite (something I am working on) I am trying to reduce impact as much as possible.

jiaqiluo: My bad. I misread the code and thought the select part was outside of the for loop when reviewing the code.  
~But we still have the problem that the loop never ends: if `getMachineConfigV2ByID` somehow always returns a 404/403 error, then the function goes to the `case <-time.After(rancher2RetriesWait * time.Second):` case and keep retrying.  Right?~

Never mind, the `ctx` will time out and make the function exit the loop.

jiaqiluo (changes requested):

rancher2/resource_rancher2_machine_config_v2.go
jiaqiluo: I think we should unset the ID in the case of timing out, ie. adding `d.SetId("")` to the  `case <-ctx.Done():` section.

matttrach: done, thank you so much for your help!

jiaqiluo (approved):

PR #1422, opened 2024-10-11
make `must_change_password` attribute immutable after creation

alegrey91:
## Problem
Currently, the `must_change_password` is not immutable. Since **rancher** doesn't allow to change that option after the user creation, we should handle this from the terraform provider.

 
## Solution
Set the `must_change_password` attribute as immutable after creation.

 
### Manual Testing

Setup a rancher instance from docker:

```bash
sudo docker run --privileged -it --rm -p 8080:80 -p 443:443 rancher/rancher
```

In order to test this new attribute I've used the following terraform script to ensure the new user had this attribute set:
```hcl
resource "rancher2_user" "testuser" {
  name     = "Test User"
  username = "testuser"
  password = "password12345"
  must_change_password = true
  enabled     = true
}

# Assign global role bindings to the user (optional)
resource "rancher2_global_role_binding" "new_user_role" {
  user_id       = rancher2_user.testuser.id
  global_role_id = "user-base"
}
```

Once terraform completed its run, you can login with the new created user `testuser`.
Once logged in the UI will show the reset password page.

After that, you can change `must_change_password` attribute, and verify that terraform needs to destroy the resource in order to set the new value.

### DIFF
--- a/docs/data-sources/user.md
+++ b/docs/data-sources/user.md
@@ -26,5 +26,6 @@ data "rancher2_user" "foo" {
 * `name` - (Computed) The user common name (string)
 * `annotations` - (Computed) Annotations of the resource (map)
 * `enabled` - (Computed) The user is enabled (bool)
+* `must_change_password` - (Computed) The user must change password at first login (bool)
 * `principal_ids` - (Computed) The user principal IDs (list)
 * `labels` - (Computed) Labels of the resource (map)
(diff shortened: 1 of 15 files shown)


bigkevmcd:
This change looks fine to me, following @alegrey91's instructions works too.

Checked against.

https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#forcenew

andreas-kupries (approved):
looks ok.

bigkevmcd (approved):

matttrach:
Based on https://github.com/rancher/rancher/issues/45736#issuecomment-2427609100
I am setting this to ready for release (it appears to have already been checked by QA against v6.0.0-rc2)

This appears to be targeting only Rancher v2.10, which correlates to to our v6

matttrach:
This was released with https://github.com/rancher/terraform-provider-rancher2/releases/tag/v6.0.0

Issue #1426, opened 2024-10-17
[BUG] rke2 machineSelectorFiles.fileSources have multiple duplicate entries


kinarashah:
### Rancher Server Setup
- Rancher version: v2.9.1

### Information about the Cluster
- Kubernetes version: Any
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster?  RKE2 Node driver cluster 

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? Built locally off master commit - `8356b631d76f5fd69276906c405233571beb6bdc` 
- What is the version of Terraform in use? Terraform v1.9.8

### Describe the bug
1. The default value of `dynamic` field using terraform provider should be consistent with Rancher API. `dynamic` is false when editing cluster object using API/UI but set to `true` using terraform. 
2. Multiple duplicate entries for `fileSources` are added to the cluster object under `machineSelectorFiles`. Only one entry is expected for the same secret and hash value. 

### To Reproduce
- Provision a RKE2 node driver cluster using terraform, make sure to set   `default_pod_security_admission_configuration_template_name`. I set it to "rancher-restricted". 
- After the cluster is provisioned, edit `snapshot_retention` or any other cluster field to trigger cluster reconciliation. 
- Observe that a new entry is added to the cluster object: 
```
    - fileSources:
      - configMap:
          name: ""
        secret:
          items:
          - dynamic: true
            hash: LW5oTV9pmpX7+xWMjgC3IgbHtLkSMlCgyaKXG13CihA=
            key: admission-config-psact
            path: /etc/rancher/rke2/config/rancher-psact.yaml
          name: molmedocis-admission-configuration-psact
      machineLabelSelector:
        matchLabels:
          rke.cattle.io/control-plane-role: "true"
    - fileSources:
      - configMap:
          name: ""
        secret:
          items:
          - hash: LW5oTV9pmpX7+xWMjgC3IgbHtLkSMlCgyaKXG13CihA=
            key: admission-config-psact
            path: /etc/rancher/rke2/config/rancher-psact.yaml
          name: molmedocis-admission-configuration-psact
      machineLabelSelector:
        matchLabels:
          rke.cattle.io/control-plane-role: "true" (edited)  
 ``` 
 - Re-edit a field on the cluster object to trigger cluster reconciliation again. 
 - A new `fileSource` gets added to the cluster object with duplicate values on every cluster update. 

### Expected Result
[...]

PR #1427, opened 2024-10-17
default dynamic to false

kinarashah:
## Issue: 
https://github.com/rancher/terraform-provider-rancher2/issues/1426
 
## Solution
I removed default `true`for the `dynamic` field to align with the default API settings, which also has the added benefit of resolving the issue of duplicate entries being added.
 
### Manual Testing
- Set up an RKE2 cluster following the steps in the linked issue. Ran multiple updates on the cluster to confirm that no duplicate entries were added. 
- Changed the default PSACT from `rancher-restricted` to `rancher-privileged` and made sure it worked without causing the cluster to error out with the message `secret xxxx-psact does not contain the expected content`.


jiaqiluo (approved):

matttrach (approved):

matttrach:
looks like this needs to be backported to v5 to target Rancher v1.9

matttrach:
@kinarashah please sign your commits and this is good to merge.

PR #1439, opened 2024-11-08
Add OutBound Type to aks_config_v2

mjura:
Issue: https://github.com/rancher/aks-operator/issues/721


rancher2/schema_cluster_aks_config_v2.go
```diff
@@ -2,13 +2,18 @@ package rancher2
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 const (
 	clusterAKSV2Kind   = "aksV2"
 	clusterDriverAKSV2 = "AKS"
 )
 
+var (
+	clusterAKSOutboundType = []string{"loadBalancer", "managedNATGateway", "userAssignedNATGateway", "userDefinedRouting"}
```
yiannistri: Optional suggestion: Not sure if these are case sensitive (probably not) but maybe we should match these to how they appear in the [Azure docs](https://learn.microsoft.com/en-us/azure/aks/egress-outboundtype#outbound-types-in-aks) i.e. `managedNatGateway` and `userAssignedNatGateway`.

rancher2/schema_cluster_aks_config_v2.go
```diff
@@ -263,6 +268,13 @@ func clusterAKSConfigV2Fields() map[string]*schema.Schema {
 			Computed:    true,
 			Description: "The AKS node resource group name",
 		},
+		"outboung_type": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Default:      "loadBalancer",
+			Description:  "The AKS outbound type for the egress traffic",
+			ValidateFunc: validation.StringInSlice(clusterAKSOutboundType, true),
```
yiannistri, line 276: TIL about `validation.StringInSlice` :+1:

yiannistri (approved):

rancher2/schema_cluster_aks_config_v2.go
mjura: validation function has already non case sensitive flag enabled

rancher2/schema_cluster_aks_config_v2.go
mjura: it is `non case sensitive`

matttrach (approved):

PR #1445, opened 2024-11-15
Remove telemetry field from rancher2_bootstrap resource

adamkpickering:
This PR removes the `telemetry` field from the `rancher2_bootstrap` resource. It also removes related code and documentation. This is part of the work associated with SURE-9295, which is about removing legacy telemetry from Rancher Manager. Please see that issue for related PRs in other repositories.

Looking at https://github.com/rancher/terraform-provider-rancher2?tab=readme-ov-file#branching-the-provider, this PR should only be merged into `release/v7` because legacy telemetry will be removed in Rancher Manager 2.11. I don't think `release/v7` exists yet, so I have added the `version/rancher-v2.11` label to this PR.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -45,7 +45,6 @@ resource "rancher2_bootstrap" "admin" {
   provider = "rancher2.bootstrap"
 
   password = "blahblah"
-  telemetry = true
 }
 
 # Provider config for admin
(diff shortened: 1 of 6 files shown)


rohitsakala (approved):

adamkpickering:
The `v6.0.0` tag and the `release/v6` branch have been created. Merging this PR for inclusion into `v7.0.0` and `release/v7` when they are eventually created.

Issue #1193, opened 2023-08-07
[Feature] Enable optional OS Selection


frank-at-suse:
### Is your feature request related to a problem? Please describe.
When creating an RKE2 cluster declaratively, a type of `machineOS` can be set:

```yaml
machinePools:
  - name: pool1
    etcdRole: false
    controlPlaneRole: false
    workerRole: true
    hostnamePrefix: ''
    quantity: 1
    unhealthyNodeTimeout: 0m
    machineConfigRef:
      kind: VmwarevsphereConfig
      name: nc-undefined-pool1-sv92k
    machineOS: windows
```

This is not currently available via rancher2 Terraform provider, causing new Windows cluster deployments to fail and additions of Windows machine pools to existing clusters not possible.

### Describe the solution you'd like
Enable `machineOS` as an optional machine_pool schema to allow creation of Windows Clusters/node pools.  There should also be a check/error handling to ensure that Windows pools can only be assigned the `worker` role.

### Describe alternatives you've considered
Manual registration of Windows worker nodes

burthawth:
I am just running into this would be good to get this enhancement in place so I can build a cluster with windows nodes.

matttrach:
Hello, thanks for the suggestion! sorry for the late response, is this still relevant?

Issue #1090, opened 2023-03-23
[BUG] Unable to deploy a machine pool of Windows nodes, only linux is available


santimar:
### Rancher Server Setup
- Rancher version: 2.7.1

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? 1.25.0

### Describe the bug
I am creating a RKE2 cluster using rancher on a vsphere server and there is no way to set the OS of a given machine pool.

I am creating a machine pool like
```
resource "rancher2_machine_config_v2" "template-windows" {
  generate_name = "template-vm-windows"
  vsphere_config {
    clone_from    = var.vm_template_windows
    cpu_count     = var.cpu_count
    creation_type = "template"
    datacenter    = var.datacenter
    datastore     = var.datastore
    disk_size     = var.disk_size
    folder        = var.folder
    memory_size   = 4 * 1024
    network       = [var.network]
    pool          = var.vcenter-pool
    cfgparam      = []
  }
}
```
then I let rancher create a RKE2 cluster using the `rancher2_cluster_v2` resource
```
resource "rancher2_cluster_v2" "cluster" {
  kubernetes_version           = "v1.24.9+rke2r2"
  name                         = "cluster"
  cloud_credential_secret_name = rancher2_cloud_credential.vsphere_credential.id
  rke_config {
    machine_selector_config {
      config = {
        cloud-provider-name = "rancher-vsphere"
      }
    }
    machine_global_config = <<EOF
cni: "calico"
disable-kube-proxy: false
etcd-expose-metrics: false
tls-san:
  - my-kubernetes-domain.com
EOF
    machine_pools {
      cloud_credential_secret_name = rancher2_cloud_credential.vsphere_credential.id
      name                         = "pool-control-plain"
      ...
    }
    machine_pools {
      cloud_credential_secret_name = rancher2_cloud_credential.vsphere_credential.id
      name                         = "pool-workers"
      ...
    }
    machine_pools {
      cloud_credential_secret_name = rancher2_cloud_credential.vsphere_credential.id
      name                         = "pool-windows"
      control_plane_role           = false
      etcd_role                    = false
      worker_role                  = true
      quantity                     = 1
      machine_config {
        kind = rancher2_machine_config_v2.template-windows.kind
        name = rancher2_machine_config_v2.template-windows.name
      }
    }

    chart_values = <<EOF
rke2-calico: {}
rancher-vsphere-cpi:
  ...
rancher-vsphere-csi:
[...]

PR #1447, opened 2024-11-22
Enable provisioning of Windows nodes using vSphere templates

Ihor-kolomiiets:
This PR adds support for provisioning Windows nodes using vSphere templates in the Rancher2 provider. It enables users to create and manage Windows-based nodes using the same vSphere template logic as for Linux nodes.

## Changes
- Modified the Rancher2 provider to support Windows machine pool in cluster_v2
- Updated the provider logic to handle vSphere Windows template.
- Added new string into tests to ensure proper functionality for Windows nodes.

## Related Issues
- Closes [https://github.com/rancher/terraform-provider-rancher2/issues/1193]
- Closes [https://github.com/rancher/terraform-provider-rancher2/issues/1090]


adonskoy:
@matttrach @jiaqiluo can someone review this? This is really lacking feature in the provider

matttrach:
@Ihor-kolomiiets thank you so much for your contribution!!
This looks good from the Terraform side of things, but I am getting confirmation from the other teams to make sure it makes sense from the Rancher API and vSphere side of things. In the mean time would you mind signing your commits?
We require all commits signed by a verified key. If we don't hear from you we will probably need to resubmit this under a different account, just letting you know.

docs/resources/node_template.md
```diff
@@ -436,6 +436,7 @@ The following attributes are exported:
 * `network` - (Optional) vSphere network where the docker VM will be attached (list)
 * `password` - (Optional/Sensitive) vSphere password. Mandatory on Rancher v2.0.x and v2.1.x. Use `rancher2_cloud_credential` from Rancher v2.2.x (string)
 * `pool` - (Optional) vSphere resource pool for docker VM (string)
+* `os` - (Optional) Type of virtual machine OS in vSphere. Default `linux`(string)
```
HarrisonWAffel: Are you sure we need to update the node template resource? Isn't that RKE1 specific? Currently we only support rke2 Windows workers, rke1 support has been EOL for a few years

HarrisonWAffel (commented):
generally lgtm, thanks for the contribution! I'm going to test this locally on Monday and give a final review at that point. Only a single question for the time being

docs/resources/node_template.md
Ihor-kolomiiets: cluster and machine_config in needed, i think i just added it here as well so every resource would have ability to set OS to Windows

Ihor-kolomiiets:
> @Ihor-kolomiiets thank you so much for your contribution!! This looks good from the Terraform side of things, but I am getting confirmation from the other teams to make sure it makes sense from the Rancher API and vSphere side of things. In the mean time would you mind signing your commits? We require all commits signed by a verified key. If we don't hear from you we will probably need to resubmit this under a different account, just letting you know.

i resubmited my commit under one name, with GPG key, because one of the commit was with working mail

matttrach:
@Ihor-kolomiiets please update your branch with the latest changes from master?

HarrisonWAffel:
@Ihor-kolomiiets  thanks for raising this PR1

Also, thanks for your patience while I test this PR locally. I was able to use these changes to provision a Windows vSphere node driver cluster in our test environment. I think this LGTM, but I have two comments

1. It seems the unit tests added in this PR are failing, could you take a look? 
2. I'm still not convinced that we need to add the `os` field to the node template resource, as that resource is specific to rke1 and rke1 Windows nodes have not been supported in over a year. This isn't a hill I'm willing to die on though, and since rke1 is EOL in rancher 2.12 and we should probably clean up a few of the rke1 resources in the provider anyway.

Ihor-kolomiiets:
> @Ihor-kolomiiets thanks for raising this PR1
> 
> Also, thanks for your patience while I test this PR locally. I was able to use these changes to provision a Windows vSphere node driver cluster in our test environment. I think this LGTM, but I have two comments
> 
> 1. It seems the unit tests added in this PR are failing, could you take a look?
> 2. I'm still not convinced that we need to add the `os` field to the node template resource, as that resource is specific to rke1 and rke1 Windows nodes have not been supported in over a year. This isn't a hill I'm willing to die on though, and since rke1 is EOL in rancher 2.12 and we should probably clean up a few of the rke1 resources in the provider anyway.

I have updated tests, and locally it passes for me, can't run it here to test

also i removed `os`  field on node template, since it's EOL

HarrisonWAffel (approved):

matttrach (approved):

PR #1458, opened 2025-01-08
Add support for group_membership_filter in azuread

bigkevmcd:
## Problem

It's not currently possible to configure the user-group filter from Terraform.
 
## Solution

This adds support for configuring the `group_memembership_filter` for Azure which configures a filter for querying groups for a user.

### Manual Testing

Configured a simple "azuread" resource...

```hcl
resource "rancher2_auth_config_azuread" "azuread" {
 // other fields omitted for simplicity
  group_membership_filter = "startswith(displayName,'admin')"
}
```

And confirmed that these are configured on the azuread `AuthConfig` resource, and that they are updated when changed.

### Automated Testing

Tests were added for the new behaviour.

### DIFF
--- a/rancher2/00_provider_test.go
+++ b/rancher2/00_provider_test.go
@@ -35,7 +35,7 @@ func init() {
 	testAccRancher2AdminPass = testAccRancher2DefaultAdminPass
 	err := testAccCheck()
 	if err != nil {
-		log.Fatalf("%v", err)
+		log.Fatalf("failed check %s", err)
 	}
 }
 
@@ -64,14 +64,8 @@ func testAccCheck() error {
 		secretKey := os.Getenv("RANCHER_SECRET_KEY")
 		caCerts := os.Getenv("RANCHER_CA_CERTS")
 		adminPass := os.Getenv("RANCHER_ADMIN_PASS")
-		insecure := false
-		if os.Getenv("RANCHER_INSECURE") == "true" {
-			insecure = true
-		}
-		bootstrap := false
-		if os.Getenv("RANCHER_BOOTSTRAP") == "true" {
-			bootstrap = true
-		}
+		insecure := os.Getenv("RANCHER_INSECURE") == "true"
+		bootstrap := os.Getenv("RANCHER_BOOTSTRAP") == "true"
 
 		if apiURL == "" {
 			return fmt.Errorf("RANCHER_URL must be set for acceptance tests")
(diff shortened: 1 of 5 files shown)


bigkevmcd:
There's an issue here, but it's not just this addition that's affected.

You can't clear the value once you've set it, you can change it, but not clear it.

There are a number of bugs that cause this...

1. The `AzureADConfig` field `GroupMembershipFilter` is declared with `omitempty` which means that if you set it to an empty string, it won't actually be marshalled.
2. Our code generation adds `omitempty` to every struct, field (it should really only be struct pointers that get omitempty) even if the field isn't declared as `omitempty` (I've tried stripping `omitempty` from the field, and the generated version gets `omitempty`), I have tried manually modifying the generated code to strip `omitempty` and this **does** fix the issue.
3. We can't calculate a patch either, because we don't support PATCHing the `AuthConfigs`
  > Error: [ERROR] Updating Auth Config azuread: Bad response statusCode [405]. Status [405 Method Not Allowed]. Body: [baseType=error, code=MethodNotAllow, message=Method PATCH not supported] from [https://rancher.bigkevmcd.com:8443/v3/azureADConfigs/azuread]

To be clear, this impacts on the other fields in the AzureAD config, you can't clear the _optional_ fields used for custom endpoints either:

>    The following problems may be the cause of any confusing errors from downstream operations:
      - .auth_endpoint: was cty.StringVal(""), but now cty.StringVal("https://example.com")

enrichman (approved):
LGTM :heavy_check_mark:

rancher2/00_provider_test.go
```diff
@@ -35,7 +35,7 @@ func init() {
 	testAccRancher2AdminPass = testAccRancher2DefaultAdminPass
 	err := testAccCheck()
 	if err != nil {
-		log.Fatalf("%v", err)
+		log.Fatalf("failed check %s", err)
```
alegrey91, line 38: wdyt of using `%w` instead o `%s`?

bigkevmcd: That won't work?

https://go.dev/play/p/kOQqlRqoAIM

alegrey91: ops, sorry, my mistake

alegrey91 (approved):
LGTM

matttrach:
This is targeting Rancher v2.11 according to the milestone set in the Rancher issue https://github.com/rancher/rancher/issues/46209

matttrach:
Since the target release is Rancher v2.11, this should not be backported.

PR #1467, opened 2025-02-03
fix: remove handling of auto-install objects

### LINKED ISSUE #1195: [RFE] Don't automatically uninstall apps declared via `catalog.cattle.io/auto-install`
### Is your feature request related to a problem? Please describe.

Currently, on uninstall of app `A`, terraform automatically uninstalls also whatever app was declared in app `A` via `catalog.cattle.io/auto-install` ([this code](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_app_v2.go#L184-L196)).

This behaviour is particular to the terraform provider, and is not followed by Rancher UI.

IMHO this behaviour is dangerous:
- You can have 2 separate charts both declare the same `catalog.cattle.io/auto-install: my-crds=1.0.0`. On uninstall of any of those chart releases, `my-crds` chart gets removed, breaking the remaining chart release in the system as it is now missing the needed CRDs. This is the case for `kubewarden-controller` and `kubewarden-defaults`, as both depend on `kubewarden-crds`. See issue raised by [community here](https://github.com/kubewarden/helm-charts/issues/272).
- You can have several charts declare `catalog.cattle.io/auto-install: rancher-monitoring`. On uninstall of one of the chart release, the remaining charts are now broken.

### Describe the solution you'd like

Make the terraform behaviour on `catalog.cattle.io/auto-install` optional and skippable on uninstall of chart `A`.

Maybe it it could be better improved if it would automatically skip a chart `B` specificed via `catalog.cattle.io/auto-install` , if that chart `B` has a `provides-gvr` for some CRDs that the chart `A` has in a `requires-gvr`. But this maybe doesn't fully solve transitive dependencies, and charts that provide or depend on several gvrs, as one cannot list several on those annotations. But this needs some consideration.

For the current behaviour (on `catalog.cattle.io/auto-install`)  in Terraform to work, it would be needed to have a proper dependency management system for Helm charts and/or OCI artifacts. This dependency system would know about dependencies, optional dependencies, their versions, if they are namespaced, and to delegate to it if a chart release is not needed in the system anymore or can be scheduled for removal.

### Describe alternatives you've considered

- Forking charts, so one can overwrite the `catalog.cattle.io/auto-install` annotations, effectively disabling the stated dependency and managing those dependencies manually.

- Forking the terraform provider and commenting [this code](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_app_v2.go#L184-L196).

- Adding a dependency system that is aware of rancher annotations and can be used as a drop-in substitute of Helm binary and library: https://hypper.io (disclaimer, helped write that thing).

### Additional context

See https://github.com/kubewarden/helm-charts/issues/272

matttrach:
Updating the delete app function to more closely align with the API. Terraform shouldn't try to find and delete auto-install objects when deleting an app. The API should manage cleaning up child objects if it is necessary to do so. In this case Terraform was removing CRDs that multiple apps depend on, causing app removal to break running apps. 

This addresses #1195

### DIFF
--- a/rancher2/resource_rancher2_app_v2.go
+++ b/rancher2/resource_rancher2_app_v2.go
@@ -180,41 +180,7 @@ func resourceRancher2AppV2Delete(d *schema.ResourceData, meta interface{}) error
 	if waitErr != nil {
 		return fmt.Errorf("[ERROR] waiting for app (%s) to be deleted: %s", app.ID, waitErr)
 	}
-	if app.Spec.Chart.Metadata != nil && app.Spec.Chart.Metadata.Annotations != nil && len(app.Spec.Chart.Metadata.Annotations) > 0 && len(app.Spec.Chart.Metadata.Annotations["catalog.cattle.io/auto-install"]) > 0 {
-		namespace := d.Get("namespace").(string)
-		if len(app.Spec.Chart.Metadata.Annotations["catalog.cattle.io/namespace"]) > 0 {
-			namespace = app.Spec.Chart.Metadata.Annotations["catalog.cattle.io/namespace"]
-		}
-		chartAuto := splitBySep(app.Spec.Chart.Metadata.Annotations["catalog.cattle.io/auto-install"], "=")
-		if len(chartAuto) != 2 {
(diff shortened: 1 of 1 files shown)


rohitsakala (approved):
I don't know much of terraform, but logic wise looks good to me.

HarrisonWAffel (approved):

PR #1485, opened 2025-03-04
(feat): Cluster Agent PC and PDB support

HarrisonWAffel:
## Problem
Rancher v2.11.0 introduces a new optional feature for configuring a Pod Disruption Budget and Priority Class for the cattle cluster agent. This includes new fields on the `v3.Cluster` and `v1.Cluster` objects. This feature is supported in Node Driver provisioned clusters, custom clusters, and imported clusters. 
 
## Solution

Update the terraform schema to include these new fields, add flatteners and expanders, unit tests, and documentation. Right now this feature is only supported for the cattle cluster agent, and not the fleet agent. The schema has been updated to only allow for these fields to be set when customizing the cattle cluster agent, but can later be changed to support both agents. 

This PR also removes logic for catalog v1, as the latest versions of Rancher no longer include the related types (see https://github.com/rancher/rancher/pull/47661). 
 
## Testing
I've manually tested rke2 node driver, custom, and imported clusters. 

### Automated Testing
Several new unit tests have been written and existing ones updated

n/a


go.mod
```diff
@@ -1,71 +1,76 @@
...
+	golang.org/x/crypto v0.33.0
+	golang.org/x/sync v0.11.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.30.2
-	k8s.io/apimachinery v0.30.2
-	k8s.io/apiserver v0.30.1
+	k8s.io/api v0.32.2
+	k8s.io/apimachinery v0.32.2
+	k8s.io/apiserver v0.32.2
 	k8s.io/client-go v12.0.0+incompatible
```
jiaqiluo, line 22: Interestingly we are using such an old version of the package, can you give it a try to bump it to v0.32.2 ?

jiaqiluo (approved):

go.mod
HarrisonWAffel: `k8s.io/client-go` is later replaced with `k8s.io/client-go v0.32.2` at the bottom of the file. I'm unsure why a replace statement is used, but I kept it as is in case there is a functional reason for this.

jiaqiluo: We usually put the replace directive at the top of the file, but that is not the case here, I guess that is why I missed it. :)

thatmidwesterncoder (approved):

HarrisonWAffel:
Latest force push contains [no new changes](https://github.com/rancher/terraform-provider-rancher2/compare/c90161a..25c334f) and simply adds proper verification to existing commits

Issue #1495, opened 2025-04-17
[BUG] rancher_cluster_v2 machine config max_surge and max_unavailable fails when using absolute values


pvlkov:
### Rancher Server Setup
- Rancher version: v2.10.2
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): v1.30.10+rke2r1

### Information about the Cluster
- Kubernetes version: v1.30.10+rke2r1
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Self-Hosted VMWare vSphere, provisioned via Rancher

### User Information
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom): Admin

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use?
- What is the version of Terraform in use?
```
Terraform v1.11.3
on linux_amd64
+ provider registry.terraform.io/rancher/rancher2 v6.1.4
```

### Describe the bug

On a RKE2 cluster provisioned via Rancher using the rancher2_cluster_v2 resource, when setting max_surge and max_unavailable options on machine pools to an absolute value instead of a percentage the ClusterAPI validating webhook fails with the following error: 
```
admission webhook "validation.machinedeployment.cluster.x-k8s.io" denied the request: MachineDeployment.cluster.x-k8s.io "xxxx" is invalid: [spec.strategy.rollingUpdate.maxSurge: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:"1"}: must be either an int or a percentage: invalid value for IntOrString: invalid type: string is not a percentage, spec.strategy.rollingUpdate.maxUnavailable: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:"1"}: must be either an int or a percentage: invalid value for IntOrString: invalid type: string is not a percentage]
```

This is because these values are converted to intstr using intstr.FromString(v) [here](https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_v2_rke_config_machine_pool.go#L135).

intstr.FromString always generates an Intstr with "Type: 1" and empty "IntVal" fields ([intstr documentation](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/intstr#FromString)). 

[...]

PR #1496, opened 2025-04-17
Replace intstr.FromString with intstr.Parse to allow for int values

pvlkov:
## Issue: 

Solves https://github.com/rancher/terraform-provider-rancher2/issues/1495

## Problem

Specifying max_surge and max_unavailable as absolute values on a Rancher-provisioned CAPI cluster's node-pool gets rejected by the Machinedeployment validating webhook.
 
## Solution

As described in the issue, this is due to the use of intstr.FromString(v) which only ever produces intstr with Type: 1 and empty IntVal fields.

intstr.Parse should be used instead, which tries to call intstr.FromInt32 first and falls back to intstr.FromString if that fails.
 
### Manual Testing

Built custom TF provider binary and used it via developer override to apply to an existing cluster. 
Terraform output:
```
# rancher2_cluster_v2.cluster will be updated in-place
~ resource "rancher2_cluster_v2" "cluster" {
         id                         = "fleet-default/testcluster"
         name                       = "testcluster"
        # (9 unchanged attributes hidden)
       ~ rke_config {
             # (3 unchanged attributes hidden)
          ~ machine_pools {
                 name                           = "workers"
                 # (16 unchanged attributes hidden)
               ~ rolling_update {
                   ~ max_surge       = "1" -> "3"
                  ~ max_unavailable = "1" -> "3"
                 }
                 # (1 unchanged block hidden)
             }
             # (5 unchanged blocks hidden)
        }
         # (3 unchanged blocks hidden)
     }

```

Resulting MachineDeployment resource:
```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachineDeployment
...
  name: xxxx
  namespace: fleet-default
  ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1beta1
      kind: Cluster
      name: xxx
spec:
 ...
  strategy:
    rollingUpdate:
      deletePolicy: Oldest
      maxSurge: 3
      maxUnavailable: 3
    type: RollingUpdate
```

Before the change, the MachineDeployment would not be changed when applying this kind of configuration because the validating webhook would reject it.

## QA Testing Considerations

IMO this should not change any existing behavior, since values provided as percentages work as before and values provided as absolute values were broken anyway.

The same safeguards in the Cluster API validating webhook are still in place and should catch any other misconfigurations as before.
 
None that I can see...

### DIFF
--- a/rancher2/structure_cluster_v2_rke_config_machine_pool.go
+++ b/rancher2/structure_cluster_v2_rke_config_machine_pool.go
@@ -145,11 +145,11 @@ func expandClusterV2RKEConfigMachinePoolRollingUpdate(p []interface{}) *provisio
 	in := p[0].(map[string]interface{})
 
 	if v, ok := in["max_surge"].(string); ok && len(v) > 0 {
-		maxSurge := intstr.FromString(v)
+		maxSurge := intstr.Parse(v)
 		obj.MaxSurge = &maxSurge
 	}
 	if v, ok := in["max_unavailable"].(string); ok && len(v) > 0 {
-		maxUnavailable := intstr.FromString(v)
+		maxUnavailable := intstr.Parse(v)
 		obj.MaxUnavailable = &maxUnavailable
 	}
 


snasovich (approved):
Changes look sane to me and should allow accommodating both absolute values (like `1`) and percentages (like `10%`).

jiaqiluo (approved):
LGTM.

pvlkov:
Rebased branch to upstream/master and signed the commit

PR #1497, opened 2025-04-23
Add test framework

matttrach:
## Problem
Current acceptance test framework doesn't work. 

## Solution
Introduce new testing framework using Terratest and manage testing dependencies with Nix.

## Testing
Install Nix https://nixos.org/download/
Add AWS credentials to environment. 
- there are several ways to do this, in my environment I have an rc file saved at `~/.config/aws/default` with 
```
#!/bin/env sh
export AWS_ACCESS_KEY_ID='ABC123'
export AWS_SECRET_ACCESS_KEY='abc123'
export AWS_REGION='us-west-2'
export AWS_DEFAULT_REGION='us-west-2'
export ZONE='abc.123'
```

source .envrc
./scripts/run_tests.sh

### Manual Testing
I manually deployed examples using Terraform.

### Automated Testing
I use the new scripts/run_test.sh to automatically test and deploy examples.

## QA Testing Considerations
This doesn't change any of the provider's behaviors, hopefully it will help reduce QA burden.
 It would be nice if there are use cases that QA can describe to please create an issue and I will add an example for that use case to be automatically tested.


test/base/base_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	aws "github.com/gruntwork-io/terratest/modules/aws"
+	g "github.com/gruntwork-io/terratest/modules/git"
+	"github.com/gruntwork-io/terratest/modules/ssh"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	util "github.com/rancher/terraform-provider-rancher2/test"
+)
+
+// This test makes sure we can deploy Rancher on AWS
+func TestBase(t *testing.T) {
```
HarrisonWAffel, line 16: This function has some indentation problems, looks like a mix of tabs and spaces

test/util.go
```diff
@@ -0,0 +1,433 @@
...
+	return id
+}
+
+func CreateTestDirectories(t *testing.T, id string) error {
+	gwd := g.GetRepoRoot(t)
+	fwd, err := filepath.Abs(gwd)
+	if err != nil {
+		return err
+	}
+	tdd := fwd + "/test/data"
```
HarrisonWAffel: Small nit, but you could use `filepath.Join` to more cleanly build the full paths, and could place this in a loop, which would make it easier to read.

e.g. 

```go
paths := []string{
 filepath.Join(fwd, "test/data"),
 filepath.Join(fwd, "test/data", id),
 filepath.Join(fwd, "test/data", id, "providers"),
 filepath.Join(fwd, "test/data", id, "plugins"),
} 

for _, path := range paths { 
  	err = os.Mkdir(path, 0755)
	if err != nil && !os.IsExist(err) {
		return err
	}
}

File: test/util.go [hunk 3]
```diff
@@ -0,0 +1,433 @@
...
+}
+
+func GetAwsSessionToken() string {
+	return os.Getenv("AWS_SESSION_TOKEN")
+}
+
+func GetBuild() bool {
+	if os.Getenv("SKIP_BUILD") == "true" {
+		return false
+	} else {
```
HarrisonWAffel: Small nit: You could just return `os.Getenv("SKIP_BUILD") != "true"`

scripts/run_tests.sh
```diff
@@ -0,0 +1,190 @@
...
+  IDENTIFIER="$(echo a-$RANDOM-d | base64 | tr -d '=')"
+  export IDENTIFIER
+fi
+echo "id is: $IDENTIFIER..."
+if [ -z "$GITHUB_TOKEN" ]; then echo "GITHUB_TOKEN isn't set"; else echo "GITHUB_TOKEN is set"; fi
+if [ -z "$GITHUB_OWNER" ]; then echo "GITHUB_OWNER isn't set"; else echo "GITHUB_OWNER is set"; fi
+if [ -z "$ZONE" ]; then echo "ZONE isn't set"; else echo "ZONE is set"; fi
+echo 'if tmp directory is missing, try restarting dev environment'
+
+#if [ -z "" ]; then echo ""; else echo ""; fi
```
HarrisonWAffel: Is this comment intentional?

test/util.go
```diff
@@ -0,0 +1,434 @@
...
+	if len(versions) > 1 {
+		stable = versions[1]
+	}
+	if len(versions) > 2 {
+		lts = versions[2]
+	}
+	return latest, stable, lts, nil
+}
+
+func getRancherReleases() ([]*github.RepositoryRelease, error) {
```
HarrisonWAffel: You could combine `getRancherReleases` and `getRke2Releases` into a single `getReleases` function, and then pass the organization and repo as parameters. Right now those two functions are nearly identical

examples/downstream/outputs.tf
```diff
@@ -0,0 +1,22 @@
...
+  description = <<-EOT
+    The kubeconfig for the server.
+  EOT
+  sensitive   = true
+}
+output "address" {
+  value = module.rancher.address
+}
+# output "admin_token" {
+#   value     = module.rancher.admin_token
```
HarrisonWAffel: Do these comments need to be removed?

HarrisonWAffel (commented):
Still wrapping my head around the full PR, but I have a few comments (mostly just nits)

examples/downstream/outputs.tf
matttrach: removed, thank you!

test/util.go
matttrach: Done, that took a bit of work, but I think the refactor looks a lot better, thank you!

scripts/run_tests.sh
matttrach: nope, sorry

test/util.go
matttrach: that is much more concise, thank you!

test/util.go
matttrach: You are right, plus it prevents me from needing a hundred error check statements, I am using this.

test/base/base_test.go
matttrach: Yeah, something is going on with my code editor, it should be running gofmt on save, but that doesn't appear to be happening. I think I got it now.

examples/downstream/outputs.tf
HarrisonWAffel: Looks like they may have been added back accidentally

HarrisonWAffel (commented):
Just one last comment. Over all this looks really cool!

examples/downstream/outputs.tf
matttrach: oops, fixed

HarrisonWAffel (approved):

jiaqiluo (approved):
Changes look good to me. Good job!

PR #1515, opened 2025-05-23
fix: name space imports should not change project

matttrach:
## Problem

Importing a namespace to your terraform state should never result in a change to that namespace.

## Solution

Remove the code that moves the namespace between projects if a different project is specified.
Instead, return an error specifying that the project specified doesn't match the actual resource.
 
## Testing

The majority of the work in this PR is in the testing.
Testing terraform import can be a bit difficult, it requires orchestrating multiple terraform states. I chose to orchestrate terraform with terraform. In the example I generate a Rancher node, then a downstream cluster, and finally a namespace.
I created a local module named "deploy" which will deploy a Terraform module in its own state. This can actually be pretty helpful if you need to break down complex or large builds into separate states. The deploy module deploys a local module named "import" which imports the namespace and the downstream cluster. This verifies that the namespace can still be imported even with the removal of the project change.

I tested this using the included run_tests.sh in the ./scripts directory. `./scripts/run_tests.sh -t TestDownstreamImport`.
The script will build the provider locally and ensure that Terraform uses it when executing the import example, it does this with a .terraformrc that the script builds on the fly. The run_tests script assumes you have some dependencies and environment variables set, which are found in the flake.nix and .envrc files in the repo.

A big part of the run_tests.sh script is the amount of work I put into making sure that it can clean up if there are any problems with the build. Using an opensource tool called "leftovers" which is designed exactly for this purpose I am able to find and remove any objects in AWS which may be left over after a failed test run.

## QA Testing Considerations

I am not sure why the project is linked to the namespace so tightly. I deployed a rancher_cluster_v2, but had to get the project_id from a data_rancher_cluster resource in order to import the namespace.
Testing Terraform imports automatically is tricky, and there are some weird things I did to avoid dynamically writing Terraform code to validate the error. Instead of adding dynamic resources or overrides, I added a variable "project_mismatch" which, when set to true, will trigger the failure path. I expect the error in Terratest and fail the test if the error isn't returned.


rancher2/import_rancher2_namespace.go
matttrach, line 1: The most relevant change is here

docs/resources/namespace.md
matttrach, line 1: Here are the docs that go along with the change.

.github/workflows/release.yaml
```diff
@@ -44,8 +44,8 @@ jobs:
           echo '${{ env.GPG_PASSPHRASE }}' | gpg --detach-sig --pinentry-mode loopback --passphrase-fd 0 --output "${SHASUM_FILE}.sig" --sign "${SHASUM_FILE}"
 
           echo "Validating signature..."
-          gpg --verify "${SHASUM_FILE}.sig" "${SHASUM_FILE}"
-          if [ $? -eq 0 ]; then
+
+          if ! gpg --verify "${SHASUM_FILE}.sig" "${SHASUM_FILE}"; then
```
matttrach, line 48: This was a shellcheck suggestion.

rancher2/import_rancher2_namespace.go
```diff
@@ -1,39 +1,34 @@
...
+	log.Printf("[INFO] Splitting given id %s", d.Id())
 	projectID, resourceID := splitID(d.Id())
-
+	log.Printf("[INFO] Splitting project %s", projectID)
 	clusterID, projectID := splitProjectID(projectID)
+	log.Printf("[INFO] Using cluster id %s", clusterID)
 
 	client, err := meta.(*Config).ClusterClient(clusterID)
 	if err != nil {
+		log.Printf("[INFO] Problem getting cluster client for cluster with id \"%s\"", clusterID)
```
jiaqiluo, line 22: nit
```suggestion
		log.Printf("[ERROR] Problem getting cluster client for cluster with id \"%s\"", clusterID)
```

jiaqiluo (approved):
LGTM! Just a minor nit, but I’ll go ahead and approve so it doesn’t block merging.

jiaqiluo:
I'd suggest making an issue to track the changes and linking it to the PR if there is not already one. 
This can be considered a breaking change, so it should be called out in the release note.

HarrisonWAffel (approved):
Nice! lgtm. I agree with Jiaqi though, we might want to make an issue to make this change more visible to others when updating their provider versions.

matttrach:
https://github.com/rancher/terraform-provider-rancher2/issues/1516
Thank you!

PR #1517, opened 2025-05-28
feat: Implement support for GCE machine configs

HarrisonWAffel:
## Problem

Rancher v2.12.0 will add support for a Google Compute Engine node driver, however the associated machine config is not supported in the provider.
 
## Solution

Add support for the GCE machine config, allowing users to provision GCE clusters using the terraform provider.
 
## Testing
I've tested these changes against a Rancher version which allows for the provisioning of GCE clusters. I was able to successfully provision and destroy a simple GCE cluster which exposes ports publicly, creates an internal network, and applies custom labels and tags.

### Manual Testing
I've done the above 

### Automated Testing
n/a

## QA Testing Considerations

Some fields shouldn't be set when testing 
+ `auth_encoded_json` should be omitted, and instead cloud credentials should be used to provide this information. 
+ `use_existing` should be omitted, as Rancher node drivers do not support importing existing nodes 

While these fields should not be used, I added them to the config so that it fully represents the actual machine config CR. I'm open to removing them, but similar fields exist in other machine configs offered by the provider and I wanted to follow the established pattern. 
 
n/a, as this PR follows the established pattern of other machine configs in the provider.


matttrach:
Please make sure that your commit messages fit this convention before merge:
https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits

HarrisonWAffel:
Thanks for the link! Both the title and commit should be correct now.

matttrach (approved):
This looks good to me

matttrach:
created #1533 to track progress

Issue #1499, opened 2025-04-29
fix: [RFE] [BUG] GKE - add Node Pools to existing clusters


palexus:
### Is your feature request related to a problem? Please describe.
In the rancher2_cluster resource, there is the gke_config_v2 block which supports a list of node_pools. In creation, it can take several node pools, but for existing gke clusters, adding a node_pool triggers the recreation of the entire cluster, which is not desirable. In the Rancher UI or with the google provider (by not using the rancher2 provider) it is possible.

When modifying the node_pools list—such as by adding a new node pool—the provider treats the entire list as a single entity. Consequently, any change to this list is interpreted as a complete replacement of the existing configuration. This behavior leads Terraform to plan the destruction and recreation of the entire cluster, rather than applying the change incrementally.

This limitation is inherent in the current implementation of the Rancher2 Terraform provider. The provider lacks the granular control necessary to manage individual node pools separately within the node_pools list. As a result, even minor changes to the node pool configuration can trigger a full cluster replacement, even though it is not marked with "ForceNew". I think, this behaviour is not intended and that's why I am not sure, if it is a feature request or a Bug.

### Describe the solution you'd like
Please add this functionality. Manually managing the node pools introduces a divergence between the Terraform state and the actual infrastructure, which can complicate future management.

### Describe alternatives you've considered
Manually managing Node Pools.

### Additional context
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/structure_cluster_gke_config_v2.go

https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/schema_cluster_gke_config_v2.go

vivek-shilimkar:
Tested the bug fix on rancher v2.12-head with rancher2 provider [v8.0.0-rc1](https://github.com/rancher/terraform-provider-rancher2/releases/tag/v8.0.0-rc1). 
Manually changing the node count do not require recreation cluster, the new nodes gets added successfully. Since the issue is fixed it can be closed.

matttrach:
This has been released.

PR #1518, opened 2025-05-29
fix: GKE cluster update behavior

krunalhinguu:
## Issue  
https://github.com/rancher/terraform-provider-rancher2/issues/1499 This PR resolves an issue where **any change to a GKE cluster's `node_pool` in Terraform caused a full cluster replacement**, which was disruptive and not necessary in most cases.

---

## Problem  
In the existing implementation of the `rancher2_cluster` resource for GKE clusters, the `node_pool` block was marked with `ForceNew`, which forced a full replacement of the cluster whenever any node pool change was made. This made it impossible to perform in-place updates like scaling or updating node pool properties without recreating the entire cluster.

---

## Solution  
- **Removed `ForceNew`** from the `node_pool` field in the Terraform schema.
- **Added custom diff logic** in the update function to properly detect and apply in-place changes to `node_pool` entries.

These changes allow users to modify existing GKE node pools without triggering full cluster recreation, improving usability and minimizing downtime.

---

### Manual Testing  
- Created a GKE cluster via Rancher using Terraform.
- Modified node pool fields (e.g., machine type, labels, node count).
- Verified that Terraform plan shows in-place updates, and apply works without cluster replacement.
- Confirmed updated configuration reflected correctly in Rancher UI and GCP Console.

### Automated Testing  
- No new automated tests added for this change.
- Existing tests for `rancher2_cluster` still pass, confirming backward compatibility.

---

## QA Testing Considerations  
- Test updating an existing GKE cluster's node pool (e.g., changing node count, version, labels).
- Confirm no full cluster replacement occurs.
- Validate that updates are reflected correctly in Rancher and on GKE.
- Test both new cluster creation and update flows to confirm schema compatibility.

---

- Changes to `node_pool` update behavior could potentially cause issues if Rancher expects fields that are no longer being sent.
- Regressions are **low to medium risk**, but QA should test edge cases like:
  - Reducing node count to 0
  - Removing a node pool
  - Updating multiple properties at once

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -35,6 +35,19 @@ func resourceRancher2Cluster() *schema.Resource {
 					d.SetNew("eks_config_v2", flattenClusterEKSConfigV2(newObj, []interface{}{}))
 				}
 			}
+
+			if d.Get("driver") == clusterDriverGKEV2 && d.HasChange("gke_config_v2") {
+				old, new := d.GetChange("gke_config_v2")
+				oldObj := expandClusterGKEConfigV2(old.([]interface{}))
+				newObj := expandClusterGKEConfigV2(new.([]interface{}))
+
+				if reflect.DeepEqual(oldObj, newObj) {
+					d.Clear("gke_config_v2")
+				} else {
+					d.SetNew("gke_config_v2", flattenClusterGKEConfigV2(newObj, []interface{}{}))
+				}
+			}
+
 			return nil
 		},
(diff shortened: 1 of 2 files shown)


matttrach:
This change is currently targeting Rancher v2.12.0 which coincides with v8.0.0 of this project. Please let me know if you need this to be backported to other versions.

mjura (commented):
LGTM

matttrach (approved):

matttrach:
Please sign your commits with a valid key so that we can merge.

Issue #1523, opened 2025-05-30
[Update] Support for `importedConfig.privateRegistryURL` on Cluster resource


snasovich:
### Does the update make a breaking change?
No, want to add equivalent for `importedConfig.privateRegistryURL` to https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/cluster
Note the resource currently doesn't even have `importedConfig` there.

### Which API is changing?
`cluster.management.cattle.io/v3` -> added `importedConfig.privateRegistryURL` field

## Which team is responsible for the API?
Hostbusters

### What version of Rancher will the change be released in?
Released with 2.11.0 as part of https://github.com/rancher/rancher/pull/49088 / https://github.com/rancher/rancher/issues/35192

### How did you verify the change?
N/A

MichaelGerhart:
Rancher Hosted customer checking in. We are looking forward to this being completed so we can configure all of our imported clusters to use an internal registry via the terraform we already use to create them all.

matttrach:
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v8.0.0-rc4
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v7.3.0-rc3
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v6.7.0-rc1

Josh-Diamond:
Ticket# 1545 - Test Results - ✅ 
---

Verified with TFP Rancher2 `v8.0.0-rc4` on Rancher `v2.11.3`:

| Scenario | Test Case | Result |
|---|---|---|
| 1. | Using rancher2 provider, create an imported cluster object, with a configured private registry | ✅ |

---

Scenario 1 - ✅ 
1. Fresh install of Rancher `v2.11.3`
2. Using tfp rancher2 `v8.0.0-rc4`, run the following:
```terraform
terraform {
  required_providers {
    rancher2 = {
      source  = "terraform.local/local/rancher2"
      version = "8.0.0-rc4"
    }
  }
}

provider "rancher2" {
  api_url   = "<REDACTED>"
  token_key = "<REDACTED>"
  insecure  = true
}

resource "rancher2_cluster" "jkeslar-imported" {
  name        = "jkeslar-imported"
  imported_config {
    private_registry_url = "<REDACTED>"
  }
}
```
3. Once ran, navigate to Rancher UI and run installation command against standalone RKE2 cluster (`v1.32.5+rke2r1`)
[...]

PR #1528, opened 2025-06-06
fix: add support for imported cluster config

kinarashah:
## Issue: 
https://github.com/rancher/terraform-provider-rancher2/issues/1523
 
## Problem
Terraform currently does not support setting the `ImportedConfig` field explicitly. The assumption has been that Rancher creates an imported cluster automatically when no configuration is provided.

However, this assumption no longer holds as a new required field for supporting private registries was introduced for imported clusters. As a result, explicit support for `ImportedConfig` is now necessary.
 
## Solution
Introduced support for `imported_config` in `rancher2_cluster`. As far as I know, imported clusters are not supported with `rancher2_cluster2` since these are handled via the `v3/clusters` endpoint.

The resulting state differs slightly when `imported_config` is used so added the new state to the expected state list. However, this is consistent with the behavior observed when creating imported clusters directly through the Rancher API. The difference is not significant as the registration URL is correctly generated in both cases and the cluster becomes active after a successful import.
 
## Engineering Testing
Manual testing, created cluster using both old and new format: 
```
resource "rancher2_cluster" "foo" {
  name        = "foo"
}

resource "rancher2_cluster" "bar" {
  name        = "bar"
  imported_config {
    private_registry_url = "test102.io"
  }
}
```

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -20,7 +20,19 @@ resource "rancher2_cluster" "foo-imported" {
 }
 ```
 
-Creating Rancher v2 RKE cluster
+### Creating Rancher v2 imported cluster with custom configuration. For Rancher v2.11.x and above.
+
+```hcl
+# Create a new rancher2 imported Cluster with custom configuration 
+resource "rancher2_cluster" "foo-imported" {
+  name        = "foo-imported"
+  imported_config {
+    private_registry_url = "test.io"
+  }
+}
+```
+
+### Creating Rancher v2 RKE cluster
 
 ```hcl
 # Create auditlog policy yaml file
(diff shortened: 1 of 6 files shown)


matttrach:
Please sign your commits with a verified key.

matttrach:
What version of the provider is this targeting?
Is this expected for the v8 release correlating to Rancher's v2.12 or something earlier?

matttrach:
Config is not a valid word for the commit messages, please use "configuration".

matttrach:
or just "fix: add support for imported cluster" should be fine

kinarashah:
@matttrach Thanks for the review, pushed an updated commit message with verified key. The field was first introduced with Rancher v2.11, so it's not scoped to v2.12+ only and I'd like to backport it to v7 release line if possible.

matttrach:
@kinarashah your commit wasn't signed with a verified key yet: 
You can see in the link below that your commit is missing the "verified" badge.
https://github.com/rancher/terraform-provider-rancher2/pull/1528/commits

matttrach (approved):

matttrach:
This needs one more approval before merge, maybe @HarrisonWAffel ?

matttrach:
main issue: https://github.com/rancher/terraform-provider-rancher2/issues/1523

HarrisonWAffel (changes requested):
Core changes look good, but I think we need to update the docs with this new field

docs/resources/cluster.md
```diff
@@ -20,7 +20,19 @@ resource "rancher2_cluster" "foo-imported" {
 }
 ```
 
-Creating Rancher v2 RKE cluster
+### Creating Rancher v2 imported cluster with custom configuration. For Rancher v2.11.x and above.
+
+```hcl
+# Create a new rancher2 imported Cluster with custom configuration
```
HarrisonWAffel, line 26: This is a great example, but since this is a new top level field we should also include it in the list starting on line 618, and add an entry under the `Nested blocks` header which explains the purpose of the `private_registry_url`. 

It's a small nit, but we've been pretty good about tracking all fields (even if they're self explanatory).

HarrisonWAffel (changes requested):

docs/resources/cluster.md
kinarashah: Ah, didn't realize we manually track all fields. Added it there as well, thanks!!

HarrisonWAffel: Nice!

HarrisonWAffel (approved):

matttrach:
sorry, I missed a comma when resolving merge conflicts. I don't have access to your branch to make the change, so you will need to fix. Then I can merge.

PR #1563, opened 2025-07-02
fix: specify signing key

matttrach:
Release is blocked due to a signing error, seems like the secret key is not being selected appropriately by default.
This change specifies the key to use for signing, it relies on a new secret privateKeyId.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,9 +28,9 @@ jobs:
         uses: rancher-eio/read-vault-secrets@main
         with:
           secrets: |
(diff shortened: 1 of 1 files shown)


HarrisonWAffel (approved):

Hypertrophik (commented):
Can we update the path to `secret/data/github/repo/${{ github.repository }}/signing/gpg` ?

Hypertrophik (approved):
Once the path is changed we should be set. We will need to let Hashicorp know the key expiration was changed.

Issue #1498, opened 2025-04-24
[BUG] kube_config of rancher2_cluster_v2 not updated after ca certificate updated


boettluSICKAG:
### Rancher Server Setup
- Rancher version: v2.10.3
- Installation option (Docker install/Helm Chart): helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): v1.31.7+rke2r1
- Proxy/Cert Details: custom ca cert

### Information about the Cluster
- Kubernetes version: v1.31.7
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): vSphere

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? 7.0.0
- What is the version of Terraform in use? v1.10.5

### Describe the bug
When the certificate of the rancher server is changed, the terraform provider does not update the kube_config in the terraform state of the downstream clusters.

### To Reproduce
1. Create Downstream cluster with terraform provider
2. Change private ca certificate on the Rancher server according to [this](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/resources/update-rancher-certificate) documentation.
3. Try to create Kubernetes resources with the `rancher2_cluster_v2.example.kube_config` attribute

### Actual Result
```
kubernetes_namespace.test: Creating...
╷
│ Error: Post "https://<rancher-fqdn>/k8s/clusters/<cluster-id>/api/v1/namespaces": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

### Expected Result
The terraform provider should provide a valid kube_config with the updated certificate-authority-data

### Additional context
This seems to be related to #934 and the kubeconfig validation in the provider. Since the refresh of the kubeconfig token works fine, the logic should be extended to also refresh the certificate-authority-data.

matttrach:
The error you mentioned is from openssl, the provider doesn't cache or trust Rancher certificates. Try using curl to contact the Rancher URL, I suspect you will see the same response. Also try using kubectl to contact the cluster, this will give us a good idea of what might be causing the error.
As a workaround, if you trust network path to the cluster, the provider has an argument to ignore TLS
https://registry.terraform.io/providers/rancher/rancher2/latest/docs#insecure-1
This is not recommended for production unless you are in complete control of the network (such as a thoroughly vetted air gapped local environment).

[...]

PR #1585, opened 2025-07-07
fix: validate kubeconfig connection to apiserver

lukasboettcher:
## Problem
see #1498 
 
## Solution
During validation of the kubeconfig, use the client to check the apiserver version. This fails if the ca certificate in the kubeconfig is invalid, which then causes a new kubeconfig to requested by rancher.
This requires a tls connection from the terraform client to the apiserver, which I think is a reasonable requirement.
 
## Testing
I validated this against my own environment, where I replicated the issue.


matttrach:
@lukasboettcher please sign your commit with a valid key, if you look at the commits [here](https://github.com/rancher/terraform-provider-rancher2/pull/1585/commits) each commit should have the "verified" badge.

matttrach:
@lukasboettcher would you mind sharing a little bit of your config so that I can add an e2e test making sure this works in the future? I think the important part is how you are invalidating your downstream kubeconfig.

jiaqiluo (approved):

matttrach (approved):

lukasboettcher:
@matttrach here is a repo with minimal steps to reproduce the problem https://github.com/lukasboettcher/rancher-tf-provider-ca-bug

you need a rancher server, create a downstream cluster, trigger a change in the ca-certificate of the rancher server and then try to use the kubeconfig from the rancher downstream cluster.

PR #1587, opened 2025-07-10
fix: Clean up test suite

matttrach:
QOL updates and a test validating downstream cluster contact

QOL updates:
- renamed some examples and tests to better describe what was being done
- removed tests that don't work
- pin the Helm provider version in all of the tests/examples
- added a custom dictionary for spell checking commit messages
- upgraded all of the examples to use the new Rancher module release
- added a test to ensure that users can connect to downstream clusters directly by retrieving their kubeconfig from data "rancher2_cluster_v2"


rancher2/resource_rancher2_cluster.go
```diff
@@ -567,62 +567,10 @@ func isKubeConfigTokenValid(c *Config, config string) (string, bool, error) {
...
-			return "", fmt.Errorf("error creating Token: %s", err)
-		}
-	}
-	kubeconfig.AuthInfos[0].AuthInfo.Token = currentToken.Token
-	return getKubeConfigFromObj(kubeconfig)
-}
-
-func getClusterKubeconfig(c *Config, id, origconfig string) (*managementClient.GenerateKubeConfigOutput, error) {
+func getClusterKubeconfig(c *Config, id string) (*managementClient.GenerateKubeConfigOutput, error) {
 	action := "generateKubeconfig"
```
HarrisonWAffel: I'm not super experienced with Norman, but does this mean we will be generating a completely new kubeconfig each time the cluster read operation occurs? 

I think this may lead to an accumulation of `tokens.management.cattle.io` CRDs in the local cluster, since it looks like a new one is generated each time the action is invoked ([1](https://github.com/rancher/rancher/blob/main/pkg/api/norman/customization/cluster/actionhandler.go#L29), [2](https://github.com/rancher/rancher/blob/main/pkg/api/norman/customization/cluster/actions_kubeconfig.go#L38-L49)). When I invoke the `generateKubeconfig` action in the Rancher UI (by downloading a kubeconfig for a downstream cluster) the API tokens list gets a new entry

HarrisonWAffel (commented):
question / observation

rancher2/resource_rancher2_cluster.go
matttrach: Is there a better action that could be called? 
It sounds like what you are saying is that the UI doesn't cache this information either.
While a terraform config may call it more often than the user in the UI it doesn't seem abusive to the API's purpose.
Is there any way to solve this problem in Norman rather than here in the provider? My goal is to reduce the amount of code in this provider that isn't being used to facilitate communication with Rancher.

matttrach: ok, I have removed the commit that has this in it. After talking this over in private we are going to accept the change that the user provided in #1585.

HarrisonWAffel (commented):
Updates to the examples and CI look good, but I'm a bit concerned with removing so many tests. How are they currently broken? A lot of the tests are definitely out of date, but some of them focus on feature that we still support and should have some coverage for.

matttrach:
None of the tests run, they are based on an unknown configuration of Docker that was present in our old CI infrastructure.

HarrisonWAffel (approved):
After discussing more over Slack I agree that the acceptance tests being removed here are too out-dated to save and it would be a better use of time to port the relevant cases into the new testing framework 

LGTM!

jiaqiluo (approved):

PR #1595, opened 2025-07-21
feat(machine-config): harvester gpu support

walkafwalka:
## Problem

 
vGPUs parameter is not supported in the Harvester config for the Machine Config resource.

## Solution

Adds support for the vGPU parameter.
 
## Testing

Specify a `vgpu_info` parameter and check that it is populated in the HarvesterConfig.

### Manual Testing

Currently using this fork in our environment.


walkafwalka:
@matttrach Any bandwidth to look into this?

m-ildefons (changes requested):
I'm not a fan of letting users input JSON strings directly. Especially if there isn't ample documentation about the expected schema and other information available to users.

The vGPU information should be a subresource and the Terrafrom provider should marshall the user input into JSON automatically.
This lets us apply a wide variety of validation and makes it more user friendly, because the expected schema of the user input is better documented.

matttrach (approved):

matttrach:
This appears to be out of date and there is no response from the user since the reviewer asked for changes, so I am closing this PR for now, please open a new one based on the current main if this is still desired.

Issue #1588, opened 2025-07-10
[BUG] rancher2_cluster read error Invalid address to set imported_config


jcsston:
### Environment Information

- Terraform version: "v1.7.1"
- Provider version: "7.3.2"
- Rancher version: "v2.11.2"
- Operating System of the Rancher Node: "Ubuntu 24.04 + k3s v1.31"
- Operating System of the machine running Terraform: "Docker Image ghcr.io/runatlantis/atlantis:v0.33.0"
- Kubernetes Distro (k3s/rke2): "EKS"
- Kubernetes Version: "v1.31"
- Infrastructure Provider (AWS/GCP/vSphere/etc): "AWS"
- What is the scope of the token given to the provider?

### Describe the bug

I believe the recent addition of the imported_config PR
https://github.com/rancher/terraform-provider-rancher2/pull/1528
added a bug in 7.3 where trying to read rancher2_cluster results in Error: Invalid address to set: []string{"imported_config"}

Rolling the Rancher2 Terraform Provider back to v7.2.0 resolves the issue.

### To Reproduce

provider "rancher2" {
  api_url   = local.rancher_url
  token_key = local.rancher_token
}

data "rancher2_cluster" "cluster" {
  name = local.cluster_name
}

### Actual Result

data.rancher2_cluster.cluster: Reading...

Error: Invalid address to set: []string{"imported_config"}
  with data.rancher2_cluster.cluster,
  on main.tf line 40, in data "rancher2_cluster" "cluster":
  40: data "rancher2_cluster" "cluster" {

### Expected Result

data.rancher2_cluster.cluster: Reading...
data.rancher2_cluster.cluster: Read complete after 0s [id=c-5jpqp]

MichaelGerhart:
We are encountering the same issue, unfortunatley we were also excited because we needed the new functionality for setting the cluster level registry on imported clusters :(.



Planning failed. Terraform encountered an error while generating this plan.
╷
│ Error: Invalid address to set: []string{"imported_config"}
│ 
│   with data.rancher2_cluster.ue1,
│   on provider-ue1.tf line 7, in data "rancher2_cluster" "ue1":
│    7: data "rancher2_cluster" "ue1" {
│ 
╵
╷
│ Error: Invalid address to set: []string{"imported_config"}
│ 
│   with data.rancher2_cluster.ue2,
│   on provider-ue2.tf line 7, in data "rancher2_cluster" "ue2":
[...]

PR #1596, opened 2025-07-21
fix: registry configuration for imported hosted clusters

HarrisonWAffel:
## Problem

A prior PR implemented support for private registry configuration of imported clusters. However, this did not account for all imported cluster types, such as imported hosted clusters (AKS, EKS, GKE). Note that this PR does not add support for configuring a private registry for _provisioned_ hosted clusters. 

This PR also contains a change which aims to address https://github.com/rancher/terraform-provider-rancher2/issues/1588. Though I'm still working on reproducing that issue locally. 

## Solution

Remove the existing `conflictsWith` entries in favor of a more nuanced approach within the `CustomizeDiff` function. This PR does not remove all entries within `conflictsWith`, which helps to simply the logic added to the `CustomizeDiff` function. 

This change currently only applies for AKS, GKE, and EKS, as OKE does not have a single field which identifies a cluster as being either imported or provisioned. Due to this, OKE was not removed from the `conflictsWith` list. Support for OKE can be easily added later if a way to identify imported/provisioned is found. 

## Testing
+ Provision an imported cluster with an `imported_configuration` value which points to a sample private registry
+ Provision an _imported_ AKS/EKS/GKE cluster and using terraform 
    + Do this once without an `imported_configuration` and once with the field included
+ Configure a terraform file to _provision_ a new AKS/EKS/GKE cluster, include the `imported_configuration` block.
     + Confirm that an error is returned when running `terraform plan` or `terraform apply`

I'm still working on reproducing / validating #1588, I'll add more detailed steps once I do. 

### Manual Testing
I've done the above

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -48,6 +48,15 @@ func resourceRancher2Cluster() *schema.Resource {
 				}
 			}
 
+			// Allow the configuration of the imported_config field only if the
+			// cluster is an imported generic cluster or an imported hosted cluster (e.g. AKS, GKE, EKS).
+			// Previously defined 'conflictsWith' entries already handle other cluster types (rke, rke2, k3s)
+			// so they do not need to be reconsidered here.
+			importCnf, ok := d.Get("imported_config").([]interface{})
+			if ok && len(importCnf) > 0 && !isImportedCluster(d) {
+				return fmt.Errorf("The rancher2_cluster.imported_config field can only be used when working with generic imported clusters or imported hosted clusters (e.g. AKS, GKE, EKS)")
+			}
+
 			return nil
 		},
 		Schema:        clusterFields(),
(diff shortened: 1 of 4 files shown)


jakefhyde (approved):

snasovich (approved):
Looks sane to me

rancher2/resource_rancher2_cluster.go
```diff
@@ -48,6 +48,15 @@ func resourceRancher2Cluster() *schema.Resource {
 				}
 			}
 
+			// Allow the configuration of the imported_config field only if the
+			// cluster is an imported generic cluster or an imported hosted cluster (e.g. AKS, GKE, EKS).
+			// Previously defined 'conflictsWith' entries already handle other cluster types (rke, rke2, k3s)
+			// so they do not need to be reconsidered here.
+			importCnf, ok := d.Get("imported_config").([]interface{})
```
pedromfcarvalho, line 55: Just checking the usage of `Get` here: does it return the expected state after diffing? I'm wondering why we use `Get` here and `GetChange` below.

HarrisonWAffel: Good point, double checked and there isn't a meaningful difference between the two in this case. I've simplified to just using `Get`

pedromfcarvalho (approved):

kinarashah (approved):

matttrach (approved):

eva-vashkevich:
Hi @HarrisonWAffel @snasovich does this mean UI should also add private registry support to imported hosted clusters?

snasovich:
@eva-vashkevich , this was initially on the list for 2.12 enhancement for UI but was dropped. @Sahota1225 / @mitulshah-suse may have more info - I'll look for Slack messages on it too.

Issue #1609, opened 2025-08-04
prometheus-federator global.cattle.url value not documented when installing prometheus-federator


matttrach:
Duplicate of https://github.com/rancher/rancher/issues/49955 to track.

matttrach:
Main PR: #1608

matttrach:
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v8.0.0-rc13 has been cut to facilitate QA testing

axeal:
The examples in https://github.com/rancher/terraform-provider-rancher2/pull/1608/files and https://github.com/rancher/terraform-provider-rancher2/pull/1611/files set an incorrect value for cattle.global.url. The cattle.global.url value should be the Rancher URL, i.e. match the server-url setting of Rancher (protocol and hostname), e.g. https://rancher.example.com (that hostname is most commonly configured via the [helm chart hostname value](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster#5-install-rancher-with-helm-and-your-chosen-certificate-option), since we expect user's to run with a helm install of Rancher). Given Rancher is then deployed behind an ingress controller, setting cattle.global.url to the IP address that the Rancher URL resolves to, or of a local cluster node, will result in the monitoring URLs generated by the prometheus-federator failing to route to the Rancher Pods.

matttrach:
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v8.0.0-rc14 has been cut to reflect this change.

matttrach:
this has been released

PR #1613, opened 2025-08-05
fix: change NODE_IP for RANCHER_SERVER_URL on docs

susesamu:
## Issue: 

https://github.com/rancher/rancher/issues/49955
 
## Problem

When a user installs the prometheus-federator chart for Project Monitoring via the Rancher UI, the value global.cattle.url is auto-populated by Rancher during installation. When a user installs the same chart via the Rancher2 Terraform Provider, using a rancher2_app_v2 resource, this does not occur cause the values are inputted by the user when creating the resource. 
 
## Solution

This isn't the definitive solution, as we could create a resource for prometheus-federator and configure it in the required format. However, there's no documentation that exemplifies the correct creation using the rancher2_app_v2 resource, so adding it.
 
### Manual Testing

No testing, just documentation.

### DIFF
--- a/docs/guides/apps_marketplace.md
+++ b/docs/guides/apps_marketplace.md
@@ -208,7 +208,7 @@ global:
       enabled: false
     systemDefaultRegistry: registry.rancher.com
     systemProjectId: <PROJECT_ID>
-    url: https://<NODE_IP>
+    url: <RANCHER_SERVER_URL>
     clusterName: custom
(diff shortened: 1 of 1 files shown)


matttrach (approved):

matttrach:
addresses comment in #1609

matttrach:
Needs team approval before I can merge.

adamkpickering (approved):

Issue #730, opened 2021-08-21
Stale launch template version causing inconsistent final plan


sgran:
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for module.eks[0].rancher2_cluster.this to include new values learned so far during apply, provider "registry.terraform.io/rancher/rancher2" produced an invalid new value for
│ .eks_config_v2[0].node_groups[0].launch_template[0].version: was cty.NumberIntVal(0), but now cty.NumberIntVal(4).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

Given a rancher2_cluster with a nested block:

```
eks_config_v2 {
    node_groups {
      launch_template {
        id      = module.launch_template.id
        version = module.launch_template.latest_version
      }
    }
  }
  ```

This can be triggered by changing the launch_template parameters in the external resource.

**Comment by rawmind0** [comment]
It seems an issue managing tf modules dependencies, where the `version` value seems modified during the `rancher2_cluster`  execution. Are `module.launch_template` tasks finished before starting `rancher2_cluster` tasks?

**Comment by sgran** [comment]
Yes.  The DAG implies they must, in order to be evaluated correctly.

**Comment by melissa** [comment]
Hi, I'm running into this issue as well. I'm creating node groups with a launch template. My code has the same nested block as above, using `eks_config_v2`.

When I update the launch template, the plan output gives me `~ version = 1 -> 0`, where I would expect to see `version = (known after apply)`.

**Comment by nreisbeck** [comment]
Confirming that this is still an issue that we have run into as well, any update causing an update to the `launch_template` is causing an error like this.
```
terraform {
  required_version = ">= 1.5"

  required_providers {
    rancher2 = {
      source  = "rancher/rancher2"
      version = "3.2.0"
    }
  }
}
```
```
Initializing provider plugins...
- Finding hashicorp/null versions matching ">= 2.1.0"...
- Finding hashicorp/template versions matching "~> 2.2.0"...
- Finding rancher/rancher2 versions matching "3.2.0"...
- Finding hashicorp/aws versions matching ">= 2.0.0"...
- Finding hashicorp/tls versions matching ">= 2.0.0"...
- Finding hashicorp/local versions matching ">= 1.3.0"...
- Installing rancher/rancher2 v3.2.0...
- Installed rancher/rancher2 v3.2.0 (signed by a HashiCorp partner, key ID 2EEB0F9AD44A135C)
[...]

PR #1616, opened 2025-08-06
fix: change EKS node group launch template version

krunalhinguu:
## Issue: 
https://github.com/rancher/terraform-provider-rancher2/issues/730

## Problem
The Rancher2 provider does not reconcile computed launch template version values properly during Terraform runs. When the launch template version changes (e.g., due to new AMI or configuration updates), Terraform’s plan phase uses a default or stale version (0), but the apply phase sees the actual updated version (e.g., 4). This mismatch causes the provider to report an invalid new value error and fail the apply.

## Solution
The issue was resolved by updating the Rancher2 Terraform provider to mark the `launch_template.version` field as `computed = true` and removing the default value (`default = 1`). 

This change ensures that Terraform correctly waits for the actual launch template version value to be known during the apply phase, instead of relying on a hardcoded default or outdated value from the plan phase. 

With this fix, the provider can now handle dynamic version changes from the launch template (such as when a new version is triggered by configuration changes), preventing inconsistent plans and apply failures.

### Manual Testing
- Created an EKS cluster using Rancher and Terraform with launch templates for node groups.
- Updated the launch template to trigger a new version.
- Verified that Terraform successfully planned and applied the changes without errors.
- Checked the AWS console and Rancher UI to confirm the node groups used the expected launch template version.
- Destroyed and recreated the cluster to ensure reproducibility.

### Automated Testing
No automated tests were added, as the fix primarily addresses a runtime inconsistency in Terraform provider behavior rather than a logic bug.

## QA Testing Considerations
- Verify EKS cluster provisioning via Terraform with launch template versions explicitly set.
- Confirm that updating the launch template version triggers the correct node group updates without plan/apply errors.
- Test cluster upgrades involving node group changes to ensure launch template versions remain consistent.
- Validate Rancher UI correctly displays node group configurations and launch template details.

## Regressions Considerations
- The fix addresses only the launch template version handling and should not impact unrelated Rancher or EKS features.
- High chance of regressions in scenarios where launch templates are used with EKS node groups, if the version is omitted or incorrectly specified.
- Low risk of regressions for clusters not using launch templates or unmanaged via Terraform.

### DIFF
--- a/rancher2/schema_cluster_eks_config_v2.go
+++ b/rancher2/schema_cluster_eks_config_v2.go
@@ -58,7 +58,7 @@ func clusterEKSConfigV2NodeGroupsLaunchTemplateFields() map[string]*schema.Schem
 		"version": {
 			Type:        schema.TypeInt,
 			Optional:    true,
-			Default:     1,
+			Computed:    true,
 			Description: "The EKS node group launch template version",
 		},
 	}


mjura (approved):
LGTM

matttrach:
What version does this target?

matttrach (approved):

Issue #1637, opened 2025-09-03
[BUG] importedConfig.privateRegistryURL is not applied to an existing imported hosted cluster (AKS)


aeltai:
### Environment Information

- Provider version: 8.0.0, 7.4.0
- Rancher version: 2.12.0/2.12.1
- Operating System of the Rancher Node: /
- Operating System of the machine running Terraform: MacOS
- Kubernetes Distro (k3s/rke2): AKS
- Kubernetes Version: 1.33
- Infrastructure Provider (AWS/GCP/vSphere/etc): Azure
### Describe the bug
For existing imported AKS-Clusters, when running "terraform/tofu plan" which would patch the imported_config, private_registry_url section of the yaml it correctly shows me the changes it wants to do. (See Attachment)
Then it runs successful. 
However the Cluster yaml is not touched by that. 
When i manually put in the line in the yaml it lets me to that but does not trigger a redeployment of workloads from the new registry and manually triggering that also does not lead to deployments from the registry.
My workaround at the moment is to delete the cluster from cluster management and then run a new terraform plan with the import resource. This works fine.

### To Reproduce

- Create and import an AKS cluster into Rancher Manager 2.12.0/2.12.1.
- Add the importedConfig.privateRegistryURL to the terraform ressource.
- Run "tofu plan"
- Observe the planned change
- Run tofu apply
- Observe the success message
- Check cluster-yaml of the the imported AKS cluster
- Check that importedConfig.privateRegistryURL was not set

### Actual Result

-  importedConfig.privateRegistryURL was not modified/set

### Expected Result

-  importedConfig.privateRegistryURL is modified/set &&  a redeployment of (e.g. rancher-system-agent pulling from the new private Registry) is triggered

<img width="1320" height="300" alt="Image" src="https://github.com/user-attachments/assets/4337db75-2795-4d90-b58b-c0bc185d3423" />
<img width="1264" height="448" alt="Image" src="https://github.com/user-attachments/assets/e351a667-8dfa-4c31-a35e-1c46658abd9c" />

matttrach:
I am not sure if this is your issue, but OpenTofu isn't supported yet for this provider. While it may work, we are using the older SDKv1 which ToFu may have trouble with. What version of Terraform/ToFu are you running?

This is more likely to be a caching problem or an unexpected interaction with kubernetes, but I want to cover all our bases.

aeltai:
Hi Matt,
I have tried again using Terraform:
```
[...]

PR #1651, opened 2025-09-17
fix: process imported_config outside switch

krunalhinguu:
## Issue
- https://github.com/rancher/terraform-provider-rancher2/issues/1637

## Problem
Currently, the `imported_config` block is only processed when the driver is explicitly set to `"imported"`.  
This causes updates to fail or silently ignore `imported_config` settings (e.g., `private_registry_url`) for hosted imported clusters, since they use their respective driver blocks (`aks_config_v2`, `eks_config_v2`, `gke_config_v2`) instead of the `"imported"` driver.

## Solution
The handling of `imported_config` has been pulled out of the driver switch block and is now applied independently of driver type.  
This ensures `imported_config` is always processed when defined, including for hosted imported clusters during updates.  

## Testing
- Import an existing AKS/EKS/GKE cluster into Rancher using Terraform.  
- Apply an update that adds an `imported_config` block (e.g., setting `private_registry_url`).  
- Confirm that the update succeeds and the imported config is applied in Rancher.  
- Validate that clusters without `imported_config` remain unaffected.  

### Manual Testing
- Imported an AKS cluster (`aks_config_v2` with `imported = true`) into Rancher.  
- Applied an update with `imported_config` defined and verified the changes were applied correctly.  
- Imported an EKS cluster and repeated the above validation.  
- Confirmed that `"imported"` driver clusters still accept `imported_config`.  

### Automated Testing
- No automated tests added at this stage.  
- Regression risk is limited to update handling for clusters with `imported_config`.  

## QA Testing Considerations
- Import AKS/EKS/GKE clusters into Rancher, then update them with `imported_config` to ensure changes apply.  
- Validate that `"imported"` driver clusters continue to work.  
- Ensure updates to non-imported hosted clusters (AKS/EKS/GKE without `imported_config`) succeed without errors.  

- Regression risk: updates to `"imported"` driver clusters.  
- Probability: low, as logic was only refactored to run before driver-specific handling.

### DIFF
--- a/rancher2/resource_rancher2_cluster.go
+++ b/rancher2/resource_rancher2_cluster.go
@@ -321,6 +321,18 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
+	// imported_config is special because it applies to *any imported cluster*,
+	// including hosted providers (AKS, EKS, GKE) where `imported = true` is set.
+	// If we keep it inside the driver-specific switch, it only gets processed
+	// for the "imported" driver, but NOT for hosted imported clusters.
+	// To avoid this bug, we always check imported_config here before switching on driver.
+	if v, ok := d.GetOk("imported_config"); ok {
+		importedConfig := v.([]interface{})
+		if len(importedConfig) > 0 {
+			update["importedConfig"] = expandClusterImportedConfig(importedConfig)
+		}
+	}
+
 	replace := false
 	switch driver := ToLower(d.Get("driver").(string)); driver {
 	case ToLower(clusterDriverAKSV2):
@@ -346,8 +358,6 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
 	case clusterDriverRKE2:
 		update["rke2Config"] = expandClusterRKE2Config(d.Get("rke2_config").([]interface{}))
 		replace = d.HasChange("cluster_agent_deployment_customization")
-	case clusterDriverImported:
-		update["importedConfig"] = expandClusterImportedConfig(d.Get("imported_config").([]interface{}))
 	}
 
 	// update the cluster; retry til timeout or non retryable error is returned. If api 500 error is received,


rancher2/resource_rancher2_cluster.go
```diff
@@ -321,6 +321,18 @@ func resourceRancher2ClusterUpdate(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
+	// imported_config is special because it applies to *any imported cluster*,
+	// including hosted providers (AKS, EKS, GKE) where `imported = true` is set.
```
mjura, line 325: I am not sure, if we are mixing here settings. For hosted providers `imported` flag is set inside hosted_provider_spec, e.g. https://github.com/rancher/aks-operator/blob/main/pkg/apis/aks.cattle.io/v1/types.go#L43 

Even if hosted provider imported cluster is set, then it should be managed by hoster provider operator

matttrach (approved):

rancher2/resource_rancher2_cluster.go
krunalhinguu: We got this example from customer 

So do you mean that this is incorrect way of initialising it? And it should be inside the `aks_config_v2`?

```
resource "rancher2_cluster" "aks_imported" {
  name        = var.aks_cluster_name
  description = "Imported AKS cluster via Terraform"
  aks_config_v2 {
    cloud_credential_id = rancher2_cloud_credential.aks.id
    resource_group      = var.resource_group_name
    resource_location   = var.azure_location
    imported            = true
  }
  imported_config {
      private_registry_url = "registry.suse.com"
  }
}

matttrach:
This looks good from a general Terraform point of view, but we still need approval from the responsible team to merge.

rancher2/resource_rancher2_cluster.go
mjura: sorry, it looks good to me, I had to understand bug description :+1:

mjura (approved):

Issue #1652, opened 2025-09-17
[BUG] Remove `PodSecurityAdmissionConfigurationTemplate` succeeds but stuck on destroying until timeout


jiaqiluo:
### Environment Information

any 

### Describe the bug

This is a bug in the [resourceRancher2PodSecurityAdmissionConfigurationTemplateDelete](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/resource_rancher2_pod_security_admission_configuration_template.go#L106) and [podSecurityAdmissionConfigurationTemplateStateRefreshFunc](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/resource_rancher2_pod_security_admission_configuration_template.go#L150) functions that do not properly handle the state of the resources during the removal process, so that Terraform always times out waiting for the PSACT to be deleted.

### To Reproduce

- use TF to create a `rancher2_pod_security_admission_configuration_template` resource 
- run `terraform destroy` to remove it

### Actual Result

The PSACT resource is removed, but  TF is stuck in the "Still destroying"  statue and eventually times out. 

```
rancher2_pod_security_admission_configuration_template.t1: Destroying... [id=jiaqi-foo]
...
rancher2_pod_security_admission_configuration_template.t1: Still destroying... [id=jiaqi-foo, 01m20s elapsed]
rancher2_pod_security_admission_configuration_template.t1: Still destroying... [id=jiaqi-foo, 01m30s elapsed]
...
rancher2_pod_security_admission_configuration_template.t1: Still destroying... [id=jiaqi-foo, 03m10s elapsed]
╷
│ Error: [ERROR] failed to remove PodSecurityAdmissionConfigurationTemplate with ID jiaqi-foo: couldn't find resource (21 retries)

```

### Expected Result

TF should remove the resource and return successfully timely.

jiaqiluo:
Hi @matttrach, it will be great if you can help manage the milestone and backporting. Please let me know if anything I can help.

markusewalker:
Reproduced this on `head` https://github.com/rancher/rancher/commit/a4e92bf8e657916186fc98bc064da8237bb97947 on Rancher2 `8.1.0`.

markusewalker:
Closing issue as this has been validated in issues https://github.com/rancher/terraform-provider-rancher2/issues/1670 and https://github.com/rancher/terraform-provider-rancher2/issues/1671.

PR #1653, opened 2025-09-17
fix: PSACT deletion hanging

jiaqiluo:
## Issue: 

https://github.com/rancher/terraform-provider-rancher2/issues/1652 

## Problem

 
When deleting a PSACT, Terraform always times out waiting for the PSACT to be deleted.  This is because the  [podSecurityAdmissionConfigurationTemplateStateRefreshFunc](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/resource_rancher2_pod_security_admission_configuration_template.go?rgh-link-date=2025-09-17T18%3A22%3A58.000Z#L150) functions that do not properly handle the state of the resources during the removal process 

## Solution

 
Fix the function to properly reflect the status of the targeted PSACT resource. 

### Manual Testing

Tested locally with the rancher-provider built from the branch to confirm that the deletion is not hanging anymore. 

### Automated Testing

n/a 

## QA Testing Considerations

The same use case as provded in the linked GH issue. 

 
n/a

### DIFF
--- a/rancher2/resource_rancher2_pod_security_admission_configuration_template.go
+++ b/rancher2/resource_rancher2_pod_security_admission_configuration_template.go
@@ -147,8 +147,15 @@ func resourceRancher2PodSecurityAdmissionConfigurationTemplateDelete(d *schema.R
 
 // podSecurityAdmissionConfigurationTemplateStateRefreshFunc returns a resource.StateRefreshFunc that is used to watch
 // a Rancher PodSecurityAdmissionConfiguration Template
-func podSecurityAdmissionConfigurationTemplateStateRefreshFunc(client *managementClient.Client, pspID string) resource.StateRefreshFunc {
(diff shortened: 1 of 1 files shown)


jiaqiluo:
The CI is driving me crazy 😢 
The resource has a long name : `PodSecurityAdmissionConfigurationTemplate`. 
But the CI does not allow `psact` and also won’t let me use more than 50 characters.

HarrisonWAffel (approved):
Yeah I agree the CI can be fickle, I had similar issues with the GCE work. Maybe it can could be improved to accept acronyms if they're in call caps (i.e. PSACT) as a way to differentiate between actual spelling mistakes and intentional acronyms 

It also looks like the commit needs to be signed (verified)

jakefhyde (approved):

matttrach (approved):

matttrach:
To allow an acronym, update the https://github.com/rancher/terraform-provider-rancher2/blob/main/aspell_custom.txt

Issue #1674, opened 2025-10-02
fix: improve registry documentation


matttrach:
This is the main issue tracking #1664 
Copied from PR:
## Issues: 
- https://github.com/rancher/terraform-provider-rancher2/issues/1632
- https://github.com/rancher/terraform-provider-rancher2/issues/1048#issuecomment-3329824082
 
## Problem
The current private registry documentation has two issues 

1. The `config` field was changed in [August of 2023](https://github.com/rancher/terraform-provider-rancher2/pull/1181), after the initial documentation was added. It's now expected to be a string representing a yaml object. 
2. The secret which is being referenced in terraform is only briefly mentioned, and users would like a more explicit guide on how to create it. 

## Solution
Update the documentation to fix how `config` is passed, and to add an example secret resource which creates the registry secret. 
 
## Testing
+ Standup a pull-through private registry server, configured with https (potentially behind ngrok)
+ Configure a terraform file following the documentation. I can provide QA with some resources if needed
+ Apply the file, ensure that the cluster comes up properly and that the system images are pulled from the configured registry

### Manual Testing
I've done the above

### Automated Testing
N/A

## QA Testing Considerations
You'll need a private registry configured with a domain name and a valid certificate. Using ngrok is probably the easiest way to do this. 
 
### Regressions Considerations
N/A, just docs updates

matttrach:
These have been cut to facilitate testing:
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v13.0.0-rc3
https://github.com/rancher/terraform-provider-rancher2/releases/tag/v8.3.0-rc1

matttrach:
released

PR #1679, opened 2025-10-02
fix: improve registry documentation (#1664)

matttrach:
- backport #1664 to release/v13 for #1674 
- addresses #1675
Copied from PR:
## Problem
The current private registry documentation has two issues 

1. The `config` field was changed in [August of 2023](https://github.com/rancher/terraform-provider-rancher2/pull/1181), after the initial documentation was added. It's now expected to be a string representing a yaml object. 
2. The secret which is being referenced in terraform is only briefly mentioned, and users would like a more explicit guide on how to create it. 

## Solution
Update the documentation to fix how `config` is passed, and to add an example secret resource which creates the registry secret. 
 
## Testing
+ Standup a pull-through private registry server, configured with https (potentially behind ngrok)
+ Configure a terraform file following the documentation. I can provide QA with some resources if needed
+ Apply the file, ensure that the cluster comes up properly and that the system images are pulled from the configured registry

### Manual Testing
I've done the above

### Automated Testing
N/A

## QA Testing Considerations
You'll need a private registry configured with a domain name and a valid certificate. Using ngrok is probably the easiest way to do this. 
 
N/A, just docs updates


HarrisonWAffel (approved):
This looks good for the documentation backport, I think the PR description is mentioning a different issue than whats being backported here though

jiaqiluo (approved):

Issue #1500, opened 2025-05-02
[RFE] Add `networking` and `stack_preference` child blocks to `rke_config` to better test IPv6 downstream clusters


markusewalker:
### Is your feature request related to a problem? Please describe.

As part of broader QA efforts to testing Rancher in an IPv6 environment, we're currently enhancing our test framework to support this. During this, one of the enhancements needed to our custom cluster tests is accounting for different values found in `spec.rkeConfig.networking.stackPreference`. However, looking at the documentation, there is no support for `networking` or child resource `stackPreference`.

This would be a great help to have this supported so that we can add this to our automation that we run each release testing. While `rke_config` already allows for `cluster-cidr` and `service-cidr`, the complete use-case from QA standpoint would be adding `networking.stackPreference` so that we can test each option `ipv4`, `dual` and `ipv6`.

### Describe the solution you'd like

`rke_config` to be enhanced to support a `networking` block along with a child block `stackPreference` so that QA can better test custom clusters with an IPv6 environment.

markusewalker:
@matttrach as QA is ramping up on IPv6 testing, I wanted to follow back on this ticket. Is it a possibility that this may be prioritized for an upcoming release?

jiaqiluo:
Waiting for RC to be cut from [the release/v13 branch](https://github.com/rancher/terraform-provider-rancher2/tree/release/v13)

jiaqiluo:
This issue can be validated on [v13.0.0-rc.5](https://github.com/rancher/terraform-provider-rancher2/releases/tag/v13.0.0-rc.5)

markusewalker:
Validated that this is addressed on `head` https://github.com/rancher/rancher/commit/024946c206a7db788dd9d72a7a90ecec9f244bd8 with Rancher2 provider version `v13.0.0-rc.6`. Please find below the test scenarios that were considered:

|#|Scenario|Result|
|---|---|---|
|1|Setup IPv6 Rancher environment, provision a downstream AWS RKE2 node driver cluster using Rancher2 provider with stack preference set to `ipv6` and CIDRs defined|:white_check_mark:|
|2|Setup dual-stack Rancher environment, provision a downstream AWS RKE2 node driver cluster using Rancher2 provider with stack preference set to `dual` and CIDRs defined|:white_check_mark:|
[...]

PR #1682, opened 2025-10-02
support AWS node-driver IPv6/dual-stack downstream clusters 

jiaqiluo:
## Issue: 

 
- https://github.com/rancher/rancher/issues/52174
- https://github.com/rancher/terraform-provider-rancher2/issues/1500

## Problem

 
We would like to enhance rancher-terraform-provider to support AWS node-driver IPv6 downstream clusters 

## Solution

This PR introduces the new fields to the corresponding resources:

- Add new fields to machineConfigV2Amazonec2Fields
- Add the networking field to ClusterV2RKEConfig

 
 
### Manual Testing

Using the rancher-terraform-provider binary built from this PR, I was able to successfully provision AWS node-driver IPv6 / dual-stack downstream clusters.

image

### Automated Testing

Unit tests are added at proper levels. 

## QA Testing Considerations

 
We should be able to provision AWS node-driver IPv6/ dual-stack downstream clusters. 

We should be able to provision AWS node-driver IPv4 downstream clusters.

### DIFF
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -150,6 +150,10 @@ The following attributes are exported:
 * `use_private_address` - (Optional) Force the usage of private IP address. Default `false` (bool)
 * `userdata` - (Optional) Path to file with cloud-init user data (string)
 * `volume_type` - (Optional) Amazon EBS volume type. Default `gp2` (string)
+* `http_protocol_ipv6` - (Optional) Enables or disables the IPv6 endpoint for the instance metadata service. Options: enabled, disabled (string)
+* `ipv6_address_count` - (Optional) The number of IPv6 addresses to assign to the network interface. It must be greater than zero when Ipv6AddressOnly is true (string)
+* `ipv6_address_only` - (Optional) Indicates whether the instance has only IPv6 address. Useful when the VPC or subnet is configured as IPv6-only (bool)
+* `enable_primary_ipv6` - (Optional) Indicates whether the instance’s first assigned IPv6 address is set as the primary IPv6 address (bool)
 
 ### `azure_config`
 
--- a/docs/resources/machine_config_v2.md
+++ b/docs/resources/machine_config_v2.md
@@ -151,7 +151,7 @@ The following attributes are exported:
 * `userdata` - (Optional) Path to file with cloud-init user data (string)
 * `volume_type` - (Optional) Amazon EBS volume type. Default `gp2` (string)
 * `http_protocol_ipv6` - (Optional) Enables or disables the IPv6 endpoint for the instance metadata service. Options: enabled, disabled (string)
-* `ipv6_address_count` - (Optional) The number of IPv6 addresses to assign to the network interface. It must be greater than zero when Ipv6AddressOnly is true (string)
+* `ipv6_address_count` - (Optional) The number of IPv6 addresses to assign to the network interface. It must be greater than zero when `ipv6_address_only` is true (string)
 * `ipv6_address_only` - (Optional) Indicates whether the instance has only IPv6 address. Useful when the VPC or subnet is configured as IPv6-only (bool)
 * `enable_primary_ipv6` - (Optional) Indicates whether the instance’s first assigned IPv6 address is set as the primary IPv6 address (bool)
 
--- a/rancher2/schema_cluster_v2_rke_config.go
+++ b/rancher2/schema_cluster_v2_rke_config.go
@@ -5,9 +5,15 @@ import (
 	"reflect"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 // Types
+
+var (
+	// StackPreference is the networking stack used by the cluster
+	StackPreference = []string{"dual", "ipv6", "ipv4"}
+)
 
 func clusterV2RKEConfigFieldsV0() map[string]*schema.Schema {
 	s := map[string]*schema.Schema{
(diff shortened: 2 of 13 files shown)


rancher2/schema_cluster_v2_rke_config.go
```diff
@@ -172,6 +182,24 @@ func clusterV2RKEConfigFieldsV0() map[string]*schema.Schema {
 	return s
 }
 
+func clusterV2Networking() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		"stack_preference": {
+			Type:     schema.TypeString,
```
jferrazbr, line 194: Would it be worth adding `validation.StringInSlice` so we catch invalid `stack_preference` values?

rancher2/structure_cluster_v2_rke_config_network.go
```diff
@@ -0,0 +1,38 @@
+package rancher2
+
+import (
+	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+)
+
+// Flatteners
+
+func flatterClusterV2Networking(in *rkev1.Networking) []interface{} {
```
jferrazbr: nit: I think this was meant to be flattenClusterV2Networking (currently flatterClusterV2Networking). 🙇

rancher2/structure_machine_config_v2_amazonec2.go
```diff
@@ -51,6 +51,25 @@ type machineConfigV2Amazonec2 struct {
...
+	// Indicates whether the instance’s first assigned IPv6 address is set as the primary IPv6 address.
+	// Enable this option if the instance requires a stable, persistent IPv6 address.
+	// This option does not affect whether IPv6 addresses are assigned to the instance.
+	// For more information, see EnablePrimaryIpv6 on
+	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html
+	EnablePrimaryIpv6 bool `json:"enablePrimaryIpv6,omitempty" yaml:"enablePrimaryIpv6,omitempty"`
+
+	// The number of IPv6 addresses to assign to the network interface.
+	// It must be greater than zero when Ipv6AddressOnly is true
+	Ipv6AddressCount string `json:"ipv6AddressCount,omitempty" yaml:"ipv6AddressCount,omitempty"`
```
jferrazbr, line 68: Would it be worth adding a CustomizeDiff to validate this constraint in the resourceRancher2MachineConfigV2?

rancher2/schema_cluster_v2_rke_config.go
jiaqiluo: Good point, the validation was added.

rancher2/structure_cluster_v2_rke_config_network.go
jiaqiluo: Fixed!

rancher2/structure_machine_config_v2_amazonec2.go
jiaqiluo: Using CustomizeDiff for value validation would be an incorrect usage of the field. 
I also verified that the ValidateFunc field cannot be applied to the `amazonec2_config` schema since its type is a list.

Given these limitations and the fact that both field descriptions and documentation already clearly state the requirements,  I think it’s reasonable not to add additional validation in this case. 

WDYT?

jferrazbr: Yeah, that makes total sense. I dug in and found HashiCorp confirming CustomizeDiff can handle [cross-attribute validation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/customdiff#All), but it [can cause unintended side effects](https://discuss.hashicorp.com/t/cross-attribute-validation/24605/3). With that in mind, I’m fine to skip it here and rely on documentation. Thanks for the guidance!

jferrazbr (approved):

thatmidwesterncoder (approved):

rancher2/structure_machine_config_v2_amazonec2.go
```diff
@@ -51,6 +51,25 @@ type machineConfigV2Amazonec2 struct {
 	VolumeType              string   `json:"volumeType,omitempty" yaml:"volumeType,omitempty"`
 	VpcID                   string   `json:"vpcId,omitempty" yaml:"vpcId,omitempty"`
 	Zone                    string   `json:"zone,omitempty" yaml:"zone,omitempty"`
+
+	// Enables or disables the IPv6 endpoint for the instance metadata service.
```
HarrisonWAffel: non-blocking nit: 
```suggestion
	// HttpProtocolIpv6 Enables or disables the IPv6 endpoint for the instance metadata service.
```
Same for other fields as well, but again, non-blocking

docs/resources/machine_config_v2.md
```diff
@@ -150,6 +150,10 @@ The following attributes are exported:
 * `use_private_address` - (Optional) Force the usage of private IP address. Default `false` (bool)
 * `userdata` - (Optional) Path to file with cloud-init user data (string)
 * `volume_type` - (Optional) Amazon EBS volume type. Default `gp2` (string)
+* `http_protocol_ipv6` - (Optional) Enables or disables the IPv6 endpoint for the instance metadata service. Options: enabled, disabled (string)
+* `ipv6_address_count` - (Optional) The number of IPv6 addresses to assign to the network interface. It must be greater than zero when Ipv6AddressOnly is true (string)
```
HarrisonWAffel: small nit
```suggestion
* `ipv6_address_count` - (Optional) The number of IPv6 addresses to assign to the network interface. It must be greater than zero when `ipv6_address_only` is true (string)
```

HarrisonWAffel (approved):
Nice! just two optional comments

rancher2/structure_machine_config_v2_amazonec2.go
jiaqiluo: Fixed.
Thank you for reviewing.

docs/resources/machine_config_v2.md
jiaqiluo: fixed

HarrisonWAffel (approved):

PR #1692, opened 2025-10-23
fix: add automation for release issues and tests

matttrach:
## Problem
Our end to end tests don't run automatically.
Too much time is spent generating tracking issues and backports.
 
## Solution
Automate the generation of tracking tickets, backports, rc-releases (including artifact generation), and artifacts.
Include e2e tests in release automation.
 
## Testing
Actionlint is the best we can do for the workflows, I have run all of the e2e tests manually.
Most of this automation is working in the [File provider](https://github.com/rancher/terraform-provider-file), excluding the release workflow which needed to change to include the more intense e2e testing needed for this project.

### Manual Testing
I have run all of the e2e tests manually.

### Automated Testing
Added new automation for releases which includes running all e2e tests.

## QA Testing Considerations
This adds testing, so hopefully there isn't anything for QA to validate.


.github/workflows/backport-issues.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          script: |
+            const labelName = context.payload.label.name;
+            const parentIssue = context.payload.issue;
+            const parentIssueTitle = parentIssue.title;
+            const parentIssueNumber = parentIssue.number;
+            const repo = context.repo.repo;
+            const owner = context.repo.owner;
+            const parentIssueBody = parentIssue.body;
+            const assignees = ['matttrach', 'jiaqiluo', 'HarrisonWAffel'];
```
jiaqiluo: The list of assignees is currently hardcoded. This can be difficult to maintain as team members change.

Let's make this list more manageable by moving it to the workflow's env block. This centralizes configuration and makes it easier to update without digging into the script.

```
env:
  BACKPORT_ASSIGNEES: '["matttrach", "jiaqiluo", "HarrisonWAffel"]'
```

then use it here
```
const assignees = JSON.parse(process.env.BACKPORT_ASSIGNEES);
```

.github/workflows/backport-issues.yml
```diff
@@ -0,0 +1,79 @@
...
+            // Attach the sub-issue to the parent, use REST API because there isn't a github-script API yet.
+            await github.request('POST /repos/{owner}/{repo}/issues/{issue_number}/sub_issues', {
+              owner: owner,
+              repo: repo,
+              issue_number: parentIssueNumber,
+              sub_issue_id: subIssueId,
+              headers: {
+                'X-GitHub-Api-Version': '2022-11-28'
+              }
+            });
```
jiaqiluo: If this API call fails, the script will fail, but the backport issue will have already been created, leaving it unlinked. 
You can make this more resilient by wrapping the API call in a try...catch block to log a warning instead of failing the entire job.

.github/workflows/backport-pr-manual.yml
```diff
@@ -0,0 +1,123 @@
...
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script
+        env:
+          MERGE_COMMIT_SHA: ${{ inputs.merge_commit_sha }}
+        with:
+          script: |
+            const execSync = require('child_process').execSync;
+            const owner = "rancher";
+            const repo = "terraform-provider-file";
+            const mergeCommitSha = process.env.MERGE_COMMIT_SHA;
+            const assignees = ['matttrach', 'jiaqiluo', 'HarrisonWAffel'];
```
jiaqiluo: Could you move it to an env var? See my other commit for details.

.github/workflows/backport-prs.yml
```diff
@@ -0,0 +1,120 @@
...
+          fetch-depth: 0
+      - name: 'Find Issues and Create Cherry-Pick PRs'
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script
+        with:
+          script: |
+            const execSync = require('child_process').execSync;
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            const mergeCommitSha = context.payload.head_commit.id;
+            const assignees = ['matttrach', 'jiaqiluo', 'HarrisonWAffel'];
```
jiaqiluo: Could you move it to an env var? See my other commit for details.

.github/workflows/main-issue.yml
```diff
@@ -0,0 +1,69 @@
...
+          script: |
+            const repo = context.repo.repo;
+            const owner = context.repo.owner;
+            const pr = context.payload.pull_request;
+            const newLabels = ['internal/main'];
+            const releaseLabel = pr.labels.find(label => label.name.startsWith('version/v'));
+            if (releaseLabel) {
+              newLabels.push(releaseLabel);
+            }
+            const assignees = ['matttrach', 'jiaqiluo', 'HarrisonWAffel'];
```
jiaqiluo: Could you move it to an env var? See my other commit for details.

.github/workflows/release.yml
```diff
@@ -0,0 +1,290 @@
...
+
+  report:
+    needs:
+      - release
+      - test
+      - cleanup
+    if: always() && needs.release.outputs.release_pr
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v8
```
jiaqiluo:
```suggestion
      -  uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script/commits/main
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,290 @@
...
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.rest.issues.createComment({
+              issue_number: ${{ fromJson(needs.release.outputs.release_pr).number }},
+              owner: "${{ github.repository_owner }}",
+              repo: "${{ github.event.repository.name }}",
+              body: "End to End Tests Passed! \n ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            })
+      - uses: actions/github-script@v8
```
jiaqiluo:
```suggestion
      -  uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script/commits/main
```

.golangci.yml
```diff
@@ -0,0 +1,44 @@
+
+
```
jiaqiluo: Can empty lines be removed?

.goreleaser.yml
```diff
@@ -0,0 +1,65 @@
+
+
```
jiaqiluo: empty lines?

.goreleaser.yml
```diff
@@ -0,0 +1,65 @@
...
+    ignore:
+      - goos: windows
+        goarch: arm
+archives:
+  - formats: [ 'zip' ]
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+checksum:
+  extra_files:
+    - glob: 'terraform-registry-manifest.json'
+      name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
```
jiaqiluo, line 39: We can  use a YAML anchor to define this block once and reuse it. This reduces duplication and makes future changes easier.

```
extra_files: &extra_files
  - glob: 'terraform-registry-manifest.json'
    name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
```

Then use it 
```
checksum:
  extra_files:
    <<: *extra_files
```

.goreleaser_rc.yml
```diff
@@ -0,0 +1,62 @@
...
+    ignore:
+      - goos: windows
+        goarch: arm
+archives:
+  - formats: [ 'zip' ]
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+checksum:
+  extra_files:
+    - glob: 'terraform-registry-manifest.json'
+      name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
```
jiaqiluo, line 36: We can use a YAML anchor to define this block once and reuse it

examples/use-cases/production/variables.tf
```diff
@@ -36,15 +36,14 @@ variable "rke2_version" {
...
-    eg. v1.30.2+rke2r1
   EOT
 }
 variable "rancher_version" {
   type        = string
   description = <<-EOT
     The version of rancher to install on the rke2 cluster.
   EOT
-  default     = "2.9.2"
+  default     = "2.9.1"
```
jiaqiluo: Such an old version, can we bump it to 2.12.3?

run_tests.sh
```diff
@@ -0,0 +1,277 @@
...
+elif [ -d "test" ]; then
+  TEST_DIR="test"
+else
+  echo "Error: Unable to find tests directory" >&2
+  exit 1
+fi
+
+run_tests() {
+  local slow_mode=$1
+  local TEST_DIR="$2"
```
jiaqiluo: use lower case for local variable 
```suggestion
  local test_dir="$2"
```

run_tests.sh
```diff
@@ -0,0 +1,277 @@
...
+
+  REPO_ROOT="$(git rev-parse --show-toplevel)"
```
jiaqiluo: 1/ The same value is set at the global level 
2/ Should it be a local variable?

test/one/basic_test.go
```diff
@@ -0,0 +1,182 @@
...
+		err = aws.DeleteEC2KeyPairE(t, keyPair)
+		if err != nil {
+			t.Logf("Failed to destroy key pair: %v", err)
+		}
+		os.RemoveAll(testDir)
+		t.Fatalf("Error creating test key pair: %s", err)
+	}
+	sshAgent := ssh.SshAgentWithKeyPair(t, keyPair.KeyPair)
+	t.Logf("Key %s created and added to agent", keyPair.Name)
+
```
jiaqiluo: I'll propose creating a `TestConfig` struct and a `NewTestConfig` helper function to encapsulate this common setup.

test/one/basic_test.go
```diff
@@ -0,0 +1,182 @@
...
+	backendTerraformOptions, err := util.CreateObjectStorageBackend(t, testDir, id, owner, region)
+	tfOptions := []*terraform.Options{backendTerraformOptions}
+	if err != nil {
+		t.Log("Test failed, tearing down...")
+		util.Teardown(t, testDir, exampleDir, tfOptions, keyPair, sshAgent)
```
jiaqiluo: This is repetitive and can be fragile. If a panic occurs, the cleanup code might not be executed.
Go's `defer` statement is the perfect solution.

test/summarize.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
+
```
jiaqiluo, line 3:
```suggestion
set -e # Exit immediately if a command exits with a non-zero status.
```

test/summarize.sh
```diff
@@ -0,0 +1,38 @@
...
+  jq --slurp -r --arg action "$action" \
+    '
+    # select all objects with a .Test listed that matches $acton and store it as $tests
+    (map(select(.Test and .Action == $action) | .Test ) | unique) as $tests |
+    # iterate through tests and save the test name as $prefix
+    $tests[] | . as $prefix |
+    # filter out any test names that dont exactly match the current test name, but do have the test name in them followed by a slash
+    select(any($tests[]; . != $prefix and startswith($prefix+"/"))|not)
+    # this leaves only test names that dont have duplicate prefixes, ie. the actual tests and not the parent blocks
+    '\
```
jiaqiluo: Can you move the inline comments out of the blob for readability?

jiaqiluo (changes requested):
Left some comments, and some can be applied to multiple places in multiple files.

.github/workflows/backport-issues.yml
HarrisonWAffel: `r/r` uses a nice pattern of defining a scripts directory for GHA workflows specifically, and then invoking them in the step. It looks like this is possible with the `github-script` action, https://github.com/actions/github-script#run-a-separate-file. Taking this approach may make handling Jacks comments a bit easier, as the length of the script won't negatively impact the readability of this workflow file.

I'll admit this approach requires a bit of javascript magic, but it would help with maintainability going forward

.github/workflows/backport-issues.yml
HarrisonWAffel: I kind of disagree with the idea of assigning everyone on the team to _every backport issue_ created. It's going to be very noisy. Is it possible to just carry forward the assignee's from the root issue into the backport? This would be useful when someone from another part of Rancher (or the community) is responsible for a bug fix or feature.

.github/workflows/backport-pr-manual.yml
```diff
@@ -0,0 +1,123 @@
...
+jobs:
+  create-cherry-pick-prs:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+      actions: write
+    steps:
+      - name: 'Wait for merge to settle'
```
HarrisonWAffel, line 22: Is there anyway we could we do a backoff / polling approach here? 

I'm not sure how long it typically takes for the merges to settle, but I can imagine a situation where GH is under load and this takes longer than expected. I guess it's not a big deal, since we can retry the action run manually, but it would still be a nice QOL improvement IMO

.github/workflows/main-issue.yml
HarrisonWAffel: Again, I disagree with the approach of assigning the whole team to every _issue_ created. I understand the benefit of all of us being tagged when PRs come in, but I don't think its useful to have notifications sent for the full life cycle of every issue.

.github/workflows/backport-issues.yml
```diff
@@ -0,0 +1,79 @@
+name: Backports
+# This workflow generates "backport" issues when a release branch label is added to an issue
+on:
+  issues:
+    types: [labeled] # triggered when any label is added to an issue
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.event.label.name, 'version/v') }}
```
HarrisonWAffel: I could be misunderstanding the workflow, but does this handle the case where a version label is being set for the primary issue? If the primary issue is targeting the upcoming release (e.g. `version/v9`), which doesn't need a backport, will this still create a backport issue (effectively a duplicate)?

.github/workflows/backport-pr-manual.yml
```diff
@@ -0,0 +1,123 @@
...
+          fetch-depth: 0
+      - name: 'Find Issues and Create Cherry-Pick PRs'
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script
+        env:
+          MERGE_COMMIT_SHA: ${{ inputs.merge_commit_sha }}
+        with:
+          script: |
+            const execSync = require('child_process').execSync;
+            const owner = "rancher";
+            const repo = "terraform-provider-file";
```
HarrisonWAffel: I think this is pointing to the wrong repo

.github/workflows/release.yml
```diff
@@ -0,0 +1,290 @@
...
+        with:
+          fetch-depth: 0
+      - uses: googleapis/release-please-action@c2a5a2bd6a758a0937f1ddb1e8950609867ed15c # v4.3.0 https://github.com/googleapis/release-please-action/commits/main/
+        name: release-please
+        id: release-please
+        with:
+          target-branch: ${{ github.ref_name }}
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
+      # These run only if a release PR was opened or modified, so not when the PR is merged
```
HarrisonWAffel, line 31: I'm a bit confused how this works for PRs, as the workflow only specifies `on.push.branches`. Is this something special that `release-please` can do?

HarrisonWAffel (changes requested):
I have some comments / suggestions. Generally excited to see this land, should do a lot of the heavy lifting for this repo!

.github/workflows/backport-issues.yml
matttrach: Upcoming releases still need a backport, all releases are backported now.
That was one of the process changes I made last month to enable this automation.

.github/workflows/backport-issues.yml
matttrach: The backport needs to be managed by a maintainer. My thought was that the first person to pick it up could remove the others. I can move it to an environment variable, or I could just assign it to me. I really wanted to be able to assign it to a group, but apparently that is not possible in GitHub yet.

matttrach: The tracking issue issue is also assigned to all three of us using the same logic. I can work on trying to get the assignees from the initial PR at the time of merge, but that is going to take some time because I have to find the PR from the tracking issue then get the assignees from the PR. I can't set the assignees from the PR to the tracking issue because most likely the person creating the PR won't assign anyone and the tracking issue is generated when the PR is created (rather than merged).

matttrach: I can make it just me that gets automatically assigned, but that might make it more difficult when I have to take time off.

.github/workflows/backport-issues.yml
matttrach: I also need to change this from 'version' to 'release', I forgot they need to match the branch name.

.github/workflows/backport-issues.yml
matttrach: Checking out the repository is considered a security issue in a workflow triggered in this way because the default GITHUB_TOKEN has elevated permissions.

matttrach: I will add a warning in a try..catch, but I think the workflow should fail if either parts of it don't happen.

matttrach: I definitely agree that having the scripts separated would be an improvement, but the elevated permissions along with the ability to trigger this workflow on label create makes that a security risk.

.github/workflows/backport-pr-manual.yml
matttrach: yep, thanks!

.github/workflows/backport-pr-manual.yml
matttrach: OK, so now it waits 10 seconds, then tries to look up the PR before moving on. This should include a backoff built into the github-script's API command.

.github/workflows/release.yml
matttrach: kind of, the key is the output along with the if statements:
see `if: steps.release-please.outputs.pr`

.github/workflows/release.yml
matttrach: thank you! I went ahead and corrected all 'uses' that don't use a commit id for the version.

.goreleaser.yml
matttrach: The yaml in this is very specific to goreleaser, I tried adding this, but it failed to validate:
``` ⨯ command failed                                  
    error=
    │ yaml: unmarshal errors:
    │   line 3: field extra_files not found in type config.Project```

File: .goreleaser_rc.yml (hunk 11 above)
Comment (matttrach) [reply .goreleaser_rc.yml:36]: this is not a valid option for goreleaser

File: run_tests.sh (hunk 14 above)
Comment (matttrach) [reply run_tests.sh]: It's not exported at the global level, I can make it local.

File: .github/workflows/backport-issues.yml (hunk 1 above)
Comment (matttrach) [reply .github/workflows/backport-issues.yml]: I moved everything to a repo variable and set the variable to only me rather than all of us.

File: .github/workflows/backport-issues.yml (hunk 2 above)
Comment (matttrach) [reply .github/workflows/backport-issues.yml]: I have updated all API calls to fall inside a try..catch with explicit error messages on failure.

File: .github/workflows/backport-pr-manual.yml (hunk 3 above)
Comment (matttrach) [reply .github/workflows/backport-pr-manual.yml]: yes, moved to a repo variable

File: .github/workflows/backport-prs.yml (hunk 4 above)
Comment (matttrach) [reply .github/workflows/backport-prs.yml]: moved to repo variable

File: .github/workflows/main-issue.yml (hunk 5 above)
Comment (matttrach) [reply .github/workflows/main-issue.yml]: yes, moved to a repo variable and shortened to only myself

File: test/one/basic_test.go (hunk 15 above)
Comment (matttrach) [reply test/one/basic_test.go]: Done, this was a pretty big change, but I think it simplified things a lot, thank you!

File: test/one/basic_test.go (hunk 16 above)
Comment (matttrach) [reply test/one/basic_test.go]: done, moved all teardown calls into defer statements

File: test/summarize.sh (hunk 17 above)
Comment (matttrach) [reply test/summarize.sh:3]: done

File: test/summarize.sh (hunk 18 above)
Comment (matttrach) [reply test/summarize.sh]: done

File: .github/workflows/backport-issues.yml [hunk 23]
```diff
@@ -0,0 +1,100 @@
...
+                repo: repo,
+                issue_number: parentIssueNumber,
+                sub_issue_id: subIssueId,
+                headers: {
+                  'X-GitHub-Api-Version': '2022-11-28'
+                }
+              });
+            } catch (error) {
+              core.setFailed(`Failed to link backport issue to main issue: ${error.message}`);
+            }
```
jiaqiluo, line 100: Nit:
While this is correctly wrapped in a try-catch block, using core.setFailed() will cause the entire workflow run to be marked as a failure if the linking API call fails. However, the primary goal, creating the backport issue, has already succeeded.
A more resilient approach would be to log a warning instead of failing the job. This ensures the workflow is still considered successful while alerting you that the sub-issue needs to be linked manually.

.github/workflows/backport-pr-manual.yml
```diff
@@ -0,0 +1,171 @@
...
+    steps:
+      - name: 'Wait for merge to settle'
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0 https://github.com/actions/github-script
+        env:
+          TERRAFORM_MAINTAINERS: ${{ env.TERRAFORM_MAINTAINERS }}
+        with:
+          script: |
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            const mergeCommitSha = context.payload.head_commit.id;
```
jiaqiluo: `context.payload.head_commit.id` is only available on push events, not workflow_dispatch events.
ref: 
- https://docs.github.com/en/webhooks/webhook-events-and-payloads#:~:text=the%20ref.-,head_commit,-object%20or%20null

.github/workflows/release.yml
```diff
@@ -0,0 +1,302 @@
...
+          LATEST_PATCH_NUM=$(git tag | grep "^${BASE_VERSION}\.${LATEST_MINOR_NUM}\." | awk -F- '{print $1}' | awk -F. '{print $3}' | sort -n | tail -1)
+          echo "Latest patch num: $LATEST_PATCH_NUM"
+          LATEST_RC_NUM=$(git tag | grep "^${BASE_VERSION}\.${LATEST_MINOR_NUM}\.${LATEST_PATCH_NUM}-rc\." | awk -F. '{print $4}' | grep -v '^$' | sort -n | tail -1)
+          echo "Latest rc num: $LATEST_RC_NUM"
+          if [ -z "$LATEST_RC_NUM" ]; then
+            # the last minor was a full release, we need to start the next minor release candidate
+            # this sort of assumes there wont be a patch release
+            LATEST_MINOR_NUM=$((LATEST_MINOR_NUM + 1))
+            LATEST_PATCH_NUM=0
+            NEXT_RC_NUM=0
```
jiaqiluo, line 200: When a new RC cycle starts (i.e., no previous RC exists for the latest patch), the script sets NEXT_RC_NUM=0. This will result in tags like vX.Y.Z-rc.0. Is it expected? Should it be changed to NEXT_RC_NUM=1?

jiaqiluo (changes requested):

.github/workflows/release.yml
matttrach: vX.Y.Z-rc.0 is expected

.github/workflows/backport-issues.yml
matttrach: I would prefer that the workflow fail so that I know to look into it and see the problem if either the creation or the link fails. If it doesn't fail I might miss that the issue was created and manually create one anyways.

.github/workflows/release.yml
HarrisonWAffel: Oh interesting, this is a cool action!

.github/workflows/release.yml
```diff
@@ -0,0 +1,302 @@
...
+              core.setFailed(`Failed to create comment failure on release PR: ${error.message}`);
+            }
+
+  rc-release:
+    needs:
+      - release
+      - test
+      - cleanup
+      - report
+    if: always() && needs.release.outputs.release_pr && (needs.test.result == 'success') #Ensure the test jobs succeeded, and that a release PR was created.
```
HarrisonWAffel, line 178: Is `always()` needed here? It makes sense for the `report` and `cleanup` steps, since they should run even if the test fails.

.github/workflows/backport-pr-manual.yml
matttrach: Good catch! Fixed

.github/workflows/backport-issues.yml
jiaqiluo: Gotcha, SGTM.

.github/workflows/backport-prs.yml
```diff
@@ -0,0 +1,169 @@
...
+                continue;
+              }
+              const targetBranch = releaseLabel.name
+              core.info(`Processing sub-issue #${subIssueNumber} for target branch: ${targetBranch}`);
+              const newBranchName = `backport-${pr.number}-${targetBranch.replace(/\//g, '-')}`;
+              try {
+                execSync(`git config user.name "github-actions[bot]"`);
+                execSync(`git config user.email "github-actions[bot]@users.noreply.github.com"`);
+                execSync(`git fetch origin ${targetBranch}`);
+                execSync(`git checkout -b ${newBranchName} origin/${targetBranch}`);
```
HarrisonWAffel, line 130: Sorry if I'm missing this somewhere, but is there a process to clean up these branches when the PR gets merged? Or does that just happen automatically (i seem to recall that being a default setting somewhere in github, but I could be wrong)

jiaqiluo (approved):

test/three/basic_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	util "github.com/rancher/terraform-provider-rancher2/test"
+)
+
+func TestThreeBasic(t *testing.T) {
+	t.Parallel()
+	config := util.NewTestConfig(t, "use-cases/three")
+
+	defer config.Teardown(t)
```
HarrisonWAffel, line 14: IIUC `Teardown` will delete the directory where the kubeconfig is stored, right? If so then we should get the logs first, and then tear it all down.

```suggestion
	defer func() {
	    config.GetErrorLogs(t)
	    config.Teardown(t)
	}()	
```

HarrisonWAffel (commented):
Just a few comments / questions / suggestions. Thanks for updating the issue assignee behavior. I wonder if there is some middle ground with assigning us all to new root issues, or another approach that will help me and Jack pick up any slack when you're away.

.github/workflows/backport-prs.yml
matttrach: yes, it gets automatically cleaned up by GitHub

.github/workflows/release.yml
matttrach: My thought was that if the release and test passes but the report or cleanup fails it will still generate the rc release.

test/three/basic_test.go
matttrach: I had to look this up, defer happens in LIFO order, so deferring the logs after the teardown will get the logs first.

matttrach: https://go.dev/tour/flowcontrol/13

.github/workflows/release.yml
HarrisonWAffel: Oh I see, okay cool

test/three/basic_test.go
HarrisonWAffel: TIL! In the future it may be better to defer an anonymous function so that the execution order is more clear, but as long as this works as expected I think its fine for now

HarrisonWAffel (approved):

Issue #1688, opened 2025-10-14
[BUG] `rancher2_user` resource hits a timeout on the latest `head` commits on multiple Rancher2 provider versions


markusewalker:
### Environment Information

- Provider version: `13.0.0-rc3`
- Rancher version: `head` https://github.com/rancher/rancher/commit/6886e3d097a2143bb15eb466a583ab43cd8928f4

### Describe the bug
When attempting to create a user with the `rancher2_user` resource in the specified head commit, I get the following:

```
ancher2_user.auto-testuser-orvba: Still creating... [03m20s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [03m30s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [03m40s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [03m50s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m00s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m10s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m20s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m30s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m40s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [04m50s elapsed]
rancher2_user.auto-testuser-orvba: Still creating... [05m00s elapsed]
╷
│ Error: [ERROR] waiting for user (u-8l54r) to be created: timeout while waiting for state to become 'active' (timeout: 5m0s)
│ 
│   with rancher2_user.auto-testuser-orvba,
│   on main.tf line 131, in resource "rancher2_user" "auto-testuser-orvba":
│  131: resource "rancher2_user" "auto-testuser-orvba" {
│ 
```
I am seeing this on Rancher2 provider versions `13.0.0-rc3` and `8.1.0`. Additionally, when testing on the latest commits for `v2.12-head` and `v2.11-head`, I am NOT seeing this behavior. So, this is specific to the latest `head` commits and happens with multiple Rancher2 provider versions.

### To Reproduce
1. Setup Rancher with latest `head` commit.
2. Try to create a Rancher user with the `rancher2_user`. See my `main.tf` below:
```
terraform {
  required_providers {
    # rancher2 = {
    #   source  = "terraform.local/local/rancher2"
    #   version = "13.0.0-rc3"
    # }
    rancher2 = {
[...]

PR #1693, opened 2025-10-23
Remove broken resources

bigkevmcd:
## Problem
Rancher dropped support for RKE1 in June 2025 - the terraform provider still provides resources for these  - they will not work with Rancher 2.13. 
 
Bumping the provider to use a newer version of Rancher fails to compile because the structs that these resources depend on have been removed.

The v3.User struct also dropped the `.State` field which the `rancher2_user` resource was relying on, this has been dropped.

This is the bug reported in #1688

## Solution
This is a bare minimum removal, the minimal amount of code to get it to compile with the latest rancher is removed.


rancher2/resource_rancher2_user.go
```diff
@@ -191,6 +191,6 @@ func userStateRefreshFunc(client *managementClient.Client, userID string) resour
 			}
 			return nil, "", err
 		}
-		return obj, obj.State, nil
+		return obj, "active", nil
```
vardhaman22, line 196: seems to me that it's not related to rke1 removal.
maybe i am wrong here but in case it's not related to rke1 removal it would be good to add reason for this in the PR description.

vardhaman22 (approved):
just one question.
overall LGTM.

rancher2/resource_rancher2_user.go
bigkevmcd: Oops...lemme remove that, that's for https://github.com/rancher/terraform-provider-rancher2/issues/1688

bigkevmcd: Ahh...of course, I can't remove that because the user dropped `.State` too.

vardhaman22: cool seems fine.

vardhaman22 (approved):

matttrach (approved):

PR #1699, opened 2025-10-31
Don't check user.State on creation

bigkevmcd:
Addresses #1688 

## Description
This changes the `userStateRefreshFunc` to not use the `.State` field that was removed.

If the user is created, then it's assumed to be `active`.

This has been removed for 2.13 and so it will never complete.

When rancher/rancher is refreshed, we can look in the Status for conditions to find out if the user is active, but as the code pulls in the rancher structs from before the public-api changes this won't work without a significant upgrade.

## Testing

I've created users and they complete.

Not a breaking change.


raulcabello (approved):
LGTM, thanks!

pmatseykanets (approved):

matttrach (approved):

PR #1702, opened 2025-10-31
fix: export github scripts and validate in CI

matttrach:
## Description

The automated backports and issue generation is failing due to variable scope changes.
- This extracts the scripts into their own file so that we can run node check and eslint validations against them before they are merged.
- It also corrects any issues that were found, including doublechecking any variable scope or type (const vs let) issues.
- I simplified the scripts by removing deconstruction because it was easier to understand.
- I updated the issue templates to simplify them, I would prefer users don't try to diagnose the type of issue (bug, feature, etc) and just explain what is causing them pain. 

## Testing

Added node check and eslint checks to validate the scripts in the CI.

Not a breaking change.


matttrach:
"Create Main Issue" is still using the old code, not what is in the PR.

jiaqiluo (approved):

Issue #1807, opened 2025-11-14
[PROBLEM] Can not drop the `rancher.io/imported-cluster-version-management` annotation from `rancher2_cluster` resource


jiaqiluo:
### Environment Information

- Terraform version: v1.13.5
- Rancher2 Provider version: v8.3.1 
- Rancher version: v2.11.0 and above 
- Operating System of the Rancher Node: N/A
- Operating System of the machine running Terraform: N/A
- Kubernetes Distro (k3s/rke2): Both 
- Kubernetes Version: N/A
- Infrastructure Provider (AWS/GCP/vSphere/etc): N/A 
- What is the scope of the token given to the provider? Admin 

### Describe the Problem

The version management feature for imported RKE2 and K3s clusters was added in Rancher 2.11.0.   
We can configure the feature by setting the `rancher.io/imported-cluster-version-management` annotation as in the example below: 

```
resource "rancher2_cluster" "my-imported-cluster-tf" {
  name = "my-imported-cluster-tf"
  annotations = {
    "rancher.io/imported-cluster-version-management" = "false" 
    "another.annotation/key"  = "another-value"
  }
}

output "kubectl-command" {
  value = ["${rancher2_cluster.my-imported-cluster-tf.cluster_registration_token.0.command}"]
}

output "insecure-kubectl-command" {
  value = ["${rancher2_cluster.my-imported-cluster-tf.cluster_registration_token.0.insecure_command}"]
}
```

**The problem is that we could not drop the annotation once it is set.** 
In the scenario where a user sets the annotation when creating the cluster and later removes it, the user might assume that the version-management feature is no longer under their control and that the default behavior has taken over, but in reality, that’s not the case.

Steps:
- Use the above example to create an imported cluster
- Register a standalone k3s cluster by using the registration command 
- Wait for the cluster to be active in Rancher 
- Edit the Terraform file to drop the annotation, i.e. the `"rancher.io/imported-cluster-version-management" = "false"` line 
- Run `terraform apply` 

Results:

- Terraform reports there is no change detected 

```
rancher2_cluster.my-imported-cluster-tf: Refreshing state... [id=c-zcwc9]
No changes. Your infrastructure matches the configuration.
Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
...
```

- On the rancher side, the "rancher.io/imported-cluster-version-management" annotation still exists: 

[...]

PR #1808, opened 2025-11-14
fix: support exceptions in diff suppression

jiaqiluo:
Addresses  https://github.com/rancher/terraform-provider-rancher2/issues/1807 

## Description

Please see https://github.com/rancher/terraform-provider-rancher2/issues/1807 for details. 

This PR adds support for exceptions when computing the diff of Rancher-managed labels and annotations. 

## Testing

Following the same steps provided in the issue, I can confirm that Terraform detects the removal of the annotation, and after applying the changes, the annotation on the cluster is set to its default value by rancher-webhook. 

Output: 
```

  # rancher2_cluster.my-imported-cluster-tf will be updated in-place
  ~ resource "rancher2_cluster" "my-imported-cluster-tf" {
      ~ annotations                = {
          - "rancher.io/imported-cluster-version-management"              = "false" -> null
            # (9 unchanged elements hidden)
        }
        id                         = "c-zcwc9"
        name                       = "jiaqi-imported-cluster-tf"
        # (13 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
...
```

Rancher side: 
Screenshot 2025-11-14 at 11 55 09 AM

### DIFF
--- a/rancher2/schema_common.go
+++ b/rancher2/schema_common.go
@@ -16,18 +16,20 @@ var exceptions = []string{
 	"rancher.io/imported-cluster-version-management",
 }
 
-// supressFunc is a DiffSuppressFunc that prevents Terraform from trying to remove Rancher-managed annotations/labels;
-// it ignores the annotation from a predefined list of annotation/label keys.
-var supressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+// suppressFunc is a DiffSuppressFunc that prevents Terraform from removing Rancher-managed annotations or labels.
+// It also ignores changes to a predefined set of Rancher-managed annotation/label keys.
+//
+// Thought it is not recommended, users can still add annotations or labels whose keys contain Rancher-managed keys,
+// but they won't be able to remove them once added.
+// Note: Terraform prefixes the key `k` with either "annotations." or "labels."
+var suppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
 	for _, exception := range exceptions {
-		// The key `k` is prefixed with "annotations." or "labels."
-		// suppress the diff if the key contains the exception
-		if strings.Contains(k, exception) {
+		// Explicitly check if the key ends with the exception
+		if strings.HasSuffix(k, exception) {
 			return false
 		}
 	}
 
-	// Suppress the diff for Rancher-managed annotations/labels
 	if (strings.Contains(k, commonAnnotationLabelCattle) || strings.Contains(k, commonAnnotationLabelRancher)) && new == "" {
 		return true
 	}
@@ -44,14 +46,14 @@ func commonAnnotationLabelFields() map[string]*schema.Schema {
 			Optional:         true,
 			Computed:         true,
 			Description:      "Annotations of the resource",
-			DiffSuppressFunc: supressFunc,
+			DiffSuppressFunc: suppressFunc,
 		},
 		"labels": {
 			Type:             schema.TypeMap,
 			Optional:         true,
 			Computed:         true,
 			Description:      "Labels of the resource",
-			DiffSuppressFunc: supressFunc,
+			DiffSuppressFunc: suppressFunc,
 		},
 	}
 	return s
--- a/rancher2/schema_common.go
+++ b/rancher2/schema_common.go
@@ -11,35 +11,47 @@ const (
 	commonAnnotationLabelRancher = "rancher.io/"
 )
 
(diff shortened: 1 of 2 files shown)


HarrisonWAffel (approved):

rancher2/schema_common.go
```diff
@@ -11,35 +11,47 @@ const (
...
+	"rancher.io/imported-cluster-version-management",
+}
+
+// supressFunc is a DiffSuppressFunc that prevents Terraform from trying to remove Rancher-managed annotations/labels;
+// it ignores the annotation from a predefined list of annotation/label keys.
+var supressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+	for _, exception := range exceptions {
+		// The key `k` is prefixed with "annotations." or "labels."
+		// suppress the diff if the key contains the exception
+		if strings.Contains(k, exception) {
```
jferrazbr: I think `strings.Contains` may return false positives in this case (e.g., `annotations.rancher.io/imported-cluster-version-management-some-suffix` would match). Would it be possible to switch to an exact key match instead?

rancher2/schema_common.go
```diff
@@ -11,35 +11,47 @@ const (
...
 )
 
+// exceptions is a list of annotation/label keys that should not be suppressed.
+var exceptions = []string{
+	"rancher.io/imported-cluster-version-management",
+}
+
+// supressFunc is a DiffSuppressFunc that prevents Terraform from trying to remove Rancher-managed annotations/labels;
+// it ignores the annotation from a predefined list of annotation/label keys.
+var supressFunc = func(k, old, new string, d *schema.ResourceData) bool {
```
jferrazbr: Tiny nit: supress -> suppress (PS: I hate english words with double consonants 😅 "process" makes me insane I always write it with 2 "c"s)

rancher2/schema_common.go
jiaqiluo: If a user uses a key like `annotations.rancher.io/imported-cluster-version-management-some-suffix` and updates its value, what should the TF Rancher provider do? My understanding is that it should detect the change and report it back to the user—correct? If so, then the current implementation is working as intended.

Also, I don’t think we’ll reuse the `rancher.io/imported-cluster-version-management` as a prefix for other features or functionality in Rancher, so it shouldn’t come back to bite us. :) 

WDYT?

rancher2/schema_common.go
jiaqiluo: Good catch. Fixed

rancher2/schema_common.go
jferrazbr: Makes sense :+1: I totally agree with your perspective. Showing diffs for any derived key (e.g., ...-some-suffix) is definitely the right user experience. To make this behavior more explicit what do you think of adding a test for suffixed key -> no suppression and a doc note that the exception check intentionally matches substrings?

jiaqiluo: Updated this part of the logic to check for suffixes instead. 
Based on my testing, users can still add annotations or labels whose keys contain Rancher-managed patterns, but once added, they can’t remove them. 
Because of this, keys like `annotations.rancher.io/imported-cluster-version-management-some-suffix` should be handled the same way as others, such as `authz.management.cattle.io/initial-sync-some-suffix`. 

New test cases were also added.

rancher2/schema_common.go
```diff
@@ -11,35 +11,50 @@ const (
...
+
+// suppressFunc is a DiffSuppressFunc that prevents Terraform from removing Rancher-managed annotations or labels.
+// It also ignores changes to a predefined set of Rancher-managed annotation/label keys.
+//
+// Thought it is not recommended, users can still add annotations or labels whose keys contain Rancher-managed keys,
+// but they won't be able to remove them once added.
+// Note: Terraform prefixes the key `k` with either "annotations." or "labels."
+var suppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+	for _, exception := range exceptions {
+		// suppress the diff if the key ends with the exception
```
jferrazbr: just a nit, I think here is "not suppress", right?

jferrazbr (approved):

rancher2/schema_common.go
jiaqiluo: Fixed with better wording.

jferrazbr (approved):

matttrach (approved):

PR #1846, opened 2025-11-26
[Feature] Add support to import machine configs v2

davidhrbac:
## Description

As we can import clusters it would be nice to have the ability to import machine configs v2 too.

## Testing
I was able to generate import file of my machine config successfully like this:

```bash
❯ cat  import.tf
import {
  to = rancher2_machine_config_v2.nc_test
  id = "fleet-default/nc-test-pool1-zs2g8"
}

❯ terraform plan -generate-config-out=generated.tf
Found existing alias for "terraform plan". You should use: "tfp"
Acquiring state lock. This may take a few moments...
helm_release.rancher: Refreshing state... [id=rancher]
rancher2_machine_config_v2.nc_test: Preparing import... [id=fleet-default/nc-test-pool1-zs2g8]
rancher2_machine_config_v2.nc_test_pool1: Refreshing state... [id=fleet-default/nc-test-pool1-qnqw8]
rancher2_cluster_v2.db-operations-sand: Refreshing state... [id=fleet-default/db-operations-sand]
rancher2_cluster.local: Refreshing state... [id=local]
rancher2_cluster_v2.test: Refreshing state... [id=fleet-default/test]
rancher2_machine_config_v2.nc_test: Refreshing state... [id=fleet-default/nc-test-pool1-zs2g8]

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Warning: Config generation is experimental
│
│ Generating configuration during import is currently experimental, and the generated configuration format may change in future versions.
╵
╷
│ Error: "generate_name": required field is not set
│
│   with rancher2_machine_config_v2.nc_test,
│   on generated.tf line 1:
│   (source code not available)
│
╵
```

Generated TF file contains `generate_name=null` which must be corrected manually. That's not the deal-breaker as we see this behaviour during the import of others resources.


matttrach (approved):

matttrach:
Hello @davidhrbac, thanks for the contribution! We need one more review approval before we can merge.

HarrisonWAffel (commented):
I suppose I'm not totally against this change if you find value in it, but what is the specific use case for this? Rancher doesn't intend for a machine config to exist without an associated cluster, and [will remove orphaned configs automatically using a cron job](https://github.com/rancher/rancher/blob/main/pkg/controllers/provisioningv2/machineconfigcleanup/controller.go#L144). 

If the goal with this is to create a standalone machine config object that can be reused for multiple clusters, it will be removed if no cluster references it for a significant amount of time. The expectation is that a machine config will be associated with a single cluster at a time. Unfortunately, machine configs are not like legacy node templates.

let me know if I'm misunderstanding your use case, thanks for raising the PR!

davidhrbac:
@HarrisonWAffel thanks for the clarification.

Yes, I’m aware that machine configs are expected to be cluster-bound and that orphaned configs are cleaned up automatically. My primary intention was slightly different.

What I wanted was the ability to **import the downstream machine config (especially the cloud-init section) from production**, at least as a reference in Terraform. That’s why I implemented the import functionality. Being able to track it with Terraform would be a bonus.

In practice, I ended up defining the machine config like this:

```hcl
resource "rancher2_machine_config_v2" "example_pool1" {
  generate_name = "example-pool1"

  vsphere_config {
    clone_from                = "/<DATACENTER>/vm/<TEMPLATE_FOLDER>/<TEMPLATE_NAME>"
    creation_type             = "template"
    cpu_count                 = 4
    memory_size               = 16384
    disk_size                 = 200000

    datacenter                = "/<DATACENTER>"
    datastore                 = "/<DATACENTER>/datastore/<DATASTORE_NAME>"
    folder                    = "/<DATACENTER>/vm/<FOLDER_PATH>"
    hostsystem                = "/<DATACENTER>/host/<CLUSTER_NAME>/<HOSTNAME>"
    pool                      = "/<DATACENTER>/host/<CLUSTER_NAME>/Resources/<RESOURCE_POOL>"
    network                   = ["DistributedVirtualPortgroup:<PORTGROUP_ID>"]

    os                        = "linux"
    ssh_user                  = "ssh_user"
    ssh_password              = "<REDACTED>"
    ssh_port                  = 22
    ssh_user_group            = "staff"

    vcenter_port              = 443
    cfgparam                  = ["disk.enableUUID=TRUE"]
    tags                      = ["<REDACTED_TAG_URN>"]
    graceful_shutdown_timeout = "0"

    cloud_config = <<EOF
#cloud-config
growpart:
  mode: auto
  devices:
    - '/'
    - '/dev/sda2'
  ignore_growroot_disabled: false

bootcmd:
  - nmcli con mod "System ens192" ipv4.dhcp-vendor-class-identifier example
  - nmcli con mod "ens192" ipv4.dhcp-vendor-class-identifier example
  - rm -f /var/lib/NetworkManager/*.lease
  - systemctl restart NetworkManager

runcmd:
  - dnf -y install dnf-automatic yum-utils
  - sed -i 's/^upgrade_type.*$/upgrade_type = default/' /etc/dnf/automatic.conf
  - sed -i 's/^apply_updates.*$/apply_updates = yes/' /etc/dnf/automatic.conf
  - sed -i 's/^download_updates.*$/download_updates = yes/' /etc/dnf/automatic.conf
  - systemctl enable --now dnf-automatic.timer
EOF
  }
}
```

And then referencing it from the cluster:

```hcl
resource "rancher2_cluster_v2" "test" {
  ...
  name = "test"
  ...
    machine_pools {
      annotations                    = {}
      cloud_credential_secret_name   = null
      control_plane_role             = true
      drain_before_delete            = true
      etcd_role                      = true
      labels                         = {}
      machine_labels                 = {}
      max_unhealthy                  = null
      name                           = "pool1"
      node_drain_timeout             = 0
      node_startup_timeout_seconds   = 0
      paused                         = false
      quantity                       = 3
      unhealthy_node_timeout_seconds = 3600
      unhealthy_range                = null
      worker_role                    = true
      machine_config {
        api_version = null
        kind        = "VmwarevsphereConfig"
        name        = rancher2_machine_config_v2.nc_test_pool1.name
      }
    }
  ...
}
```

What surprised me was that after importing the existing machine config, I was actually able to modify it via Terraform. That resulted in a new MachineSet being created and the downstream cluster being redeployed.

I’m not claiming this is the intended or best workflow. My goal was mainly to:

1. Extract the effective machine configuration from a running cluster.
2. Bring that state under Terraform management.
3. Use it as a controlled baseline moving forward.

If there’s a more appropriate or recommended way to achieve this (for example, treating machine configs differently in Terraform, or managing only specific parts), I’d definitely appreciate guidance.

The import functionality is helpful because it allows bringing live configuration into Terraform, but I’m open to suggestions on how this should ideally be handled.

HarrisonWAffel:
> Use it as a controlled baseline moving forward.

It sounds like you want to basically create a retroactive node template, which is an interesting idea. But I'm concerned by this approach, especially because you state 

> What surprised me was that after importing the existing machine config, I was actually able to modify it via Terraform. That resulted in a new MachineSet being created and the downstream cluster being redeployed.

Importing a resource means taking over management of it, including making changes to it. If you're importing a configuration from a production cluster this behavior could cause a potential production outage if you accidentally change something in your test cluster, or at least unneeded production node churn. While retrieving information like this is convenient, it seems dangerous - especially because it will not be clear what other clusters are referencing the imported machine config. 

Are you managing all of your resources using terraform? If so, it would be safer to create an intermediate resource like a config-map and simply reference the keys in it when creating the two clusters using the kubernetes provider. 

It sounds like it may be a better idea to create data sources for the configs so that they can be used in a read-only way, but I'm hesitant about that as since it may imply a config can be used across multiple clusters, when it should not.

davidhrbac:
Yes, we manage our environments with Terraform quite extensively:

* Local cluster Kubernetes upgrades
* Downstream cluster Kubernetes upgrades
* Node scaling when using platform integrations (typically vSphere)
* Rancher upgrades

The goal of this import PR is to help us **control, audit, and automate more parts of the environment** through Terraform.

Regarding your concern:

> It sounds like it may be a better idea to create data sources for the configs...

We do **not** intend to share machine configurations across multiple clusters. Our model is strictly isolated:

* One Terraform instance per Rancher environment
* Each cluster has its own explicit resource definition
* No cross-cluster config reuse

The intent behind import is not to treat machine configs as reusable templates (like legacy node templates), but rather to:

1. Capture the effective configuration from a running environment
2. Bring it under Terraform state management
3. Make future changes explicit, auditable, and reproducible

We fully respect the design expectation that a machine config belongs to a single cluster.

Our main objective is governance and lifecycle control, not template sharing.

rancher2/import_rancher2_machine_config_v2.go
```diff
@@ -0,0 +1,52 @@
...
+	if id == "" {
+		return []*schema.ResourceData{}, fmt.Errorf("Importing rancher2_machine_config_v2 requires a non-empty ID")
+	}
+
+	cfg, ok := meta.(*Config)
+	if !ok || cfg == nil {
+		return []*schema.ResourceData{}, fmt.Errorf("Importing rancher2_machine_config_v2: provider config is nil")
+	}
+
+	// Available kinds - would be good to have it not hardcoded here....
```
HarrisonWAffel: I think hardcoded is fine for now, there aren't plans to add new default node drivers to Rancher at the moment, and this covers everything we currently support.

davidhrbac:
One thing I forgot to mention: we’re also planning to build a “semi-automatic” loop around this using a small tool I wrote:

https://gitlab.com/davidhrbac/rancher-inventory

The idea is to use it as an inventory/status layer to help drive controlled Terraform updates — mainly for Rancher and Kubernetes upgrades. Not full automation, but assisted visibility + version tracking feeding into TF changes.

So the import capability fits into a broader goal of gradually bringing more of the environment under declarative and auditable control.

HarrisonWAffel:
Thanks for explaining your use case more! Sorry for any misunderstanding on my part. Sounds like you're initially provisioning in the UI (or maybe yaml?).

I brought this up to our team as well as tested it locally. In my tests I found that it was not possible to import the resource using your changes without that resource later being deleted and recreated by a `terraform apply` call. This is because the norman API returns `null` for `generate_name`, but the structure [defined within the provider specifies `ForceNew` for that field](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/schema_machine_config_v2.go#L25). Trying to set the initially null `generate_name` to the same value used in Rancher will actually result in a recreation, and leaving it null will prevent terraform apply from working at all ([as it is a required field](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/schema_machine_config_v2.go#L24)).

This can result in previously provisioned clusters which referenced that machine config to go into an error state, as the config reference will no longer be valid unless it is manually updated. I recognize that you're manually addressing that in the examples that you've posted, but this may be surprising to other users who try to import this resource and expect it to "just work".

Ultimately there are a lot of limitations in the provider due to the fact that it still relies on the Norman API, whereas the Rancher Steve API is used elsewhere (like the UI). After discussion, we agree that this would be useful but due to the current quirks we should hold off on adding it until we transition the provider to the Steve API. This is currently being handled in https://github.com/rancher/terraform-provider-rancher2/pull/1940. 

I'm gonna close this issue based off of todays discussion, but this is definitely a good idea and something we would like to add in the future once we work out some of the kinks in the provider. We will keep https://github.com/rancher/terraform-provider-rancher2/issues/1847 open to track this work so that it gets prioritized in the future.

Issue #1638, opened 2025-09-03
[Feature] Upgrade to the Terraform framework


matttrach:
Before I start generating documentation I want to have a working example that closely reflects how our team manages issues and releases.

I am writing a new provider to show a working example of the TF framework, this should give everyone a quick reference to the most common uses. I want to show how to unit test, isolate resources for testing documentation examples, and how to write e2e tests with our test framework.

I also want to use this as a proving ground for release automation, I want the new provider's releases to work seamlessly before implementing them in the Rancher provider.

We can also experiment with converting, can the SDKv1 live in the same package as the TF framework? Can we convert resources one by one, or do we need to convert everything all at once? As we introduce new resources can we test and document them without disturbing the current system? We can test this all out in the new provider.

I need the new "File" provider for use in my modules so it has clear support boundary and use cases.

matttrach:
The SDKv1 doesn't play well with the framework. There is an underlying protocol version.
Our provider protocol is version 4. When converting to protocol version 5 the import namespace changes in a breaking fashion. 

- the SDKv1 import namespace (proto v4) is `github.com/hashicorp/terraform-plugin-sdk/plugin`
- the SDKv2 import namespace (proto v5) is `github.com/hashicorp/terraform-plugin-sdk/plugin/v2/`

The framework uses protocol version 6. Hashicorp built a mechanism for converting from proto v5 to proto v6 cleanly without needing to convert all resources at once, but they didn't do that for proto v4 to proto v5

matttrach:
There is a few standard practices and a guide for converting from SDKv1 to SDKv2, then there is a mechanism where SDKv2 and the framework can exist in the same package. 

Just to reiterate, SDKv1 resources and SDKv2 resources can't exist in the same package.

matttrach:
There are two routes we can take to move to the framework, we can either convert everything to SDKv2 and slowly phase out resources using the SDKv2 for newer resources using the framework, or we can convert everything directly to the framework at once. 
I am leaning towards moving everything to the framework, I think the resulting codebase will be much cleaner and I will have the opportunity to make sure everything is well tested.

matttrach:
I am moving forward with the everything all at once approach, hopefully this will result in the fewest shocks for users. There will be a big push with a lot of changes, then everything will settle down and go back to business as usual.

matttrach:
These projects will be included in this change:
- Upgrade the provider protocol from v4 to v6
- Add unit tests for everything
[...]

PR #1940, opened 2025-12-05
feat: rewrite!

matttrach:
Addresses #1638

## Description
This is a **_DRAFT PR_**, it is in a very rough state, but I want to get the code out for everyone to see and make comments on as I write it. The goal is to be transparent with the process and allow users to voice their concerns and request changes. The unfortunate reality is we need to rewrite the provider, so lets make it better for everyone!

## Testing
I am a huge proponent of test driven development, every function that isn't a fundamental part of Terraform's framework integration will be tested. I will also be building end to end tests to validate major use cases. This is where your input becomes critical! I will cover the most basic use cases, but if you detail your use case in the comments I will try my best to work with you and get it into our tests. This will ensure that your needs are met before each release.

This is basically a rewrite of the provider as we currently have it.
The reasoning for this is detailed in #1638, but the tldr; is that we need to upgrade to the Terraform framework and this requires a breaking change.


matttrach:
I am a proponent of test driven development, so while starting work on the provider I want to build a baseline e2e test.
The first resource that anyone deploying Rancher with Terraform will implement is the rancher2_bootstrap resource.
My experience with this resource has not been good. I don't like the "magic", I don't like not knowing what it represents or what it does, I don't like that it doesn't work sometimes and it isn't clear why. I don't like having to implement the provider multiple times and juggling provider aliases. I don't like trying to figure out the dependency chain issues of two providers implemented in the same  apply. Reading the code, I think this resource can be split into multiple parts and either outright eliminated or at the very least it can be a module rather than a resource in this provider. I am also looking at the provider configuration and realizing the Rancher has several APIs that all exist at the same time. 

So I have several goals on the basic provider implementation:
1. make the provider more transparent and easier to troubleshoot
2. eliminate the need for provider aliases
3. manage API differences between resources and allow developers to control which API a resource uses without breaking user's config

matttrach:
The concept I have thought of is to have a rancher_client resource. This resource manages how the other resources connect to Rancher. Rather than configuring the client in the provider block, you can implement several clients each with potentially different information. On the back end, the provider will authenticate the client and keep a list of providers for other resources to use. Each other resource will have a required rancher_client_id that tells the resource which client to use. 
The rancher_client resource builds and authenticates a client for other resources, respecting the dependency chain. This eliminates the need for provider aliases while also exposing the client to users making things more transparent. It has the side effect of allowing users to use different users and authentication methods for different resources, enabling better security practices. On the backend, the client resource will allow other resources to choose which API they talk to as the API communication becomes a resource concern rather than a provider level concern.

matttrach:
- Do we use SDKs and do we enable multiple SDKs?
	- Strengths (pros)
		- rancher developers can use their native tooling to develop resources
		- differences between resource and SDK will become more apparent
		- if the SDK changes and the resource doesn't then the provider won't build (faster failing)
		- faster resource updates
			- since the resource will need to be updated in lockstep with Rancher
		- closer resemblance to Rancher
			- since we are using the same APIs
	- Weaknesses (cons)
		- There is a strong disconnect between the body of an API request and resource schemas
			- if resource schemas change arbitrarily (like due to reformatting) it breaks compatibility
			- troubleshooting becomes a nightmare of scouring multiple SDKs to find a problem
		- Provider maintainers will need to keep the SDKs up to date
			- provider maintainers will need to be aware of multiple SDK releases
			- if SDKs don't follow proper semver then this could have arbitrary breaking changes
			- if SDKs don't follow a release schedule there could be weird inconsistencies
			- SDKs are not Rancher
				- so they don't have to follow Rancher's release cycle
				- or backwards compatibility
		- Not only will we need to know about the SDKs themselves, but how Rancher uses them
			- although an SDK might have released a new version Rancher might not be using it yet
			- we need to follow how Rancher uses the SDK, not just the SDK itself
		- One SDK will govern many resources
			- upgrading one SDK could force breaking changes to many different resources
			- upgrading the SDK becomes a development challenge
	- Opportunities (possible strengths)
		- we may be able to use more of the code from the previous provider version
	- Threats (possible blockers)
		- I can't think of anything that would block this path
- What if we don't use SDKs?
	- Strengths
		- the provider transparently communicates with Rancher
			- this enables easier troubleshooting
			- data structures are easily exposed and bugs will be found faster
		- there is only one client
			- this significantly reduces the complexity of the provider
			- resources focus on http requests, not client configuration or internal pkg use
		- there are less dependencies
			- this means less opportunity for CVEs
			- less code overall to understand
			- less dependency management or risk of mismatch
	- Weaknesses
		- we have to build the data structures to the API specification
			- we will need to introduce any boilerplate code ourselves
			- we will need to know the actual API structure rather than the internal representations
				- slows down development
		- we lose out on any helpful things the SDK provides
			- things like composite functions that make multiple API calls to accomplish one goal
			- intricacies to the API that aren't communicated well, but are built into the SDK
	- Opportunities
		- resources have more control over how they interact with the API
			- they could add headers or attributes not sent in the SDK or required by the API
				- we could fix compliance issues on the fly
	- Threats
		- I can't think of anything that would block this path

matttrach:
Based on this, I am moving forward with no SDKs for the time being. This undermines the idea of a rancher_client resource. Instead we will have a rancher_login resource that updates the single client created at provider config time.

matttrach:
I don't like having this super long draft PR open. I would prefer to work on this in chunks and be able to review it as it progresses. I would also like to eventually release this version along with the older version for a while before deprecating the older version and merging this with the actual main.

To facilitate this I am thinking about creating a few new branches, `rancher:main-rewrite` as a place to integrate all of my changes without merging them to `rancher:main`. Then I can eventually have `rancher:release-rewrite/v1` to facilitate releases. I will need to rebase `rancher:main-rewrite` on `rancher:main` periodically so that the eventual merge is smooth, but this will essentially be like having a shadow repo within the repo with the goal of eventually merging them together.

Issue #1960, opened 2025-12-18
[BUG] Terraform Provider Cannot Read enabled value


dasarinaidu:
**Summary:**
The provider's Read function cannot parse enabled: true from Kubernetes. It always reads as enabled: false in Terraform state.

**Environment:**
Provider Version: v13.1.3-rc.3
Rancher Version: v2.13.0
Cluster ID: <downstream_cluster_id>

**Reproduction Steps:**
1. Setup
```
export TOKEN="your-token"
export RANCHER_TOKEN_KEY="$TOKEN"
```
2. Create main.tf

```
terraform {
  required_providers {
    rancher2 = {
      source = "rancher/rancher2"
    }
  }
}

provider "rancher2" {
  api_url  = "https://rancherURL"
  insecure = true
}

resource "rancher2_cluster_proxy_config_v2" "test" {
  cluster_id = "downstream_cluster_id"
  enabled    = true
}
```

**Test Case 1: Import with enabled=false**
# Set to false via kubectl
`kubectl patch clusterproxyconfig clusterproxyconfig -n c-m-kxmwspbg --type merge -p '{"enabled": false}'`

# Verify kubectl
`kubectl get clusterproxyconfig clusterproxyconfig -n c-m-kxmwspbg -o jsonpath='{.enabled}'`
# Output: false

# Import to Terraform
```
rm -f terraform.tfstate*
terraform import rancher2_cluster_proxy_config_v2.test "c-m-kxmwspbg/clusterproxyconfig"
```

# Check Terraform state
`terraform state show rancher2_cluster_proxy_config_v2.test | grep enabled`
# Output: enabled = false ✅ CORRECT

**Test Case 2: Import with enabled=true**

# Set to true via kubectl
`kubectl patch clusterproxyconfig clusterproxyconfig -n c-m-kxmwspbg --type merge -p '{"enabled": true}'`

# Verify kubectl
`kubectl get clusterproxyconfig clusterproxyconfig -n c-m-kxmwspbg -o jsonpath='{.enabled}'`
# Output: true

# Import to Terraform
rm -f terraform.tfstate*
terraform import rancher2_cluster_proxy_config_v2.test "c-m-kxmwspbg/clusterproxyconfig"

# Check Terraform state
terraform state show rancher2_cluster_proxy_config_v2.test | grep enabled
# Output: enabled = false ❌ WRONG (should be true)

**Expected Behavior**
[...]

PR #1983, opened 2026-01-06
Fix import of ClusterProxyConfigs

bigkevmcd:
Addresses #1960

## Description

This fixes a bug where the requested ID to import was not being used in the import process, which was causing an incorrect URL to be queried for and the desired resource not to be loaded correctly.

## Testing

Reproduced using the instructions in the bug and confirmed that after the fix the resource is loaded correctly.

Not a breaking change.


matttrach:
@bigkevmcd, this looks like it addresses an issue in release/v13, so I have added that one, please add labels for any other branches that this targets.

matttrach (approved):

Issue #2002, opened 2026-02-04
[PROBLEM] Unit tests Failing


matttrach:
### Environment Information

- Terraform version: n/a
- Rancher2 Provider version: v14
- Rancher version: n/a
- Operating System of the Rancher Node: n/a
- Operating System of the machine running Terraform: n/a
- Kubernetes Distro (k3s/rke2): n/a
- Kubernetes Version: n/a
- Infrastructure Provider (AWS/GCP/vSphere/etc): n/a
- What is the scope of the token given to the provider? n/a

### Describe the Problem

```
=== FAIL: rancher2 TestFlattenProject (0.00s)
panic: interface conversion: interface {} is map[string]interface {}, not string [recovered, repanicked]

goroutine 369 [running]:
testing.tRunner.func1.2({0x2bc8ac0, 0xc000de5710})
	/nix/store/0a3dyfq09dnkw28ap2i450wjimvdmv6s-go-1.25.4/share/go/src/testing/testing.go:1872 +0x237
testing.tRunner.func1()
	/nix/store/0a3dyfq09dnkw28ap2i450wjimvdmv6s-go-1.25.4/share/go/src/testing/testing.go:1875 +0x35b
panic({0x2bc8ac0?, 0xc000de5710?})
	/nix/store/0a3dyfq09dnkw28ap2i450wjimvdmv6s-go-1.25.4/share/go/src/runtime/panic.go:783 +0x132
github.com/stretchr/testify/assert.messageFromMsgAndArgs({0xc000da1ed0?, 0x30a24a0?, 0x1?})
	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.11.1/assert/assertions.go:316 +0xec
github.com/stretchr/testify/assert.Fail({0x36a65e0, 0xc0001fb180}, {0x31d8ebf, 0x3d}, {0xc000da1ed0, 0x2, 0x2})
	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.11.1/assert/assertions.go:382 +0x229
github.com/stretchr/testify/assert.FailNow({0x36a65e0, 0xc0001fb180}, {0x31d8ebf, 0x3d}, {0xc000da1ed0, 0x2, 0x2})
	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.11.1/assert/assertions.go:349 +0x8d
github.com/rancher/terraform-provider-rancher2/rancher2.TestFlattenProject(0xc0001fb180)
[...]

PR #2007, opened 2026-02-05
Fix broken tests in project structure

bigkevmcd:
Addresses: #2002 

## Description

This fixes the tests which where asserting that the result was `map[string]string` by fixing the flattening of the values.

Because the `extended` field is a `TypeMap` it's stored in the Terraform schema as `map[string]any`.

This would only impact on the tests, because the values are passing through the Terraform schema, so being converted internally as `TestFlattenProject` shows.

There was also a bug where the calls to `assert.FailNow()` were setup incorrectly which was causing a panic, I've replaced these with `assert.Equal`.

**Finally:** I added support for running the tests pre-integration (in the pull_request workflow).

## Testing

These changes only impact on tests. 

Not a breaking change.

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -7,7 +7,6 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   GITHUB_OWNER: ${{ github.repository_owner }}
-
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


rancher2/structure_project.go
```diff
@@ -40,7 +40,7 @@ func flattenProjectResourceQuotaLimit(in *managementClient.ResourceQuotaLimit) [
 	}
 
 	if len(in.Extended) > 0 {
-		obj["extended"] = in.Extended
+		obj["extended"] = mapStringToAny(in.Extended)
```
andreas-kupries, line 43: Oh, a missing type conversion :(

andreas-kupries (approved):
LGTM.

.github/workflows/pull_request.yaml
```diff
@@ -19,7 +19,15 @@ jobs:
...
-
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0 https://github.com/actions/checkout
+      - name: Run make test
+        run: |
+          mkdir -p /tmp/gotestsum
+          curl -fsL ${{ env.GOTESTSUM_URL }} | tar xvzf - -C /tmp/gotestsum
+          make GOTESTSUM=/tmp/gotestsum/gotestsum test
```
matttrach: since we are checking out the code we might as well use the same process as every other job and load the environment which has gotestsum already installed, that way we don't need to alter the make file.
```
      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0 https://github.com/actions/checkout
        with:
          fetch-depth: 0
      - name: install-nix
        run: |
          curl -L https://nixos.org/nix/install | sh
          source /home/runner/.nix-profile/etc/profile.d/nix.sh
          nix --version
          which nix
      - name: lint Go
        shell: /home/runner/.nix-profile/bin/nix develop --ignore-environment --extra-experimental-features nix-command --extra-experimental-features flakes --keep HOME --keep SSH_AUTH_SOCK --keep GITHUB_TOKEN --keep NIX_SSL_CERT_FILE --keep NIX_ENV_LOADED --keep TERM --command bash -e {0}
        run: |
          make test
```

GNUmakefile
```diff
@@ -1,3 +1,4 @@
+GOTESTSUM=gotestsum
```
matttrach: I appreciate that this allows the user to trade out gotestsum versions, but I think it is better to manage the environment all in one place, which is why the flake.nix exists. It already has gotestsum installed, so just loading that environment before calling the make command is a better approach IMO.

matttrach (changes requested):

andreas-kupries (approved):

matttrach (approved):

PR #2038, opened 2026-02-18
fix: Remove ingress nginx annotation examples

mallardduck:


matttrach:
Hello, @mallardduck, thanks for contributing!
All commits need to have a verified signature in this repo, please amend your commit with a verified signature.
When correct you will see the "verified" tag when looking at the commit on the "commits" tab above.

matttrach (approved):

mallardduck:
Should be all set now whenever you are ready to merge @matttrach

Issue #2102, opened 2026-03-06
rancher2_custom_user_token with ActiveDirectory


disaster37:
Hi,

What is the way to generate token for not local user. In my case I need to generate token for Active Directory user, and then use it for provider to call ressource `rancher2_cluster_role_template_binding`. If the token is not ActiveDirectory user, It can't bind user and group from ActiveDirectory

matttrach:
I believe this is what the https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/custom_user_token is for.

matttrach:
That would require implementing a new provider and using provider aliases.
https://developer.hashicorp.com/terraform/language/block/provider#alias

matttrach:
Let me know if this works for you, thanks!

disaster37:
Yes, I first look about rancher2_custom_user_token, but it seems to be design to only work with local provider. No way to use active directory as auth provider when use rancher2_custom_user_token.

PR #1997, opened 2026-02-03
Migrate login flow to /v1-public endpoint

pmatseykanets:
Addresses https://github.com/rancher/rancher/issues/52505

## Description

Migrates login flow to the new `/v1-public/login` (since Rancher v2.13) endpoint with automatic fallback to the deprecated `/v3-public` for backwards compatibility.

Changes:

- Switch primary login endpoint from `/v3-public` to `/v1-public/login`
- Add automatic fallback to `/v3-public` endpoint when 404 is returned by the new endpoint
- Use `json.Marsha`l to produce the request body instead of string concatenation
- Improve error handling with safe type assertions
- Handle the `ext/` prefix in token IDs (for new `tokens.ext.cattle.io` tokens)

## Testing

Added unit tests covering:

- Successful login via `/v1-public`
- Token ID with `ext/` prefix stripping
- Authentication failures
- Invalid token format handling
- Fallback to `/v3-public` on 404
- Fallback failure scenarios

Not a breaking change.

PR #2108, opened 2026-03-11
Add a way to set auth provider on rancher2_custom_user_token

disaster37:
Addresses #2102 

## Description

Add field `auth_provider` on resource `rancher2_custom_user_token` to generate token for ActiveDirectory user or others. If not set, default is `local`.
I change DoUserLogin function to use v3 API first, instead to continue to use v1. On my Rancher (2.13.1), the v1 api is still here, so it never use the new v3 API. It seems the v3 api support more auth provider like `ping` or `custom_oidc`.

## Testing

I have compiled the binaire and put it on `~/.terraform.d/plugins/terraform.local/local/rancher2/0.1.1/linux_amd64/terraform-provider-rancher2`.
You need to have Rancher with ActiveDirectory Auth provider setted (or other)

Then:

```terraform
provider "rancher2" {
  api_url    = "https://rancher.local.domain"
  insecure   = true
  token_key      = "xxx"
}

resource "rancher2_custom_user_token" "test_ad" {
  auth_provider = "activedirectory"
  username = "user1"
  password = "xxx"
  description = "foo token"
  ttl = 0
}
```

To finish:

```bash
terraform apply
```

Not a breaking change.


matttrach:
This is reverting #1997. While the name is misleading, the v3 API is older. The v1 API is the first version of a new API, while the v3 is the third version of the older API.

matttrach:
Since it reverts a change for the latest release I don't think it can be accepted as is.

disaster37:
Ok, I have go back to use v1 api first.

matttrach:
I am going to close this one based on lack of activity, please feel free to open a new PR with the change once you are ready. Sorry for the confusing API names/endpoints, we are working on it.

PR #2109, opened 2026-03-16
fix: get tests running again

matttrach:
## Description

The overall goal is to adjust tests to the changes in the client code.
Added a "round tripper" which wraps client requests and adds authorization headers.
Added new "token store" which acts as a place to put and retrieve a token in the provider memory, this creates a thread-safe way to update the provider's authentication token for the login resource.

## Testing

unit, short, and long tests added and passing

Not a breaking change.


matttrach:
Decided to add fixing the dev resource in this as well, since it is necessary for long tests to be effective.

PR #2134, opened 2026-04-08
Replace GoReleaser action

tashima42:
See issue: https://github.com/rancher/ecm-distro-tools/issues/866


matttrach:
This has now been fixed upstream, I will pin the version of the binary instead.

matttrach (approved):

Issue #1504, opened 2025-05-08
[Feature] token resource default TTL should match Rancher max token TTL


erSitzt:
### Environment Information

- Terraform version: 1.11.4
- Provider version: 7.0.0
- Rancher version: 2.8.5
- Operating System of the Rancher Node:
- Operating System of the machine running Terraform:
- Kubernetes Distro (k3s/rke2): rke2
- Kubernetes Version: 1.28.x

### Describe the bug
each run wants to recreate the token if/when ttl is not set or set to 0

### To Reproduce
create a token with no ttl or 0 ttl

```
resource "rancher2_token" "token1" {
  cluster_id = data.rancher2_cluster.currentcluster.id
  description = "token1"
}

resource "rancher2_token" "token2" {
  cluster_id = data.rancher2_cluster.currentcluster.id
  description = "token2"
  ttl = 0
}
```

### Actual Result
Each time the ttl will differ...

<img width="1104" alt="Image" src="https://github.com/user-attachments/assets/4b888315-29d0-43e5-84b5-381bdb159397" />

### Expected Result
not sure :)
not using 0 as a default
updated docs

this only happens with my newest cluster, older clusters where the token was created with an older rancher version are still working with no expiration.
I did only find some hints but no release note/ changelog entry with an exact date since when expiraton is forced ( but i spent only a few minutes searching)

samuelgautier:
Same issue with Rancher 2.11.3 and Terraform rancher2 provider 7.0.0 on kubernetes version 1.32
Is there any way to work  arround ?

Shocktrooper:
Also ran into this and looking for a solution

fabn:
Any news? I got the same issue and now my terraform code is locked because of an expired token. What is the proper way to have a never expiring token (or a refreshing one) in terraform state?

matttrach:
There is now a maximum TTL for all tokens setting in Rancher to control this, and the default is 90 days. I don't believe there is a way to have a forever token anymore. The bootstrap resource is the current mechanism to automatically refresh tokens, but we are aware that it isn't an ideal solution. We are working to improve the bootstrap resource in #1638, but there is also talk of improving it before that with no definite traction.

matttrach:
[...]

PR #2137, opened 2026-04-09
fix: update default token TTL to match Rancher

jmeza-xyz:
Addresses #1504

## Description

The rancher2 token resource defaults to 0 in the terraform provider but as of [Rancher v2.8.0](https://github.com/rancher/rancher/releases/tag/v2.8.0) the default value on the rancher server side is 90 days. This causes a drift/replacement message on each update even when the token is not changing. The PR only change the default to match Rancher's default.
## Testing

Made the changes to the schema_token.go and then built the provider locally with `make build`. Then had tofu use the locally built provider. Terraform example:
```hcl
$ cat main.tf
provider "rancher2" {
  api_url   = "https://rancher.sslip.io"
  token_key = "token-TOKEN:REDACTED"
  insecure  = true # Set to false if you have valid certs
}

data "rancher2_cluster" "local" {
  name = "local"
}

resource "rancher2_token" "token_default" {
  cluster_id  = data.rancher2_cluster.local.id
  description = "test-new-default"
}
```
Tofu apply and then tofu plan to confirm no drift from new default.

Apply/Create:
```bash
$ tofu apply -auto-approve
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - rancher/rancher2 in /home/"$USER"/Repos/github.com/Rancher/terraform-provider-rancher2/bin
│
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become
│ incompatible with published releases.
╵
data.rancher2_cluster.local: Reading...
data.rancher2_cluster.local: Read complete after 0s [id=local]

OpenTofu used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  + create

OpenTofu will perform the following actions:

  # rancher2_token.token_default will be created
  + resource "rancher2_token" "token_default" {
      + access_key  = (known after apply)
      + annotations = (known after apply)
      + cluster_id  = "local"
      + description = "test-new-default"
      + enabled     = (known after apply)
      + expired     = (known after apply)
      + id          = (known after apply)
      + labels      = (known after apply)
      + name        = (known after apply)
      + renew       = true
      + secret_key  = (sensitive value)
      + token       = (sensitive value)
      + ttl         = 7776000
      + user_id     = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

rancher2_token.token_default: Creating...
rancher2_token.token_default: Creation complete after 0s [id=token-2tx2c]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```
Plan after creation:
```bash
$ tofu plan
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - rancher/rancher2 in /home/"$USER"/Repos/github.com/Rancher/terraform-provider-rancher2/bin
│
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become
│ incompatible with published releases.
╵
data.rancher2_cluster.local: Reading...
data.rancher2_cluster.local: Read complete after 0s [id=local]
rancher2_token.token_default: Refreshing state... [id=token-2tx2c]

No changes. Your infrastructure matches the configuration.

OpenTofu has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
```

Not a breaking change.


matttrach:
Hi @jmeza-xyz, thanks for the contribution! Would you mind adding "ttl" to the aspell_custom.txt file in this PR?

docs/resources/token.md
```diff
@@ -35,9 +35,11 @@ The following arguments are supported:
 * `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
 * `description` - (Optional/ForceNew) Token description (string)
 * `renew` - (Optional/ForceNew) Renew token if expired or disabled. If `true`, a terraform diff would be generated to renew the token if it's disabled or expired. If `false`, the token will not be renewed. Default `true` (bool)
-* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `0` (int) 
+* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `7776000` (int)
 
-From Rancher v2.4.6 `ttl` is readed in minutes at Rancher API. To avoid breaking change on the provider, we still read in seconds but rounding up division if required.
+From Rancher v2.4.6 `ttl` is read in minutes at Rancher API. To avoid breaking change on the provider, we still read in seconds but rounding up division if required.
+
+From Rancher v2.8.0 Rancher API tokens default/expire at 90days (7776000 seconds). For long-live tokens, please refer to the Rancher setting `auth-token-max-ttl-minutes` and adjust accordingly. Tokens created before v2.8.0 are not affected.
```
bigkevmcd:
```suggestion
From Rancher v2.8.0 Rancher API tokens default/expire at 90days (7776000 seconds). For long-lived tokens, please refer to the Rancher setting `auth-token-max-ttl-minutes` and adjust accordingly. Tokens created before v2.8.0 are not affected.
```

bigkevmcd (changes requested):
I think the Copilot recommendation about `Computed` and `Optional` makes a lot of sense (I was gonna make the same recommendation until I spotted it).

jmeza-xyz:
> I think the Copilot recommendation about `Computed` and `Optional` makes a lot of sense (I was gonna make the same recommendation until I spotted it).

I agree. I removed the default value and added the compute=true. Let me know if anything else comes up and am happy to adjust. Thanks for reviewing!

matttrach:
You will need to change the expander to remove the 'ttl' if it wasn't specified in the config (which is hard to tell in this version of the SDK), this is probably why it has a default rather than being computed,optional. I would watch out for any AI suggestions to alter the provider like this, we are on an old version of the SDK and they tend to assume that the provider is using the Terraform provider framework.

The trick is that this will be defaulted to 0 either way because it is an 'int' type and nil in the config will automatically become the Go nil default for int. Either way we will be sending 0 to the API which is ignored (because it isn't valid?). As far as the API's Get response, when the ttl isn't set it will respond similarly with the 0 value which is what the resource expects. 

In short I think the AI is basically saying not to make this change at all. I disagree though, I think there is a good reason to align the default here with Rancher since it conveys the actual value the TTL will have.

matttrach:
The AI suggestion was also based on your description, it was assuming that the API was going to respond with 90days if we sent in 0, which isn't the case. Our API responds with 0, but we know there is another system which limits the TTL beyond what we are setting in the object. Aligning what we know with the resource config seems like a very good way to make things more transparent.

docs/resources/custom_user_token.md
```diff
@@ -52,7 +52,7 @@ The following arguments are supported:
 * `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
 * `description` - (Optional/ForceNew) Token description (string)
 * `renew` - (Optional/ForceNew) Renew token if expired or disabled. If `true`, a terraform diff would be generated to renew the token if it's disabled or expired. If `false`, the token will not be renewed. Default `true` (bool)
-* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `0` (int) 
+* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `0` (int)
```
matttrach, line 55: should this be `7776000`?

matttrach (approved):

docs/resources/custom_user_token.md
jmeza-xyz: I'm pretty certain that doc is for this schema which has not been modified in this PR. https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/schema_custom_user_token.go#L79

Should I update the `custom_user_token` schema/resource to match the token schema?

matttrach: nope that's fine, just checking.

matttrach (approved):

matttrach:
In case someone is following along and not understanding the most recent change after some testing it turns out the Rancher API responds with `7776000` if the provider sends `0`, therefore this needs to be a computed field.

docs/resources/token.md
```diff
@@ -35,9 +35,11 @@ The following arguments are supported:
 * `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
 * `description` - (Optional/ForceNew) Token description (string)
 * `renew` - (Optional/ForceNew) Renew token if expired or disabled. If `true`, a terraform diff would be generated to renew the token if it's disabled or expired. If `false`, the token will not be renewed. Default `true` (bool)
-* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `0` (int) 
+* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `7776000` (int)
```
bigkevmcd: How about something like this:

Defaults to the `auth-token-max-ttl-minutes` Rancher setting - **note** the value in Rancher is in minutes, the value in Terraform is in seconds (int)

rancher2/schema_token.go
```diff
@@ -66,10 +65,10 @@ func tokenFields() map[string]*schema.Schema {
 		},
 		"ttl": {
 			Type:        schema.TypeInt,
+			Computed:    true,
 			Optional:    true,
 			ForceNew:    true,
-			Default:     0,
-			Description: "Token time to live in seconds",
+			Description: "Token time to live in seconds. Rancher Default is 7776000 (90 days).",
```
bigkevmcd: See above - feel free to abridge.

bigkevmcd (changes requested):

docs/resources/token.md
jmeza-xyz: Does this work? 
```diff
diff --git a/docs/resources/token.md b/docs/resources/token.md
index 915e43da..9443c262 100644
--- a/docs/resources/token.md
+++ b/docs/resources/token.md
@@ -35,11 +35,11 @@ The following arguments are supported:
 * `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
 * `description` - (Optional/ForceNew) Token description (string)
 * `renew` - (Optional/ForceNew) Renew token if expired or disabled. If `true`, a terraform diff would be generated to renew the token if it's disabled or expired. If `false`, the token will not be renewed. Default `true` (bool)
-* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `7776000` (int)
+* `ttl` - (Optional/ForceNew) Defaults to the `auth-token-max-ttl-minutes` Rancher setting. Default `129600` (int)

 From Rancher v2.4.6 `ttl` is read in minutes at Rancher API. To avoid breaking change on the provider, we still read in seconds but rounding up division if required.

-From Rancher v2.8.0 Rancher API tokens default/expire at 90days (7776000 seconds). For long-lived tokens, please refer to the Rancher setting `auth-token-max-ttl-minutes` and adjust accordingly. Tokens created before v2.8.0 are not affected.
+From Rancher v2.8.0 Rancher API tokens default/expire at 90days (129600 minutes). For long-lived tokens, please refer to the Rancher setting `auth-token-max-ttl-minutes` and adjust accordingly. Tokens created before v2.8.0 are not affected.
```

rancher2/schema_token.go
jmeza-xyz: Same here, let me know if this work to commit/push. Thanks!
```diff
diff --git a/rancher2/schema_token.go b/rancher2/schema_token.go
index 5871cd61..63a6cee9 100644
--- a/rancher2/schema_token.go
+++ b/rancher2/schema_token.go
@@ -68,7 +68,7 @@ func tokenFields() map[string]*schema.Schema {
                        Computed:    true,
                        Optional:    true,
                        ForceNew:    true,
-                       Description: "Token time to live in seconds. Rancher Default is 7776000 (90 days).",
+                       Description: "Defaults to auth-token-max-ttl-minutes Rancher setting. Default `129600` minutes (90 days)",
                },
                "user_id": {
                        Type:        schema.TypeString,
```

bigkevmcd: I'm not sure that this makes sense, it says it defaults to two different values?

docs/resources/token.md
bigkevmcd: I don't think you need to show the number at all...putting anything in is confusing as that number is not coming from the terraform provider and it may not be the value in Rancher.

rancher2/schema_token.go
jmeza-xyz: Sorry, trying to be overly explicit makes this confusing. Similar to above, would just removing the value be best?

```diff
diff --git a/rancher2/schema_token.go b/rancher2/schema_token.go
index 5871cd61..44ccd206 100644
--- a/rancher2/schema_token.go
+++ b/rancher2/schema_token.go
@@ -68,7 +68,7 @@ func tokenFields() map[string]*schema.Schema {
                        Computed:    true,
                        Optional:    true,
                        ForceNew:    true,
-                       Description: "Token time to live in seconds. Rancher Default is 7776000 (90 days).",
+                       Description: "Defaults to auth-token-max-ttl-minutes Rancher setting",
                },
```

docs/resources/token.md
jmeza-xyz: is this better

```diff
diff --git a/docs/resources/token.md b/docs/resources/token.md
index 915e43da..b2869ba4 100644
--- a/docs/resources/token.md
+++ b/docs/resources/token.md
@@ -35,11 +35,11 @@ The following arguments are supported:
 * `cluster_id` - (Optional/ForceNew) Cluster ID for scoped token (string)
 * `description` - (Optional/ForceNew) Token description (string)
 * `renew` - (Optional/ForceNew) Renew token if expired or disabled. If `true`, a terraform diff would be generated to renew the token if it's disabled or expired. If `false`, the token will not be renewed. Default `true` (bool)
-* `ttl` - (Optional/ForceNew) Token time to live in seconds. Default `7776000` (int)
+* `ttl` - (Optional/ForceNew) Defaults to the `auth-token-max-ttl-minutes` Rancher setting.

 From Rancher v2.4.6 `ttl` is read in minutes at Rancher API. To avoid breaking change on the provider, we still read in seconds but rounding up division if required.

-From Rancher v2.8.0 Rancher API tokens default/expire at 90days (7776000 seconds). For long-lived tokens, please refer to the Rancher setting `auth-token-max-ttl-minutes` and adjust accordingly. Tokens created before v2.8.0 are not affected.
+From Rancher v2.8.0 the Rancher API and kubeconfig tokens `ttl` is managed by Rancher setting `auth-token-max-ttl-minutes`. Tokens created before v2.8.0 are not affected.
```

rancher2/schema_token.go
bigkevmcd: @jmeza-xyz Yes...that would work.

I wish we didn't have seconds and minutes but that ship sailed 😄

jmeza-xyz: Changes committed.

docs/resources/token.md
jmeza-xyz: Changes committed.

bigkevmcd (approved):
Thanks for fixing this!

PR #2147, opened 2026-04-21
feat: Support OIDC Provider OIDC Clients

bigkevmcd:
Addresses https://github.com/rancher/terraform-provider-rancher2/issues/2156 and https://github.com/rancher/rancher/issues/53639

## Description

Adds support for CRUDding OIDCClients

## Testing

Not a breaking change.


rancher2/resource_rancher2_oidc_client.go
```diff
@@ -0,0 +1,195 @@
...
+	log.Printf("[INFO] Creating OIDCClient")
+	oidcClient, err := expandOIDCClient(d)
+	if err != nil {
+		return err
+	}
+
+	client, err := meta.(managementClientGetter).ManagementClient()
+	if err != nil {
+		log.Printf("[ERROR] getting a client when creating OIDC Client: %s", err)
+		return fmt.Errorf("getting a client when creating OIDC Client: %w", err)
```
JonCrowther: This pattern of logging an error and then returning an error is repeated throughout this file. I don't see other resource files doing this as well. I think we should we remove the `log.Printf` portions unless you have a good reason to keep them.

JonCrowther (commented):
LGTM with just one nit related to the logging

JonCrowther (approved):

matttrach (approved):

docs/resources/oidc_client.md
```diff
@@ -0,0 +1,51 @@
+---
+page_title: "rancher2_oidc_client Resource"
+---
+
+# rancher2\_oidc_client Resource
+
+Provides a Rancher OIDC Client. This can be used to configure the OIDC Clients
+available for the Rancher OIDC Provider.
```
enrichman: Probably an auto-wrap for line length

```suggestion
Provides a Rancher OIDC Client. This can be used to configure the OIDC Clients available for the Rancher OIDC Provider.
```

docs/resources/oidc_client.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Example Usage
+
+### Creating a Rancher OIDC Client.
+
+```hcl
+resource "rancher2_oidc_client" "oidc-test-client" {
+  description = "Access for Test Client"
```
enrichman: We should adhere to the same structure/conventions: https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/user#example-usage


```suggestion
```hcl
# Create a new rancher2 OIDC Client
resource "rancher2_oidc_client" "foo" {
  description = "Foo OIDC Client"
```

docs/resources/oidc_client.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `description` - A human-readable description for the OIDC Client.
+* `token_expiration_seconds` - ID Token and Access Token will only be valid for this many seconds.
+* `refresh_token_expiration_seconds` - How long can the refresh token be used for?
+* `redirect_uris` - Provides a list of allowed redirect URIs for this OIDC Client.
```
enrichman: Maybe it's clearer to put before the required args, also specifying if they are required/optional?

```suggestion
* `redirect_uris` - (Required) List of allowed redirect URIs for this OIDC Client (list)
* `description` - (Optional) A human-readable description for the OIDC Client (string)
* `token_expiration_seconds` - (Optional/Computed) ID Token and Access Token will only be valid for this many seconds (int)
* `refresh_token_expiration_seconds` - (Optional/Computed) How long can the refresh token be used for (int)
```

docs/resources/oidc_client.md
```diff
@@ -0,0 +1,51 @@
...
+
+* `id` - (Computed) The ID of the resource (string)
+* `client_id` - (Computed) The ID to be used when authenticating as this OIDC Client.
+
+## Import
+
+OIDC Clients can be imported using the Client name in the format `<client_name>`
+
+```
+$ terraform import rancher2_oidc_client.foo &lt;CLIENT_NAME&gt;
```
enrichman:
```suggestion
$ terraform import rancher2_oidc_client.foo <client_name>
```

enrichman (commented):
Just some comments related to the docs, but LGTM 🙂

daleckystepan:
How this is supposed to be used because there is no `client_secret` in  attributes? This is form my POV missing in DOCu.

PR #2153, opened 2026-04-22
[Feature] Add user principal id to global role binding

### LINKED ISSUE #2152: [Feature] add user_principal_id to rancher2_global_role_binding
### Environment Information

- Rancher2 Provider version: 14.1.0

### Describe the Feature

I want a user_principal_id in the rancher2_global_role_binding resource to have support for adding remote users and assigning global roles.
I know groups are preferred, but for service accounts that we want to store in active directory it would be nice to have a user_principal_id in cases where we don't want our service accounts as local rancher users.

Raboo:
## Summary

- Add `user_principal_id` field to `rancher2_global_role_binding` resource, enabling assignment of global roles to remote users via their principal ID
- Follows the same pattern already used in `rancher2_cluster_role_template_binding` and `rancher2_project_role_template_binding`
- Adds `user_principal_id` to the data source as well for completeness

## Related

Fixes #2152

## Changes

| File | Change |
|------|--------|
| `rancher2/schema_global_role_binding.go` | Added `user_principal_id` field to schema |
| `rancher2/structure_global_role_binding.go` | Added flatten/expand support for `UserPrincipalID` |
| `rancher2/data_source_rancher2_global_role_binding.go` | Added `user_principal_id` computed attribute |
| `rancher2/structure_global_role_binding_test.go` | Added test coverage for `user_principal_id` |
| `docs/resources/global_role_binding.md` | Added argument reference, example, and note |
| `docs/data-sources/global_role_binding.md` | Added attribute reference |

## Testing

- `TestFlattenGlobalRoleBinding` — PASS
- `TestExpandGlobalRoleBinding` — PASS
- `go build ./...` — compiles cleanly

## Breaking Change Notice

This is a **breaking change** for existing `rancher2_global_role_binding` resources.

The `user_principal_id` field uses `ForceNew: true`, which means that on upgrade, Terraform will detect a diff between the config (which does not specify `user_principal_id`) and the state (which will be populated from the Rancher API during refresh). This triggers a **destroy and recreate** of all existing global role bindings managed by Terraform.

This matches the pattern used in `rancher2_cluster_role_template_binding` and `rancher2_project_role_template_binding`, where `ForceNew: true` was eventually adopted for identity-related fields. The field is immutable at the Rancher API level — a global role binding's principal cannot be changed without recreating the binding.

## Generated with

This PR was made with [OpenCode](https://github.com/opencode) using model Qwen3.6-35B-A3B.


matttrach:
Have you tested "ForceNew" here? Does it cause existing configurations which upgrade to the new version to need to be recreated? If so, that would be a breaking change.

Raboo:
> Have you tested "ForceNew" here? Does it cause existing configurations which upgrade to the new version to need to be recreated? If so, that would be a breaking change.

@matttrach how do I test it? Is there a test that will do it or do I need write a Tofu module that uses this provider to test this situation?

Raboo:
> > Have you tested "ForceNew" here? Does it cause existing configurations which upgrade to the new version to need to be recreated? If so, that would be a breaking change.
> 
> @matttrach how do I test it? Is there a test that will do it or do I need write a Tofu module that uses this provider to test this situation?

I ran a analysis using my LLM;

Yes, **it is a breaking change**. Here's the analysis:

### What's happening

The Rancher API's `GlobalRoleBinding` struct already has a `UserPrincipalID` field. When a user upgrades the provider:
1. Terraform refreshes the resource state by reading from the API
2. `flattenGlobalRoleBinding` calls `d.Set("user_principal_id", in.UserPrincipalID)`
3. If the API returns a non-empty `user_principal_id` (e.g. `"local://user-12345"`), Terraform sees a diff:
   - **Config**: `user_principal_id` = (not specified)
   - **State after Read**: `user_principal_id` = `"local://user-12345"`
4. With `ForceNew: true`, Terraform plans to **destroy and recreate** the resource

The `Computed: true` flag allows Terraform to accept API values, but it does not suppress diffs.


So knowing this is a breaking change (the field is immutable at the API level), what should I do?

Raboo:
I updated the description of the PR with a breaking change notice.

Raboo:
@matttrach do I need to do anything to get this moving along?

matttrach:
Hi @Raboo, I am getting the appropriate team to review, sorry for the wait.

JonCrowther (approved):
LGTM. Thanks for contributing!

matttrach (approved):

Raboo:
Thanks guys! Will this be a part of v15.0.0?

matttrach:
@Raboo, it still needs to be approved by our QA, but it is looking like it will go in our v15 release, yes.

Raboo:
Awesome!

Issue #2207, opened 2026-05-04
[PROBLEM] append_tolerations cannot be removed because expandAgentDeploymentCustomization omits empty lists


skanakal:
### Environment Information

- Terraform version: 1.14.0
- Rancher2 Provider version: 13.1.4
- Rancher version: 2.14
- Operating System of the Rancher Node: ubuntu
- Operating System of the machine running Terraform: opensuse
- Kubernetes Distro (k3s/rke2): rke2
- Kubernetes Version: 1.33
- Infrastructure Provider (AWS/GCP/vSphere/etc): aws
- What is the scope of the token given to the provider? Admin

### Describe the Problem

Removing append_tolerations config does not clear the existing rancher cluster customistation. 
rancher2_cluster provider omits the field from the API payload when the list becomes empty, so rancher never receives an explicit clear operation, this causes tolerations to persist.
append_tolerations cannot be removed because [expandAgentDeploymentCustomization](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/structure_agent_deployment_customization.go#L38) omits empty lists 

Repro steps:

1. Initial terraform:

~~~
resource "rancher2_cluster" "eksimport" {
  name               = "eksimport"
  
  cluster_agent_deployment_customization {
    override_resource_requirements {
      memory_request = var.cattle_cluster_agent_memory_request
    }

#    append_tolerations {
#      key      = "kubernetes.io/arch"
#      value    = "arm64"
#      effect   = "NoSchedule"
#      operator = "Equal"
#    }
  }
}

output "rancher_registration_command" {
  value       = rancher2_cluster.eksimport.cluster_registration_token[0].command
}

output "rancher_registration_insecure_command" {
  value       = rancher2_cluster.eksimport.cluster_registration_token[0].insecure_command
}
~~~
2. uncomment toleration block from  the above and apply 
3. remove append_tolerations block and reapply the toleration

Expected Behavior
toleration on cattle-cluster-agent should be added added/removed

Actual Behavior
toleration on cattle-cluster-agent not applied

Additional information:

Internal reference: SURE-11559

PR #2244, opened 2026-05-06
fix: Fix stale append_tolerations when removed from the terraform configuration

skanakal:
Addresses  https://github.com/rancher/terraform-provider-rancher2/issues/2207

## Description
The provider currently omits append_tolerations from the rancher api payload when the terraform list becomes empty because the expander gates assignment n len(v) > 0, this prevents rancher from receiving an explicit empty list and causes stale tolerations to persist on the cluster spec. 

Move the replace() decision outside the driver-specific so that all cluster types use replace() when deployment customizations change.This aligns imported cluster behavior with existing K3s/RKE2 type clusters and allows removal of deployment customization fields...

## Testing

- Add appened_tolerations 
- modify existing tolerations 
- remove tolerations while keeping the other fields like memory 
- remove entire block

Not a breaking change.


matttrach:
What happens if you don't include the append_tolerations in your config with this change? Will it unexpectedly clear them out? Wouldn't that be a breaking change for existing configurations?

matttrach:
This repo requires commits to be signed by a verified key, please see [Github's docs for how to do that](https://docs.github.com/en/authentication/managing-commit-signature-verification).

skanakal:
> What happens if you don't include the append_tolerations in your config with this change? Will it unexpectedly clear them out? Wouldn't that be a breaking change for existing configurations?

Here is the current schema,  https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/schema_agent_deployment_customization.go#L37-L43

```
"append_tolerations": {
			Type:        schema.TypeList,
			Description: "User-defined tolerations to append to agent",
			Optional:    true,
			Elem: &schema.Resource{
				Schema: tolerationFields(),
			},
		},
```


Do you mean even though the current schema suggests they are distinguishable, this relies on terraform behavior rather than explicit state comparison...

matttrach:
@skanakal, this still needs a review approval from one of your teammates before I can merge.

matttrach (approved):

rancher2/structure_agent_deployment_customization.go
```diff
@@ -44,7 +44,7 @@ func expandAgentDeploymentCustomization(p []interface{}) (*managementClient.Agen
 
 	in := p[0].(map[string]interface{})
 
-	if v, ok := in["append_tolerations"].([]interface{}); ok && len(v) > 0 {
+	if v, ok := in["append_tolerations"].([]interface{}); ok {
```
jiaqiluo: The same pattern is also seen in the `override_resource_requirements` field at line 59, which means users will run into the same issue if they want to drop override_resource_requirements values.

It looks like a better fix is to update [this logic](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/resource_rancher2_cluster.go#L270-L276) to check `replace = d.HasChange("cluster_agent_deployment_customization") || d.HasChange("fleet_agent_deployment_customization")` regardless of the cluster driver type, so in your case of an EKS imported cluster, a [Replace operation](https://github.com/rancher/terraform-provider-rancher2/blob/main/rancher2/resource_rancher2_cluster.go#L283) will be invoked, which drops the append_tolerations as expected. 

Can you give the above a try?

jiaqiluo (changes requested):

rancher2/structure_agent_deployment_customization.go
skanakal: @jiaqiluo Thank you for reviewing it.
I've updated the fix to apply the existing replace behavior across all cluster drivers whenever `cluster_agent_deployment_customization` or `fleet_agent_deployment_customization` changes...

jiaqiluo: @skanakal, did you test the fix locally to confrim it works?

skanakal: @jiaqiluo I have tested this along with the fix https://github.com/rancher/rancher/pull/54923/changes and I confirm this is woriking...
It would be great help if you can review the above fix too...

see this:


https://github.com/user-attachments/assets/5f57de53-bd83-4d7b-bea8-18a9ec294300

jiaqiluo: @skanakal, thank you for the confirmation and the proof.  I'll review the other PR.

jiaqiluo (approved):
LGTM.
See https://github.com/rancher/terraform-provider-rancher2/pull/2244#discussion_r3468143225 for proof of fix.

matttrach:
@skanakal,  this repository requires all commits to be signed with a verified key.

matttrach:
Once your commits are signed I can merge, thanks!

skanakal:
> Once your commits are signed I can merge, thanks!

@matttrach  Done. Thank you.

Issue #1229, opened 2023-09-20
[Problem] Data sources create new kubeconfig token when running terraform plan


erSitzt:
### Rancher Server Setup
- Rancher version: 2.7.5
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

### Information about the Cluster
- Kubernetes version: 1.24.x
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Imported

### User Information
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Amin
  - If custom, define the set of permissions: 

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? 3.1.1
- What is the version of Terraform in use? 1.5.6

### Describe the bug
Same Bug as described in #841 
There was no reaction in the closed issue, so to be safe i opened a new one...

I think there maybe something wrong with the changes in #1165 but im not sure.

As of my last comment in the closed issue #841 `origconfig` is always empty in case the datasource is used... with a cluster resource it may be ok and working...

This is the part
https://github.com/rancher/terraform-provider-rancher2/blob/master/rancher2/resource_rancher2_cluster.go#L668

Anddd7:
any update on this?

Anddd7:
any update on this?

nikos445:
Any update on this?

mandronic0:
[SURE-9508]

matttrach:
Some basic reproduction information, according to user this is still happening:
- Terraform version: v1.5.6
- Rancher2 Provider Version: v13.1.4
- Rancher Version: v2.13.1
- Kubernetes Version: v1.33.2+rke2r1

matttrach:
It appears the provider is generating a kubeconfig with every plan/read operation.

Priyashetty17:
Verified with Terraform Provider `v15.0.0-rc.11` against Rancher `v2.15-56295d737e805e1dc237233b049315bb5f3edd5a-head`:
- **Test1**: generate_kube_config = false: No new tokens created on plan/apply, kubeconfig is empty ✅
[...]

PR #2269, opened 2026-05-27
fix: Stop data sources from generating kubeconfig tokens on every plan

pmatseykanets:
Addresses #1229

## Description

Cluster data sources called the Rancher generateKubeconfig API on every plan and apply, creating a new API token each time. Over time this leads to accumulation of large numbers of orphaned tokens.

Data sources now skip kubeconfig generation by default, **which is a breaking change**. Users who need the kubeconfig output can opt in by setting generate_kube_config = true on the data source.

The alternative is to make it opt-out in which case the side-effect remains the default behavior.

## Testing

This is a breaking change.

`kube_config` on cluster data sources (`rancher2_cluster`, `rancher2_cluster_v2`) is now empty by default.
Previously, every plan/apply generated a new API token. Users who need `kube_config` from a data source must set `generate_kube_config = true`.

### DIFF
--- a/aspell_custom.txt
+++ b/aspell_custom.txt
@@ -8,6 +8,7 @@ azure
 backport
 cognito
 config
+datasource
 destructuring
 eks
 eslint
--- a/rancher2/data_source_rancher2_cluster.go
+++ b/rancher2/data_source_rancher2_cluster.go
@@ -37,8 +37,8 @@ func dataSourceRancher2Cluster() *schema.Resource {
 			"generate_kube_config": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
-				Description: "Generate a kubeconfig for the cluster. Warning: this creates a new API token on each plan/apply.",
+				Default:     true,
+				Description: "Generate a kubeconfig for the cluster. Set to false to avoid creating a new API token on each plan/apply. Default will change to false in a future version.",
 			},
 			"ca_cert": {
 				Type:      schema.TypeString,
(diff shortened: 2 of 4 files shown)


JonCrowther (approved):

bigkevmcd (approved):

matttrach:
@pmatseykanets,
This repository requires all commits to be signed with a verified key.

matttrach:
I see your latest commit is validated, you may need to rebase and squash to get them all validated if amending them doesn't work.

matttrach (approved):

matttrach:
Once all the commit are verified I can merge.

pmatseykanets:
@matttrach Any concerns about this being breaking change?

matttrach:
I mean, yeah, there is a concern, especially since this is a really common use case, but we are releasing it in a new major release (v15.0.0) so hopefully our users aren't caught too unaware. I know this repo isn't normal for Terraform providers and our user's may be a bit numb to our release versions, but we are following semver properly now.

matttrach:
It would be helpful to indicate a breaking change with an exclamation point in the commit message, eg. `fix!: Stop data sources from generating kubeconfig tokens on every plan`, but that isn't necessary. I will fix it when I squash merge it in along with any other message issues I find. This is used by release-please to generate the release notes, so I try to make sure it is formatted properly and doesn't have any spelling issues.

pmatseykanets:
@matttrach the description talks about the alternative too, but it less desirable IMHO as it preserves the bad side-effect. But it's an option.

> The alternative is to make it opt-out in which case the side-effect remains the default behavior.

matttrach:
This repo hasn't previously shied away from breaking changes, but maybe we should use a similar approach as other Rancher repos and make it opt-in for v15, then the default behavior for v16 ? 
I think this would be a good standard to have and it may provide some good will with our larger users.

matttrach (approved):

PR #2281, opened 2026-06-03
feat: make `terraform fmt` print diff

m-ildefons:
## Description

The CI pipeline uses `terraform fmt` to enforce the Terraform styleguide on example Terraform modules.
With the `-diff` flag, the output of `terraform fmt` will not only contain the name of the file violating the style guide, but also a diff showing where exactly the violation takes place. This makes it much easier to find the problem and fix it for PR authors.

## Testing

CI pipeline should pass on main branch as usual, but in case a bogus example Terraform module is submitted, the CI pipeline should print detailed information on what's wrong.


matttrach:
This is great, thank you!

matttrach (approved):

Issue #2284, opened 2026-06-03
[Feature] Add inheritedNamespacedRule field to global roles


JonCrowther:
### Environment Information
- Rancher2 Provider version:
- Rancher version: 2.15
- Infrastructure Provider (AWS/GCP/vSphere/etc):

### Describe the Feature

A new field was added to global roles here https://github.com/rancher/rancher/issues/50663. Would like it to be available via terraform as well.

joesims22:
Validated on Rancher `v2.15.1-0847cd1042eec4fb94973824f69a588ad2aa4f3d-head` with  terraform provider rancher2 `v15.0.0-rc.16`.

| Test Case | Description | Result | Status |
| :--- | :--- | :--- | :--- |
| **Resource Creation** | Create a `rancher2_global_role` defining `inherited_namespaced_rules` (with `namespace` and nested `rules`). | Successfully creates the Global Role with the correct nested payload sent to the API. | ✅ Passed |
| **Data Source** | Read the role via the `rancher2_global_role` data source. | Successfully fetches and outputs the exact `inherited_namespaced_rules` structure. | ✅ Passed |
| **In-Place Update** | Modify the nested `resources` or `verbs` lists in an existing configuration. | Properly triggers an in-place update (`~`) rather than a resource replacement (`ForceNew`). | ✅ Passed |
| **Block Removal** | Completely remove the `inherited_namespaced_rules` block from the Terraform config. | Cleanly nullifies/removes the rules from the Rancher API without leaving state drift. | ✅ Passed |

PR #2340, opened 2026-07-22
Add the inherited-namespaced-rules field to globalroles

JonCrowther:
- Addresses https://github.com/rancher/terraform-provider-rancher2/issues/2284

## Description
This adds the `inheritedNamespacedRules` field to the global role resource. This was added in rancher/rancher but hasn't been propagated to the Rancher Terraform Provider yet.

## Testing
I added some unit tests, but I also tested this manually by creating a `.tf` file and then doing `terraform apply` and `terraform destroy`. Then I checked that on Rancher the globalrole was providing the right permissions. All this works well.

Not a breaking change. It is additive.


matttrach:
@JonCrowther, please amend your commit message to have the proper prefix? I believe "feat:" is most likely correct.

rancher2/structure_global_role.go
```diff
@@ -85,7 +90,42 @@ func expandGlobalRole(in *schema.ResourceData) *managementClient.GlobalRole {
...
+	if len(in) == 0 {
+		return map[string][]managementClient.PolicyRule{}
+	}
+
+	out := make(map[string][]managementClient.PolicyRule, len(in))
+	for _, ruleSet := range in {
+		ruleSetMap := ruleSet.(map[string]interface{})
+		policyRules, _ := ruleSetMap["rules"].([]interface{})
+		out[ruleSetMap["namespace"].(string)] = expandPolicyRules(policyRules)
+	}
```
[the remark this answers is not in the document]
matttrach: This feels like a legitimate concern, @JonCrowther, please look into this?
If this isn't a problem please just explain, thanks!

matttrach (changes requested):

rancher2/structure_global_role.go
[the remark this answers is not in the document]
JonCrowther: Added a duplication check for the plan phase.

bigkevmcd (approved):
Looks good.

rancher2/schema_global_role.go
```diff
@@ -52,6 +52,26 @@ func globalRoleFields() map[string]*schema.Schema {
 				Type: schema.TypeString,
 			},
 		},
+		"inherited_namespaced_rules": {
+			Type:        schema.TypeSet,
```
pmatseykanets, line 57: Seems like `List` type used elsewhere for this kind of data. `Set` is unordered. There is a chance it can issues with apply.
```suggestion
			Type:        schema.TypeList,
```

Otherwise it requires justification for a chosen type.

rancher2/resource_rancher2_global_role.go
```diff
@@ -18,6 +18,18 @@ func resourceRancher2GlobalRole() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceRancher2GlobalRoleImport,
 		},
+		CustomizeDiff: func(d *schema.ResourceDiff, i any) error {
```
pmatseykanets, line 21: This new "no duplicate namespace" rule has no tests.
Ideally, we want a test with two duplicate namespace blocks (should fail) and two different ones (should pass).

rancher2/schema_global_role.go
```diff
@@ -52,6 +52,26 @@ func globalRoleFields() map[string]*schema.Schema {
...
+						Type:     schema.TypeString,
+						Required: true,
+					},
+					"rules": {
+						Type:     schema.TypeList,
+						Optional: true,
+						Elem: &schema.Resource{
+							Schema: policyRuleFields(),
+						},
+					},
```
pmatseykanets, line 77: New fields `namespace` and `rules` have no description while all adjacent ones do.

docs/resources/global_role.md
```diff
@@ -55,6 +66,13 @@ The following attributes are exported:
 * `resources` - (Optional) Policy rule resources (list)
 * `verbs` - (Optional) Policy rule verbs. `bind`, `create`, `delete`, `deletecollection`, `escalate`, `get`, `impersonate`, `list`, `manage-namespaces`, `patch`, `update`, `updatepsa`, `use`, `view`, `watch`, `own` and `*` values are supported (list)
 
+### `inherited_namespaced_rules`
```
pmatseykanets, line 69: New "namespace must be unique" rule is not documented.

rancher2/schema_global_role.go
```diff
@@ -52,6 +52,26 @@ func globalRoleFields() map[string]*schema.Schema {
...
 		},
+		"inherited_namespaced_rules": {
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "Policy rules granted in matching namespaces of every cluster besides the local cluster",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"namespace": {
+						Type:     schema.TypeString,
+						Required: true,
```
pmatseykanets: Field allows an empty string.

```suggestion
						Required: true,
						ValidateFunc: validation.StringIsNotEmpty,
```

Requires import

```go
	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
```

rancher2/structure_global_role.go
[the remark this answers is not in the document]
pmatseykanets: Since the check lives outside this function it might worth adding a comment in func doc block.

rancher2/schema_global_role.go
JonCrowther: The `inheritedNamespacedRules` is a map which doesn't guarantee order. The other fields are all slices which is why they can be `List` since they have a fixed order. So I need to use `Set`.

rancher2/resource_rancher2_global_role.go
```diff
@@ -18,6 +18,18 @@ func resourceRancher2GlobalRole() *schema.Resource {
...
+			seen := map[string]bool{}
+			for _, ruleSet := range d.Get("inherited_namespaced_rules").(*schema.Set).List() {
+				namespace := ruleSet.(map[string]any)["namespace"].(string)
+				if seen[namespace] {
+					return fmt.Errorf("inherited_namespaced_rules cannot contain duplicate entries for namespace %q", namespace)
+				}
+				seen[namespace] = true
+			}
```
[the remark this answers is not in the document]
JonCrowther: Added a check for an unset attribute

rancher2/resource_rancher2_global_role.go
JonCrowther: Moved it to it's own function to test it properly

rancher2/structure_global_role_test.go
```diff
@@ -130,6 +132,49 @@ func init() {
...
+			"node_two": "two",
+		},
+		Labels: map[string]string{
+			"option1": "value1",
+			"option2": "value2",
+		},
+		InheritedNamespacedRules: map[string][]managementClient.PolicyRule{
+			"namespace-one": testGlobalRolePolicyRulesConf,
+			"namespace-two": {},
+		},
```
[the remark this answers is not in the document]
JonCrowther: I don't think this is right at all. You absolutely can skip composite literals within a set. From the docs:

>  Within a composite literal of array, slice, or map type T, elements or map keys that are themselves composite literals may elide the respective literal type if it is identical to the element or key type of T. Similarly, elements or keys that are addresses of composite literals may elide the &T when the element or key type is *T.

matttrach (approved):

PR #2343, opened 2026-07-23
fix: Honor provider timeout

nocturo:
## Description

The provider 'timeout' setting is stored in Config.Timeout and used to bound
retry loops, but the HTTP client timeout for clients built via
CreateClientOpts() (the v2 CatalogV2/cluster clients and the custom user token
client) is never set from it. norman's clientbase.NewAPIClient then falls back
to a hardcoded 60s when ClientOpts.Timeout == 0, so these clients ignore the
configured timeout entirely.

This surfaces on rancher2_cluster_v2 reads, which build a CatalogV2 client
scoped to the local cluster and fetch the schema set from
/k8s/clusters/local/v1/schemas during NewAPIClient. On larger/busier setups
that proxied endpoint can take longer than 60s to return headers, failing with
'context deadline exceeded (Client.Timeout exceeded while awaiting headers)',
and raising the provider timeout has no effect.

Pass Config.Timeout into clientbase.ClientOpts.Timeout so the documented
setting governs the HTTP client.

## Testing

I was experiencing timeouts using provider on flaky connection as it took little
longer to fetch it but it hit this timeout first instead of honoring provider timeout.
With this patch, I did not experience timeouts.

Not a breaking change.


matttrach:
@nocturo, 
This repository requires all commits to be signed with a verified key.

Please amend the commit message with the prefix "feat:" so that it passes the commit validation CI as well.

nocturo:
@matttrach thanks, signed the commit with verified key now and updated the commit message.

bigkevmcd (approved):
Thanks for the fix.

matttrach:
looks like this was already added in #2349 sorry about that, closing.

PR #2366, opened 2026-08-03
fix: resolve user_principal_id to user_id on global role binding create

### LINKED ISSUE #2152: [Feature] add user_principal_id to rancher2_global_role_binding

Raboo:
## Summary

When creating a `rancher2_global_role_binding` with only `user_principal_id` configured, the Rancher API rejects the request with a 422 error:

```
Error: Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [baseType=error, code=InvalidBodyContent, message=must contain field [groupPrincipalId] OR field [userId]]
```

The Rancher API validator (`pkg/api/norman/customization/globalrolebinding/validator.go`) requires **exactly one** of `groupPrincipalId` OR `userId` in the create payload — `userPrincipalId` alone is not accepted.

## Fix

This PR resolves `user_principal_id` to the matching Rancher `user_id` before sending the create request:

- Adds `Config.GetUserIDByPrincipalID()` which lists users and finds the one whose `PrincipalIDs` contain the configured principal ID (mirroring Rancher's own `GetUserByPrincipalID` logic)
- Updates `resourceRancher2GlobalRoleBindingCreate` to set `UserID` from `UserPrincipalID` when only the principal is configured
- Adds unit tests for the principal → user lookup (`TestFindUserByPrincipalID`)

## Related

Fixes #2152

## Changes

| File | Change |
|------|--------|
| `rancher2/config.go` | Added `GetUserIDByPrincipalID()` and `findUserByPrincipalID()` helpers |
| `rancher2/config_test.go` | Unit tests for `findUserByPrincipalID()` |
| `rancher2/resource_rancher2_global_role_binding.go` | Resolve `user_principal_id` → `user_id` before create |
| `docs/resources/global_role_binding.md` | Documented the lookup behavior |

## Testing

- `TestFindUserByPrincipalID` — PASS
- `TestFlattenGlobalRoleBinding` — PASS
- `TestExpandGlobalRoleBinding` — PASS
- `go test ./rancher2` — full unit suite PASS
- `go build ./...` — compiles cleanly

## Generated with

This PR was made with [OpenCode](https://github.com/opencode) using model DeepSeek-V4-Flash.


rancher2/resource_rancher2_global_role_binding.go
```diff
@@ -32,6 +32,17 @@ func resourceRancher2GlobalRoleBinding() *schema.Resource {
...
+	// The Rancher API requires either userId or groupPrincipalId to be set when
+	// creating a global role binding. If only user_principal_id is configured,
+	// resolve the matching user ID before sending the request.
+	if len(globalRole.UserID) == 0 && len(globalRole.UserPrincipalID) > 0 {
+		userID, err := meta.(*Config).GetUserIDByPrincipalID(globalRole.UserPrincipalID)
+		if err != nil {
+			return err
+		}
+		globalRole.UserID = userID
+	}
```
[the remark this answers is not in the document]
Raboo: Good catch, thank you! I've updated the guard so the user_principal_id → user_id resolution only runs when no group_principal_id is configured — this prevents a payload containing both userId and groupPrincipalId, which the validator would reject. The fix is in commit c9901adb.

JonCrowther (approved):
Thank you for the fix!

rancher2/resource_rancher2_global_role_binding.go
```diff
@@ -32,6 +32,18 @@ func resourceRancher2GlobalRoleBinding() *schema.Resource {
 func resourceRancher2GlobalRoleBindingCreate(d *schema.ResourceData, meta interface{}) error {
 	globalRole := expandGlobalRoleBinding(d)
 
+	// The Rancher API requires either userId or groupPrincipalId to be set when
+	// creating a global role binding, but rejects the request when both are set.
+	// If only user_principal_id is configured, resolve the matching user ID
+	// before sending the request.
+	if len(globalRole.UserID) == 0 && len(globalRole.GroupPrincipalID) == 0 && len(globalRole.UserPrincipalID) > 0 {
```
bigkevmcd: nit: I'd use `== ""` comparisons rather than checking the length, it's easier to read.

But that's the common pattern in the terraform provider, so I understand why it's done here.

bigkevmcd (approved):

rancher2/resource_rancher2_global_role_binding.go
Raboo: Thanks! Updated to use `== ""` comparisons for readability in commit 47fd2d8e.

matttrach (approved):

Issue #1333, opened 2024-04-02
[Problem] cluster grace period can't be set to unlimited


qaz-t:
### Rancher Server Setup
- Rancher version: `v2.7.9`
- Installation option (Docker install/Helm Chart): `Helm Chart`
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): `RKE2` `v1.26.15+rke2r1`
- Proxy/Cert Details:

### Information about the Cluster
- Kubernetes version: `v1.26.15+rke2r1`
- Cluster Type (Local/Downstream): `Downstream`
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): `Custom`

### User Information
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: `Admin`

### Provider Information
- What is the version of the Rancher v2 Terraform Provider in use? `4.1.0`
- What is the version of Terraform in use? `1.7.4`

### Describe the bug

I defined a custom rke2 cluster using the following config:
```
resource "rancher2_cluster_v2" "custom_cluster" {
  name               = var.cluster_name
  kubernetes_version = var.kubernetes_version
  cluster_agent_deployment_customization {}
  fleet_agent_deployment_customization {}
  rke_config {
    chart_values          = <<EOF
rke2-cilium: {}
EOF
    machine_global_config = <<EOF
cni: "cilium"
disable-kube-proxy: false
etcd-expose-metrics: false
profile: null
EOF
    machine_selector_config {
      config = <<EOF
protect-kernel-defaults: false
system-default-registry: docker.io
EOF
    }
    upgrade_strategy {
      control_plane_concurrency = "1"
      control_plane_drain_options {
        delete_empty_dir_data                = true
        disable_eviction                     = false
        enabled                              = false
        force                                = false
        grace_period                         = -1
        ignore_daemon_sets                   = true
        ignore_errors                        = false
        skip_wait_for_delete_timeout_seconds = 0
        timeout                              = 120
      }
      worker_concurrency = "1"
      worker_drain_options {
        delete_empty_dir_data                = true
        disable_eviction                     = false
        enabled                              = true
        force                                = false
        grace_period                         = -1
        ignore_daemon_sets                   = true
        ignore_errors                        = false
        skip_wait_for_delete_timeout_seconds = 0
        timeout                              = 120
[...]

PR #2472, opened 2026-09-09
[ISSUE-1333] Allow to set cluster grace period to unlimited to respect terminationGracePeriodSeconds

riuvshyn:
[ISSUE-1333] Allow to set cluster grace period to unlimited to respect terminationGracePeriodSeconds

- Addresses https://github.com/rancher/terraform-provider-rancher2/issues/1333

## Description
allow to pass negative value for grace period to respect  `terminationGracePeriodSeconds` https://github.com/kubernetes/kubectl/blob/master/pkg/drain/drain.go#L57-L60

## Testing
Fix is validated with a local provider build from this branch, before fix:
image

after fix:
image

Not a breaking change.


riuvshyn:
hey @matttrach can you please take a look at this one 🙏🏽

matttrach (approved):
From a terraform perspective this is good, but I would like to see testing from the upgrade perspective. If someone has been using the default all along, and hasn't changed anything will this now pick up on the 0 and start altering the grace period on pods during termination?
If so, this could be considered a breaking change.

matttrach:
This still needs team approval before it can merge.

jiaqiluo (changes requested):
Thank yoy for the contribution! The changes look good to me! 
I would just suggest updating the documentation to clarify this behavior.

In docs/resources/cluster_v2.md line 1182, could we update it to something like:

```
* `grace_period` - (Optional/computed, int) Time in seconds given to each pod to terminate gracefully. `0` means "delete immediately"; a negative value (`-1`) uses the pod's `terminationGracePeriodSeconds`.
```

riuvshyn:
@jiaqiluo updated docs

@matttrach 
> If someone has been using the default all along, and hasn't changed anything will this now pick up on the 0 and start altering the grace period on pods during termination?

Default value is set by rancher not the provider and default value is currently set by rancher is `0`, which is rather dangerous misconfiguration, no one would want the terminate pods without any grace period. We are not changing any defaults here, we are only making it possible to set negative value to be able to activate "respect pod terminationGracePeriodSeconds" as in most cases different workloads have very different grace period requirements.

Updating default value in rancher from `0` to `-1` is a separate change that I want to propose to have safer defaults.

jiaqiluo (approved):

riuvshyn:
@jiaqiluo @matttrach should I update commit message ? or it can be squashed on merge? (I don't know if that check is required)

